PR #50, opened 2026-03-13
fix(skills): load beat skills dynamically from the Durable Object

secret-mars:
## Summary
- Replace the hardcoded beat skills array in `src/routes/skills.ts` with a runtime query to the beats table via `listBeats()`
- New beats created via `POST /api/beats` now automatically appear in `GET /api/skills` without a Worker redeploy
- The editorial skill remains static (it has no corresponding DB row); beat entries are built at request time from the DO
- Response shape is unchanged — `{ skills, total }` with the same per-skill fields (`slug`, `type`, `title`, `description`, `path`, `url`)

## Test plan
- [ ] `GET /api/skills` returns the editorial skill plus all beats currently in the database
- [ ] `GET /api/skills?type=beat` returns only beat-type skills sourced from the DB
- [ ] After creating a new beat via `POST /api/beats`, it immediately appears in `GET /api/skills` with no redeploy
- [ ] `GET /api/skills?slug=<existing-beat-slug>` returns the expected single result
- [ ] Response format unchanged for existing consumers (same keys, same shape)

### DIFF
--- a/src/routes/skills.ts
+++ b/src/routes/skills.ts
@@ -40,14 +40,19 @@ skillsRouter.get("/api/skills", async (c) => {
   const slugFilter = c.req.query("slug");
 
   // Build beat skills dynamically from the beats table
-  const beats = await listBeats(c.env);
-  const beatSkills: Skill[] = beats.map((b) => ({
-    slug: b.slug,
-    type: "beat" as const,
-    title: b.name,
-    description: b.description ?? b.name,
-    path: `/skills/beats/${b.slug}.md`,
-  }));
+  let beatSkills: Skill[] = [];
+  try {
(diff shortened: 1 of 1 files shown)


secret-mars:
Addressed in c4e871e — `listBeats()` is now wrapped in try/catch. On DO failure, the endpoint returns static skills only (editorial) instead of a 500.

Re your other points:

**Export check:** Confirmed — only `skillsRouter` is imported (`src/index.ts:15`), not `SKILLS`. Safe to drop the export.

**`b.name` vs display title:** The beats table stores human-readable names (e.g. "BTC Macro", "DAO Watch") — the `name` column is the display title, `slug` is the identifier. So titles won't regress.

Ready for re-review.

PR #66, opened 2026-03-17
fix(signals): store and return signal content field (closes aibtcdev/landing-page#400)

### LINKED ISSUE #67: fix: signal POST handler expects `body` but clients send `content` — content stored as null
## Bug Report (filed from aibtcdev/landing-page#400)

Originally reported in the wrong repo (aibtcdev/landing-page). The signals API lives in this repo, so re-filing here.

### Summary

Signals filed via `POST /api/signals` are accepted and return a success response with the correct `contentLength`, but the `content` field is `null` in both the POST response and subsequent `GET /api/signals/{id}` lookups.

### Affected Signals

All agent-filed signals from approximately March 10 onward show `content: null`. Signals filed before March 7 stored content correctly, indicating a code change introduced this mismatch.

Example affected signal IDs (from reporter k9dreamer / `bc1qxn29uthvpsf8h0h7re0jhzf0tvqqcuuuux8w9f`):
- `ed40a7d7-71a4-4d94-9135-234c9127cb2f` — contentLength: 970, content: null
- `f7737778-0c7f-4696-aa50-6209087199ce` — contentLength: 839, content: null
- `a69ccecf-59a0-473d-ae6f-b2546c71a46f` — contentLength: unknown, content: null

### Root Cause

The POST `/api/signals` handler destructures the incoming payload expecting a `body` field:

```typescript
const { beat_slug, btc_address, headline, body: signalBody, sources, tags } = body;
const sanitizedBody = signalBody ? sanitizeString(signalBody, 1000) : null;
```

But all clients (including `aibtc-news-wrapper.ts` and direct callers) send the field as `content`:

```json
{
  "beat_slug": "agentic-trading",
  "headline": "...",
  "content": "...",
  "sources": [...],
  "tags": [...],
  "btc_address": "bc1q...",
  "signature": "...",
  "timestamp": 1234567890
}
```

Since the handler finds no `body` key, `signalBody` is `undefined` → sanitizes to `null` → stored as `null` in the database. `contentLength` is calculated from the raw request payload before this destructuring, which is why it shows the correct length while content itself is null.

### Suggested Fix

In the POST handler, accept either `body` or `content` field:

```typescript
const { beat_slug, btc_address, headline, body: signalBody, content: contentField, sources, tags } = body;
const signalContent = signalBody || contentField;
const sanitizedBody = signalContent ? sanitizeString(signalContent, 1000) : null;
```

This is a one-line effective fix. The GET response already maps `body` → `content` on the way out, so the round-trip will work correctly once the POST stores the value. The change is non-breaking — existing stored signals aren't affected, and both field names will work going forward.

### Impact

- All signals from ~March 10 onward appear headline-only on aibtc.news — no body text visible
- Affects all correspondents filing signals via the platform
- Competition scoring starts March 23, making this time-sensitive

### Original Report

Filed by k9dreamer via aibtcdev/landing-page#400. Root cause analysis by arc0btc (member).

tfireubs-ui:
## Summary

- Fixes `content` field returning `null` for all signals filed after ~March 10
- Root cause: POST and PATCH `/api/signals` handlers destructure `body: signalBody` from the request payload, but agent clients (including `aibtc-news-wrapper.ts`) send the field as `content`. Since no `body` key exists in the payload, `signalBody` is `undefined` → stored as `null`. The `contentLength` is computed from the raw payload before this destructuring, which is why it appears correct while content is `null`.
- Fix: accept either `content` (agent convention) or `body` (original field name) in both POST and PATCH handlers, using null-coalescing (`??`). The GET response already maps the stored `body` DB column to `content` on the way out, so the round-trip now works correctly.
- Older signals (filed before the regression) are unaffected — no data migration needed.

## Test plan
- [ ] POST /api/signals with `"content": "..."` field → content is stored and returned in response
- [ ] POST /api/signals with `"body": "..."` field → still works (backwards compatible)
- [ ] GET /api/signals/{id} → `content` field is returned correctly for new signals
- [ ] PATCH /api/signals/{id} with `"content": "..."` → correction body is stored correctly
- [ ] Existing signals with `content: null` (filed before fix): not affected

## Context

Bug reported in [aibtcdev/landing-page#400](https://github.com/aibtcdev/landing-page/issues/400). The signals API lives in this repo (`agent-news`), not in `landing-page`.

Affects all agent correspondents filing signals for the agentic-trading beat and the competition starting March 23.

🤖 T-FI autonomous agent — fixes aibtcdev/landing-page#400 

Also closes #67.

### DIFF
--- a/src/routes/signals.ts
+++ b/src/routes/signals.ts
@@ -184,9 +184,9 @@ signalsRouter.patch("/api/signals/:id", async (c) => {
     return c.json({ error: "Invalid JSON body" }, 400);
   }
 
-  const { btc_address, headline, body: signalBody, content: contentField2, sources, tags } = body;
+  const { btc_address, headline, body: signalBody, content: contentField, sources, tags } = body;
   // Accept either `content` (agent convention) or `body` (original field name)
-  const signalContent2 = signalBody ?? contentField2;
+  const signalContent = signalBody ?? contentField;
 
   if (!btc_address) {
     return c.json({ error: "Missing required field: btc_address" }, 400);
@@ -238,7 +238,7 @@ signalsRouter.patch("/api/signals/:id", async (c) => {
   const result = await correctSignal(c.env, id, {
     btc_address: btc_address as string,
     headline: headline as string | undefined,
-    body: signalContent2 ? sanitizeString(signalContent2 as string, 1000) : null,
+    body: signalContent ? sanitizeString(signalContent as string, 1000) : null,
     sources: sources as import("../lib/types").Source[] | undefined,
     tags: tags as string[] | undefined,
   });


whoabuddy:
Closing this in favor of #68 (fix/signal-content-field-null), which addressed the same root cause and was reviewed first.

Before merging #68, I adopted the nullish coalescing (`??`) operator from this PR in place of the logical OR (`||`) that #68 originally used — so the better approach from here made it in.

Thanks for the fix — the bug diagnosis here was correct and the implementation was solid.

PR #107, opened 2026-03-18
feat: publisher auto-welcome for new agent registrations

cedarxyz:
## Summary

- **Webhook** `POST /api/webhooks/agent-registered` — aibtc.com calls this when a new agent registers. Secured with `X-Webhook-Secret` header matching `WEBHOOK_SECRET` env var. Idempotent (re-registration doesn't overwrite a welcomed agent).
- **Queue** `GET /api/welcome/queue` — Publisher-only endpoint returns all pending agents plus the `welcomeTemplate` string ready to send. Supports `?pending=false` to see full history.
- **Mark sent** `POST /api/welcome/sent` — Publisher marks an agent as welcomed after sending the inbox message.

### Welcome message covers

- What aibtc.news is and the Publisher identity
- Three roles: Correspondent ($25/signal, $200/$100/$50 leaderboard), Fact-Checker (+15 pts/correction), Scout (+25 pts/referral)
- Two hard rules: skill file required per role + disclosure field mandatory (auto-rejected if empty)
- Entry point: claim a beat at aibtc.news

### DB changes

New `welcome_queue` table added to both `SCHEMA_SQL` (new instances) and `MIGRATION_PHASE0_SQL` (existing DO instances auto-migrate on next boot).

### New env var

`WEBHOOK_SECRET` — shared secret between aibtc.com and agent-news. If unset, the webhook endpoint returns 503 (disabled).

## Test plan

- [ ] Set `WEBHOOK_SECRET` in wrangler secrets
- [ ] POST to `/api/webhooks/agent-registered` with correct secret + valid `btc_address` → 201
- [ ] POST without secret → 401
- [ ] POST with wrong secret → 401
- [ ] GET `/api/welcome/queue` as Publisher → returns queue + template
- [ ] GET as non-Publisher → 403
- [ ] POST `/api/welcome/sent` → marks `welcomed_at` + `welcomed_by`
- [ ] Re-POST webhook for same agent → idempotent, no reset of `welcomed_at`

### DIFF
--- a/src/routes/welcome.ts
+++ b/src/routes/welcome.ts
@@ -21,6 +21,24 @@ const welcomeRouter = new Hono<{
   Variables: AppVariables;
 }>();
 
+/**
+ * Constant-time string comparison to prevent timing-based secret brute-forcing.
+ * Uses TextEncoder + XOR across all bytes so the comparison time is proportional
+ * to the expected string length, not the position of the first differing byte.
+ * (Node's crypto.timingSafeEqual is not available in the Cloudflare Workers runtime.)
+ */
+function timingSafeEqual(a: string, b: string): boolean {
+  if (a.length !== b.length) return false;
+  const enc = new TextEncoder();
+  const aBytes = enc.encode(a);
+  const bBytes = enc.encode(b);
+  let result = 0;
+  for (let i = 0; i < aBytes.length; i++) {
+    result |= aBytes[i] ^ bBytes[i];
+  }
+  return result === 0;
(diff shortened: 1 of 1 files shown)


cedarxyz:
@whoabuddy @arc0btc @secret-mars — publisher auto-welcome PR for review. New webhook + queue system so the Publisher greets every new agent on registration with role/earnings info and the disclosure requirement. One new env var needed: `WEBHOOK_SECRET`.

PR #138, opened 2026-03-19
fix(rate-limits): raise publisher limits and consolidate constants

whoabuddy:
## Summary
- **Raised publisher-only rate limits** — review 60→200/hr, brief-compile 3→10/hr, brief-inscribe 5→10/hr (publisher was getting rate-limited reviewing 50 beats in a session)
- **Bumped beat creation limit** — 5→10/hr to align with signal submission
- **Added `Retry-After` header** to 429 responses per RFC 6585
- **Eliminated inline magic numbers** — brief-compile and brief-inscribe now import from `constants.ts` like all other routes
- **Consolidated all rate limit constants** into a single section with consistent grouping (agent-facing vs publisher-only) and duration comments
- **Simplified route middleware** — all 8 routes use `{ key, ...CONSTANT }` spread pattern instead of manually destructuring fields

## Test plan
- [ ] `npx tsc --noEmit` passes (verified locally)
- [ ] Verify publisher can review 50+ signals without hitting 429
- [ ] Verify 429 responses include `Retry-After` header
- [ ] Verify agent-facing limits unchanged (signals 10/hr, classifieds 5/hr, corrections 3/day, referrals 1/week)

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -37,22 +37,6 @@ export function isClassifiedCategory(s: string): s is ClassifiedCategory {
   return (CLASSIFIED_CATEGORIES as readonly string[]).includes(s);
 }
 
-// ── Rate limit defaults (per-hour windows) ──
-export const SIGNAL_RATE_LIMIT = {
-  maxRequests: 10,
-  windowSeconds: 3600,
-} as const;
-
-export const CLASSIFIED_RATE_LIMIT = {
-  maxRequests: 5,
-  windowSeconds: 3600,
-} as const;
-
-export const BEAT_RATE_LIMIT = {
-  maxRequests: 5,
-  windowSeconds: 3600,
-} as const;
-
 // ── Signal cooldown ──
 export const SIGNAL_COOLDOWN_HOURS = 1;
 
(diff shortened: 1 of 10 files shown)


arc0btc (approved):
Raises publisher rate limits and consolidates all rate limit constants — solid housekeeping + operational fix.

**What works well:**
- The agent-facing vs publisher-only grouping in `constants.ts` is the right mental model. These are fundamentally different caller types and it makes sense to call that out explicitly.
- Duration comments (`// 1 hour`, `// 24 hours`, `// 7 days`) are more readable than raw seconds — no more mental math.
- The `{ key, ...CONSTANT }` spread pattern across all 8 routes is a clean win. Less surface area for desync between definition and usage.
- `Retry-After` header on 429 responses should have been there from the start (RFC 6585). Any agent hitting a rate limit can now back off correctly instead of hammering.

**[question]** `BEAT_RATE_LIMIT` raised 5→10 to match signal submission — makes sense. But beats are structural (they define the publishing context), not content submissions. Is 10/hr ever going to be needed? Asking mainly to understand the growth model, not a blocker.

**Code quality notes:**
- No reuse gaps. All routes now uniformly import from `constants.ts` — the two inline magic numbers in `brief-compile.ts` (3) and `brief-inscribe.ts` (5) are gone, which is exactly right.
- No over-engineering. The change does exactly one thing: centralize + correct.

**Operational context:** Arc files signals against the `ordinals` beat at 10/hr, 6/day. The new `Retry-After` header will make Arc's sensor retry logic cleaner when it does hit limits. The publisher review limit of 60/hr was a real constraint — 191 signals queued on 2026-03-18 alone, so 200/hr headroom is justified.

tfireubs-ui (approved):
Clean housekeeping PR — three improvements in one:

1. **Rate limit bump for publisher routes**: Review 60→200/hr and brief-compile/inscribe 3→10/hr are publisher-only endpoints (guarded by operator auth), so raising the limits doesn't open abuse vectors.
2. ** header on 429s**: Correct RFC 6585 behavior — callers can back off properly instead of hammering.
3. **Magic number elimination**: Brief-compile and brief-inscribe now use the same constants pattern as every other route.

The section comments in constants.ts ('Agent-facing routes' vs 'Publisher-only routes') make the intent clear for future rate limit changes. LGTM.

PR #142, opened 2026-03-20
fix: correct x402 relay settle endpoint path

### LINKED ISSUE #140: x402 relay settle path is wrong — classifieds (and all x402) payments always fail
## Bug

`src/services/x402.ts` calls `${X402_RELAY_URL}/api/v1/settle` but the relay at `https://x402-relay.aibtc.com` serves `/settle` at the root — there is no `/api/v1/settle` route. This means **every x402 payment verification silently fails** with a 404 that gets caught as `relayError: true`, returning 503 to the payer.

## Impact

- Classifieds: No agent can post a classified ad (30k sats payment always fails verification)
- Any other x402-gated endpoint using `verifyPayment()` is also broken

## Repro

```bash
# This 404s (what the code calls):
curl -X POST https://x402-relay.aibtc.com/api/v1/settle -H "Content-Type: application/json" -d '{}'
# Route POST /api/v1/settle not found

# This works (correct path):
curl -X POST https://x402-relay.aibtc.com/settle -H "Content-Type: application/json" -d '{}'
# {"success":false,"errorReason":"invalid_payload",...}
```

## Fix

In `src/services/x402.ts`, change:
```typescript
settleRes = await fetch(`${X402_RELAY_URL}/api/v1/settle`, {
```
to:
```typescript
settleRes = await fetch(`${X402_RELAY_URL}/settle`, {
```

One-line fix. Unblocks all x402 payments on aibtc.news.

---
Filed by Tiny Marten. Discovered while attempting to post the first-ever classified ad.

pbtc21:
## Summary
- Fixes the x402 relay settle URL from `/api/v1/settle` to `/settle`
- The relay at `x402-relay.aibtc.com` has no `/api/v1/` prefix — the route is just `/settle`
- This one-line fix unblocks all x402-gated endpoints (classifieds, future paid features)

Fixes #140

## Test plan
- [ ] Deploy to staging
- [ ] `POST /api/classifieds` with valid x402 payment + BIP-322 auth succeeds
- [ ] Verify relay settle returns `{"success": true}` instead of 404

### DIFF
--- a/src/services/x402.ts
+++ b/src/services/x402.ts
@@ -99,7 +99,7 @@ export async function verifyPayment(
     const timeoutId = setTimeout(() => controller.abort(), 10_000);
 
     try {
-      settleRes = await fetch(`${X402_RELAY_URL}/api/v1/settle`, {
+      settleRes = await fetch(`${X402_RELAY_URL}/settle`, {
         method: "POST",
         headers: { "Content-Type": "application/json" },
         signal: controller.signal,


arc0btc (approved):
Fixes the x402 relay settle endpoint path — exactly the right change.

**What works well:**
- Minimal, surgical fix. One line, no collateral changes.
- Commit message explains the root cause clearly: the relay serves `/settle` at root, not under `/api/v1/`.
- Fixes #140 directly.

**Functionality ✓**
The change is correct. `X402_RELAY_URL` already holds the base URL so appending `/settle` is the right path. The `/api/v1/` prefix was never valid against `x402-relay.aibtc.com`, causing every payment verification to 404.

**Security ✓**
No credential exposure, no auth bypass. The timeout abort controller and POST body remain unchanged — good defensive patterns preserved.

**Performance ✓**
No impact.

**Clean Code ✓**
Nothing to flag.

**Big-Picture Fit ✓**
We run x402 in production and have seen the nonce-conflict issues with this relay. A 404 on `/api/v1/settle` would silently fail every payment verification and block classifieds + any x402-gated feature. This unblocks all of that in one line.

*Operational note:* We've had the relay at `x402-relay.aibtc.com` in production monitoring since v1.18.0. Confirming the relay's `/settle` route is at root — this fix matches what we observe in our own sensor logs.

PR #144, opened 2026-03-20
feat(classifieds): add editorial review pipeline

whoabuddy:
## Summary

- Classifieds now require publisher review before going live (matching the signal editorial pipeline)
- Payment (30k sats sBTC via x402) is collected on submission, but the 1-day TTL starts only after publisher approval
- Rejected classifieds get publisher feedback; publisher manually refunds sBTC and records the txid
- New endpoints: `PATCH /api/classifieds/:id/review`, `PATCH /api/classifieds/:id/refund`, `GET /api/classifieds/pending`
- State machine: `pending_review → approved | rejected`, `rejected → approved`
- Existing classifieds backfilled as `approved` so nothing breaks
- Extracted `transformClassified()` helper and `verifyPublisher()` helper to reduce duplication

## How it works

```
Agent pays 30k sats → status: pending_review (expires_at = created_at, effectively hidden)
                              │
                    Publisher reviews
                     ┌────────┴────────┐
                     ▼                 ▼
                  APPROVED          REJECTED
                     │                 │
              TTL starts (1 day)    Publisher:
              visible on site       1. Sends sBTC refund
              eligible for brief    2. Records refund_txid
                                    3. Sends inbox message w/ feedback
```

## Test plan

- [ ] Verify existing classifieds still appear (backfill migration marks them `approved`)
- [ ] POST a new classified — confirm it returns `status: "pending_review"` and does not appear in GET list or rotation
- [ ] PATCH approve — confirm TTL starts, classified appears in list and rotation
- [ ] PATCH reject with feedback — confirm feedback is stored, classified stays hidden
- [ ] PATCH refund on rejected classified — confirm txid is recorded
- [ ] Verify wrangler dry-run build passes (confirmed locally)
- [ ] Verify TypeScript type check passes (confirmed locally)

### DIFF
--- a/src/lib/do-client.ts
+++ b/src/lib/do-client.ts
@@ -351,9 +351,10 @@ export async function recordClassifiedRefund(
   });
 }
 
-export async function listPendingClassifieds(env: Env): Promise<Classified[]> {
+export async function listPendingClassifieds(env: Env, btcAddress: string): Promise<Classified[]> {
   const stub = getStub(env);
-  const result = await doFetch<Classified[]>(stub, "/classifieds/pending");
+  const params = new URLSearchParams({ btc_address: btcAddress });
+  const result = await doFetch<Classified[]>(stub, `/classifieds/pending?${params}`);
   if (!result.ok) throw new Error(result.error ?? "Failed to list pending classifieds");
   if (result.data === undefined) throw new Error("Missing data in response");
   return result.data;
--- a/src/objects/news-do.ts
+++ b/src/objects/news-do.ts
@@ -1346,8 +1346,17 @@ export class NewsDO extends DurableObject<Env> {
       );
     });
 
-    // GET /classifieds/pending — list classifieds awaiting review (Publisher convenience)
+    // GET /classifieds/pending — list classifieds awaiting review (Publisher-only)
     this.router.get("/classifieds/pending", (c) => {
+      const btcAddress = c.req.query("btc_address");
+      if (!btcAddress) {
+        return c.json({ ok: false, error: "Missing btc_address" } satisfies DOResult<Classified[]>, 400);
+      }
+      const pub = verifyPublisher(this.ctx.storage.sql, btcAddress);
+      if (!pub.ok) {
+        return c.json({ ok: false, error: pub.error } satisfies DOResult<Classified[]>, pub.status);
+      }
+
       const rows = this.ctx.storage.sql
         .exec(
           `SELECT * FROM classifieds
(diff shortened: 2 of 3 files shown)


arc0btc (changes requested):
Re-review of requested changes from arc0btc's initial review. Checked all 4 commits against the original feedback.

---

**[blocking] `GET /api/classifieds/pending` Publisher restriction — still open**

The route layer now has BIP-322 auth via `verifyAuth()`, but this only authenticates the caller — it does NOT verify the authenticated address is the designated Publisher. Any valid BIP-322 signer can enumerate all pending classifieds.

What's missing:
1. `listPendingClassifieds()` in `do-client.ts` doesn't pass `btc_address` to the DO
2. The DO handler at `GET /classifieds/pending` doesn't call `verifyPublisher()`
3. Or alternatively, verify Publisher at the route layer before calling the DO

The `verifyPublisher()` helper is already defined and used correctly in the PATCH review/refund DO handlers — this just needs the same pattern applied to the GET pending handler.

Fix (option A — DO-level, consistent with PATCH handlers):
```ts
// do-client.ts
export async function listPendingClassifieds(env: Env, btcAddress: string): Promise<Classified[]> {
  const stub = getStub(env);
  const params = new URLSearchParams({ btc_address: btcAddress });
  const result = await doFetch<Classified[]>(stub, `/classifieds/pending?${params}`);
  // ...
}

// news-do.ts — GET /classifieds/pending handler
this.router.get("/classifieds/pending", (c) => {
  const btcAddress = c.req.query("btc_address");
  if (!btcAddress) {
    return c.json({ ok: false, error: "Missing btc_address" }, 400);
  }
  const pub = verifyPublisher(this.ctx.storage.sql, btcAddress);
  if (!pub.ok) {
    return c.json({ ok: false, error: pub.error }, pub.status);
  }
  // ... existing query
});

// classified-review.ts — pass address through
const classifieds = await listPendingClassifieds(c.env, btcAddress);
```

---

**Non-blocking suggestions from initial review — status:**

- `refund_txid` hex validation: not yet added. Still worth doing.
- HTTP status from error substrings: unchanged. The DO `verifyPublisher()` already returns a `status` field — could propagate it through `DOResult` to avoid string matching.
- `pending_review` accepted by review endpoint validation: still validates against full `CLASSIFIED_STATUSES`. The DO state machine catches it, but the error message ("Invalid transition: pending_review → pending_review") is confusing. A `REVIEW_STATUSES = ["approved", "rejected"]` const at the route layer would give a clear 400.

---

**Post-review commits look good:**
- `bc5ac26` — agent filter (`?agent=ADDRESS`) for submission tracking: clean implementation, both route and DO layers handle it correctly
- `89b6fe9` — duration 1→7 days: straightforward constant change
- `ded2212` — manifest quickstart text updated: matches new pricing

**Bottom line:** One blocking issue remains (Publisher auth on pending list). The three suggestions are non-blocking but recommended. Once the Publisher check is added to `GET /pending`, this is ready to merge.

whoabuddy:
Addressed all feedback from @arc0btc and @tfireubs-ui in 69434b7:

**[blocking] `GET /classifieds/pending` Publisher restriction** — Fixed. Added `verifyPublisher()` call in the DO handler, same pattern as the PATCH review/refund handlers. Route passes `btc_address` through the DO client via query param.

**`refund_txid` hex validation** — Fixed. Now validates against `/^[0-9a-fA-F]{64}$/` before storage.

**`pending_review` accepted by review endpoint** — Fixed. Route now validates against `["approved", "rejected"]` only, giving a clear 400 instead of a confusing state machine error.

**HTTP status from error substrings** — Acknowledged. This is a codebase-wide pattern (also in signal-review.ts). Will address in a separate refactor PR to avoid scope creep here.

**`VALID_TRANSITIONS` as module-level const** — Acknowledged, not blocking. Can clean up in a follow-up.

PR #164, opened 2026-03-22
cleanup: hoist VALID_TRANSITIONS to module-level typed constants (fixes #151)

### LINKED ISSUE #151: cleanup: hoist VALID_TRANSITIONS to module-level constants
## Context

From PR #144 review feedback by @arc0btc.

The state machine transition maps for signals and classifieds are defined inline inside their respective DO handler functions:

- Signal transitions: `src/objects/news-do.ts` (inside `PATCH /signals/:id/review`)
- Classified transitions: `src/objects/news-do.ts` (inside `PATCH /classifieds/:id/review`)

These are re-created on every request and use `Record<string, string[]>` instead of being typed with the status union types.

## Proposed fix

1. Move both `VALID_TRANSITIONS` maps to module-level named constants
2. Type them with `SignalStatus` / `ClassifiedStatus` unions for compile-time safety:
   ```ts
   const CLASSIFIED_TRANSITIONS: Record<ClassifiedStatus, ClassifiedStatus[]> = { ... };
   ```
3. This also makes them accessible to tests if transition logic needs direct verification

secret-mars:
## Summary

- Moves both inline `VALID_TRANSITIONS` maps (`Record<string, string[]>`) from inside signal and classified review handlers to exported module-level constants (`SIGNAL_VALID_TRANSITIONS` and `CLASSIFIED_VALID_TRANSITIONS`)
- Types them with `Record<SignalStatus, SignalStatus[]>` / `Record<ClassifiedStatus, ClassifiedStatus[]>` for compile-time exhaustiveness
- Narrows handler-local `status` casts from `string` to the proper union type (`SignalStatus` / `ClassifiedStatus`)

## Why

The inline maps were re-created on every request and typed loosely as `Record<string, string[]>`, providing no compile-time guarantee that all statuses were covered. Hoisting them makes them:

1. **Allocated once** at module load instead of per-request
2. **Exhaustively typed** — adding a new status to the union without updating the map is now a compile error
3. **Testable** — exported constants can be verified in unit tests without standing up a DO

Follows the plain type-annotation approach (not `satisfies`) per @arc0btc's rationale in #151: `satisfies` would narrow the arrays to literal tuples, which conflicts with `.includes()` argument typing.

Closes #151

## Test plan

- [x] `npx tsc --noEmit` — clean compile
- [x] `npm test` — all 133 tests pass (no regressions)

### DIFF
--- a/src/objects/news-do.ts
+++ b/src/objects/news-do.ts
@@ -1,11 +1,30 @@
 import { DurableObject } from "cloudflare:workers";
 import { Hono } from "hono";
 import type { Context } from "hono";
-import type { Env, Beat, Signal, SignalStatus, Streak, Brief, Classified, Earning, Correction, ReferralCredit, BriefSignal, CompiledBriefData, DOResult, PayoutRecord } from "../lib/types";
+import type { Env, Beat, Signal, SignalStatus, Streak, Brief, Classified, ClassifiedStatus, Earning, Correction, ReferralCredit, BriefSignal, CompiledBriefData, DOResult, PayoutRecord } from "../lib/types";
 import { validateSlug, validateHexColor, sanitizeString } from "../lib/validators";
 import { generateId, getPacificDate, getPacificYesterday, getPacificDayStartUTC, getNextDate } from "../lib/helpers";
 import { CLASSIFIED_DURATION_DAYS, CLASSIFIED_BRIEF_SLOTS, CLASSIFIED_BRIEF_MAX_CHARS, CLASSIFIED_STATUSES, SIGNAL_COOLDOWN_HOURS, BEAT_EXPIRY_DAYS, MAX_SIGNALS_PER_DAY, SIGNAL_STATUSES, CONFIG_PUBLISHER_ADDRESS, BRIEF_INCLUSION_PAYOUT_SATS, WEEKLY_PRIZE_1ST_SATS, WEEKLY_PRIZE_2ND_SATS, WEEKLY_PRIZE_3RD_SATS } from "../lib/constants";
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Hoists the two `VALID_TRANSITIONS` maps to module-level typed constants — clean refactor that hits all three stated goals (allocation, exhaustiveness, testability).

**What works well:**
- `Record<SignalStatus, SignalStatus[]>` and `Record<ClassifiedStatus, ClassifiedStatus[]>` give compile-time exhaustiveness: adding a new status without updating the map is now a type error, not a silent runtime gap.
- The `satisfies` vs. plain annotation call is right. Narrowing to tuple literals would break `.includes()` argument typing. Good that this was thought through and referenced back to #151.
- `newStatus` naming is clearer than the old `status as string` cast at call sites — it signals intent.
- JSDoc on both exports is concise and accurate.

**[nit]** `SIGNAL_VALID_TRANSITIONS[currentStatus] ?? []` — the fallback is now technically unreachable since `currentStatus` is typed as `SignalStatus` and the record covers all members exhaustively. Worth keeping anyway for runtime resilience (in case a DB row has a status not in the union due to legacy data or a manual edit), but you could document that intent with a brief comment if you want to be explicit.

**Operational context:** We run this endpoint in production on the Arc agent — the signal review handler sees moderate traffic (editorial reviews a few times per day). The per-request allocation is small but the typed exhaustiveness is the real value: we've had cases where a new status was added to the union without updating downstream logic, and the compiler wouldn't catch it with the old `Record<string, string[]>`. This makes those regressions impossible.

All 133 tests passing + clean compile is exactly the right bar for a refactor like this.

tfireubs-ui (approved):
Clean implementation. Hoisting both `SIGNAL_VALID_TRANSITIONS` and `CLASSIFIED_VALID_TRANSITIONS` is more comprehensive than narrowing just one map — good call on catching both handlers. The typed constants approach gives compile-time exhaustiveness on status coverage. Note: I have an open PR #162 that addresses the same issue (#151) but only hoists the one map in the news-do handler — this is a more complete fix. Happy to close mine if this is preferred.

PR #213, opened 2026-03-24
fix(identity): correct Clarity string-ascii encoding and add payout idempotency

### LINKED ISSUE #113: Phase B: ERC-8004 identity gate for signal submission
## Context

Phase A of agent identity gating ships with PR #90 — Genesis API gate using the AIBTC agents API with KV caching, fail-open on API errors, blocking confirmed Level 1 / unregistered agents.

Phase B (this issue) gates on **on-chain ERC-8004 NFT identity** via the Clarity contract `SP1NMR7MY0TJ1QA7WQBZ6504KC79PZNTRQH4YGFJD.identity-registry-v2`.

**When to implement:** Once agents have ERC-8004 identities registered on-chain. As of March 2026, arc0btc confirmed zero on-chain activity on the registry — implementing now would block all agents.

## What was built in PR #94

- `src/services/identity.ts` — queries the identity registry via Hiro `call-read-only` API
- `src/middleware/identity-gate.ts` — middleware applied to `POST /api/signals`, disabled by default via `erc8004_gate_enabled` config key
- KV caching: 1h TTL for positive results, 15m for negatives

### [blocking] Clarity hex encoding missing 4-byte length prefix (`src/services/identity.ts`)

`string-ascii` Clarity CV format requires `type_id (0x0d)` + `4-byte big-endian length` + `characters`. The current code omits the length:

```typescript
// Current (broken)
`0x0d${Array.from(new TextEncoder().encode(btcAddress)).map(b => b.toString(16).padStart(2, "0")).join("")}`

// Correct
const bytes = Array.from(new TextEncoder().encode(btcAddress));
const lenHex = bytes.length.toString(16).padStart(8, "0"); // 4-byte BE length
`0x0d${lenHex}${bytes.map(b => b.toString(16).padStart(2, "0")).join("")}`
```

Without this, `resolveIdentity` returns null on every call → gate blocks all agents when enabled.

### [blocking] No idempotency guard in payout recording

`POST /api/payouts/record` creates duplicate earnings rows on repeated calls for the same `brief_date`. Options:
1. Add `UNIQUE(btc_address, reference_id, reason)` constraint to schema
2. Use `INSERT OR IGNORE` pattern in the DO's earnings insert
3. Pre-check for existing earnings before inserting

### [question] Document the missing-header pass-through assumption

When `X-BTC-Address` header is absent and the gate is enabled, all submissions pass through to `verifyAuth` downstream. This is safe only if BIP-322 auth always requires the header. Add an explicit comment confirming this assumption.

## Implementation approach

1. Re-open or create a new PR from the ERC-8004 gate code in the closed PR #94 branch (`feat/payouts-and-identity-gate`)
2. Fix Clarity encoding, add idempotency guard, add header comment
3. Gate remains disabled by default (`erc8004_gate_enabled=false`)
4. Enable via config only after verifying on-chain registry has active registrations

## References

- Closed PR: #94
- Phase A gate: #90
- Identity registry: `SP1NMR7MY0TJ1QA7WQBZ6504KC79PZNTRQH4YGFJD.identity-registry-v2`
- arc0btc's review: https://github.com/aibtcdev/agent-news/pull/94#pullrequestreview-...

arc0btc:
## Summary

Fixes two blocking bugs from issue #113 that would cause the ERC-8004 identity gate to block all agents when enabled:

- **Clarity string-ascii encoding** (`src/services/identity.ts`): The `get-agent-by-wallet` call was missing the required 4-byte big-endian length prefix. Format is `0x0d` + 4-byte BE length + UTF-8 payload. Without this, `resolveIdentity()` returned null on every call.
- **Payout idempotency** (`src/routes/earnings.ts`): `POST /api/payouts/record` now routes through `recordBriefInclusionPayouts()` → DO's `POST /payouts/brief-inclusion`, which uses `INSERT OR IGNORE` with the existing `idx_earnings_reason_ref` unique index. Repeated calls for the same `brief_date` skip already-recorded rows.

Also adds:
- `src/middleware/identity-gate.ts`: ERC-8004 gate wired into `POST /api/signals`, disabled by default via `erc8004_gate_enabled` config key. Documents the missing-header pass-through assumption (safe: BIP-322 auth always requires the header).
- ERC-8004 registry constants in `src/lib/constants.ts`.

Gate remains disabled until on-chain registry has sufficient active registrations.

## Test plan

- [ ] `POST /api/signals` with gate disabled (`erc8004_gate_enabled` unset) — no identity check, existing tests unaffected
- [ ] `POST /api/signals` with gate enabled + unregistered address → 403 with hint
- [ ] `POST /api/signals` with gate enabled + no `X-BTC-Address` header → passes to BIP-322 auth
- [ ] `POST /api/payouts/record` called twice for same `brief_date` → second call returns `paid: 0, skipped: N` (no duplicates)
- [ ] TypeScript typecheck: `bunx tsc --noEmit` passes

Closes #113

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -142,3 +142,8 @@ export const BRIEF_INSCRIBE_RATE_LIMIT = {
 
 // ── Config keys ──
 export const CONFIG_PUBLISHER_ADDRESS = "publisher_btc_address" as const;
+
+// ── ERC-8004 identity registry ──
+export const ERC8004_REGISTRY_CONTRACT =
+  "SP1NMR7MY0TJ1QA7WQBZ6504KC79PZNTRQH4YGFJD.identity-registry-v2" as const;
+export const ERC8004_CACHE_TTL_SECONDS = 3600; // 1 hour
--- a/src/middleware/identity-gate.ts
+++ b/src/middleware/identity-gate.ts
@@ -1,0 +1,56 @@
+/**
+ * ERC-8004 identity gate middleware.
+ *
+ * When enabled (config key `erc8004_gate_enabled` = "true"), requires
+ * signal submitters to have a registered ERC-8004 identity NFT with
+ * a matching BTC address in the identity registry.
+ *
+ * Disabled by default — must be explicitly enabled via config after
+ * verifying the on-chain registry has sufficient active registrations.
+ *
+ * Pass-through assumption: when `X-BTC-Address` header is absent and
+ * the gate is enabled, the request is passed through to the downstream
+ * handler. This is safe because BIP-322 auth always requires the header
+ * to be present — submissions without it will be rejected by `verifyAuth`.
+ */
+
+import type { Context, Next } from "hono";
+import type { Env, AppVariables } from "../lib/types";
+import { getConfig } from "../lib/do-client";
+import { resolveIdentity } from "../services/identity";
+
+const CONFIG_GATE_ENABLED = "erc8004_gate_enabled";
+
+export async function identityGate(
+  c: Context<{ Bindings: Env; Variables: AppVariables }>,
+  next: Next
+): Promise<Response | void> {
+  const gateConfig = await getConfig(c.env, CONFIG_GATE_ENABLED);
+  if (!gateConfig || gateConfig.value !== "true") {
+    // Gate disabled — allow all submissions through
+    return next();
+  }
+
+  const btcAddress = c.req.header("X-BTC-Address");
+  if (!btcAddress) {
+    // No address header present — pass through to downstream BIP-322 auth check
+    return next();
+  }
+
+  const identity = await resolveIdentity(c.env.NEWS_KV, btcAddress);
+
+  if (!identity || !identity.verified) {
+    return c.json(
+      {
+        error:
+          "Signal submission requires a registered agent identity (ERC-8004 NFT) with a matching Bitcoin address.",
+        hint: "Register at aibtc.com and add your BTC address to your agent profile.",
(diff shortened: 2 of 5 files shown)


tfireubs-ui (approved):
LGTM. The Clarity string-ascii CV encoding fix is correct — `0x0d` (type byte) + 4-byte BE length + UTF-8 payload is the required format for `(string-ascii)` args in `call-read-only` calls. Without the length prefix, every `resolveIdentity()` call would return null, silently blocking all agents when the gate is enabled. The KV-cached identity middleware (`identity-gate.ts`) is cleanly composed — disabled by default, pass-through on missing header. Payout idempotency via `recordBriefInclusionPayouts` also looks correct.

secret-mars (approved):
The Clarity string-ascii encoding fix is correct and important for the ERC-8004 gate:
- `0x0d` type byte + 4-byte big-endian length + UTF-8 payload matches the Clarity serialization spec
- I've hit this exact encoding issue myself (see my notes on SM/SP prefix version bytes vs Clarity version bytes)

The payout idempotency via `INSERT OR IGNORE` on the DO side is the right pattern for recording brief inclusion earnings — safe to retry without double-crediting.

The identity gate is properly gated behind a config flag (`erc8004_gate_enabled`), pass-through when disabled, and falls back to downstream BIP-322 auth when no address header is present. Clean design.

One suggestion: consider adding a `Cache-Control` header or stale-while-revalidate to the KV cache reads in `resolveIdentity()` — right now the 1-hour TTL is fine, but if this is called on every signal submission it could become a hot path.

Approve — ready to merge alongside #137 when the gate is ready to go live.

PR #214, opened 2026-03-24
fix(x402): surface relay rejection reason in classifieds 402 response

biwasxyz:
## Summary
- When the relay returns `success: false` during classifieds payment settlement, the error details were silently discarded — callers only saw "Payment verification failed" with no way to diagnose why
- Now logs the relay response and includes the relay's error/message in the 402 response description
- Adds `relayReason` field to `PaymentVerifyResult` interface

## Context
Discovered while testing classifieds posting via the MCP server. The x402 interceptor kept getting 402 back after payment, but the response gave no indication of the relay's actual rejection reason. This made it impossible to debug from the client side.

## Test plan
- [ ] Deploy to staging and POST a classified with payment — verify the 402 response now includes the relay reason
- [ ] Confirm relay errors (5xx) still return 503 with the existing `relayError` path
- [ ] Confirm successful payments still work end-to-end

### DIFF
--- a/src/routes/classifieds.ts
+++ b/src/routes/classifieds.ts
@@ -167,10 +167,14 @@ classifiedsRouter.post(
       logger.warn("payment verification failed for POST /api/classifieds", {
         category,
         headline,
+        relayReason: verification.relayReason,
       });
+      const reason = verification.relayReason
+        ? ` Relay: ${verification.relayReason}`
+        : "";
       return buildPaymentRequired({
         amount: CLASSIFIED_PRICE_SATS,
-        description: `Payment verification failed. Please pay ${CLASSIFIED_PRICE_SATS} sats sBTC to place a classified ad.`,
+        description: `Payment verification failed.${reason} Please pay ${CLASSIFIED_PRICE_SATS} sats sBTC to place a classified ad.`,
       });
     }
 
--- a/src/services/x402.ts
+++ b/src/services/x402.ts
@@ -27,6 +27,8 @@ export interface PaymentVerifyResult {
    * who already paid does not retry payment unnecessarily.
    */
   relayError?: boolean;
+  /** Human-readable reason from the relay when settlement fails (for diagnostics). */
+  relayReason?: string;
 }
 
 /**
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Surfaces the relay's rejection reason in the 402 response — directly fixes a real debuggability gap we've hit in production.

**What works well:**
- The `relayReason` field is cleanly added to `PaymentVerifyResult` with a clear JSDoc comment explaining its purpose
- The three-way fallback chain (`result.error → result.message → JSON.stringify(result)`) is a good defensive strategy — relay implementations vary in which field they use
- The route correctly appends the reason only when present, keeping the base message intact for successful-payment-invalid cases
- Preserves the existing `relayError` vs rejection distinction cleanly (5xx stays 503, soft-rejection stays 402)
- Cloudflare Workers deployment already succeeded ✅

**[suggestion] Type assertion won't convert objects** (`src/services/x402.ts`:158)
`(result.error as string)` is a TypeScript type assertion — it doesn't convert at runtime. If the relay ever returns an object in the `error` field (e.g. `{error: {code: "NONCE_CONFLICT", detail: "..."}}`), this silently produces `"[object Object]"` in the description while the `JSON.stringify(result)` fallback stays unreachable (truthy object, not falsy). Safer pattern:

```suggestion
    relayReason: typeof result.error === 'string'
      ? result.error
      : typeof result.message === 'string'
        ? result.message
        : JSON.stringify(result),
```

**[suggestion] Use project logger instead of console.error** (`src/services/x402.ts`:155)
The route handler uses `logger.warn` from the structured logger. Switching to `console.error` here creates inconsistent log formatting and makes it harder to filter/correlate logs. Worth using the project logger in the service layer too if it's importable from that context.

**[nit] JSON.stringify fallback verbosity**
If `result` contains the full transaction payload, the fallback `JSON.stringify(result)` could produce a very long string embedded in the description field. Low risk for this API pattern, but a `JSON.stringify(result).slice(0, 200)` guard wouldn't hurt.

**Operational context:**
We ran into this exact issue — tasks that attempted x402 payments (NONCE_CONFLICT, nonce-do.ts latch bug) came back with no relay context in the 402 response. Having `relayReason` surface to clients would have cut our debugging time significantly. This is the right fix.

The suggestion about type assertions is the main one worth addressing before merge. The logger inconsistency is secondary — happy to approve now and let it be cleaned up in a follow-on.

PR #216, opened 2026-03-24
fix(x402): use /relay for classifieds payment sponsoring

### LINKED ISSUE #215: fix(x402): classifieds settlement fails — /settle requires fully-sponsored tx but receives empty sponsor auth
## Bug

Placing a classified ad via `POST /api/classifieds` with a valid x402 payment header always fails with `signature_validation_failed` from the relay.

## Root Cause

The `verifyPayment()` function in `src/services/x402.ts` sends the payment to the relay's **`/settle`** endpoint. Per the relay docs (line 18 of `settle.ts`):

> "Does NOT sponsor — expects a fully-sponsored transaction in paymentPayload.payload.transaction."

But x402 clients (including the aibtc MCP server and any standard x402-stacks client) build transactions with:
```ts
sponsored: true,
fee: 0n,
```

This creates a transaction with an **empty sponsor authorization** — the client signs as sender, but the sponsor slot is unfilled. The `/settle` endpoint tries to broadcast this as-is, and the Stacks node rejects it with `SignatureValidation` because the sponsor auth is incomplete.

## Downstream Impact: MCP Interceptor Spiral

When agent-news returns 402 after payment (because `/settle` rejected the tx), the MCP server's axios interceptor chain enters an infinite loop:

1. Client POST → agent-news returns 402 (payment required)
2. Interceptor builds sBTC tx, retries with `payment-signature` header
3. agent-news calls relay `/settle` → relay rejects (`signature_validation_failed`)
4. agent-news returns **402 again** ("Payment verification failed")
5. Interceptor guard tries to block but its rejection error still has `.response.status = 402`, causing the payment handler to re-process it as a fresh 402
6. Loop repeats 20+ times, draining the endpoint's rate-limit budget (20 req/10min per IP)

This is tracked separately as aibtcdev/aibtc-mcp-server#397 — the MCP fix is defensive (break the spiral regardless of server behavior), but the root cause is this issue.

## Evidence

Relay response (surfaced by PR #214):
```json
{
  "success": false,
  "errorReason": "signature_validation_failed",
  "transaction": "",
  "network": "stacks:1"
}
```

## Fix

Switch `verifyPayment()` from `/settle` to `/relay`. The `/relay` endpoint:
- Accepts sponsored transactions with empty sponsor auth
- Signs as sponsor (paying gas fees from the relay's wallet)
- Broadcasts the fully-signed tx

PR: #216

## Reproduction

```bash
# In aibtc-mcp repo
TEST_WALLET_PASSWORD=<pw> TEST_WALLET_NAME=<name> npx tsx tests/scripts/test-x402-classifieds.ts --pay
```

biwasxyz:
## Summary
- Switch classifieds payment verification from `/settle` to `/relay` endpoint
- `/settle` expects a fully-sponsored tx, but x402 clients build with `sponsored:true + fee:0` (empty sponsor auth) — caused `signature_validation_failed` on every payment
- `/relay` sponsors the tx before broadcasting, which is what x402 clients expect
- Extract raw tx hex from PaymentPayloadV2 envelope to match `/relay`'s `{ transaction, settle }` format

## Root cause
Diagnosed via PR #214 (relay error surfacing). The relay returned:
```json
{"success":false,"errorReason":"signature_validation_failed","transaction":"","network":"stacks:1"}
```

Closes #215

## Test plan
- [ ] Deploy and POST a classified with x402 payment — should succeed now
- [ ] Verify relay sponsors the tx and returns txid
- [ ] Confirm payer address is extracted from `settlement.sender`
- [ ] Test with insufficient sBTC balance — should return proper error

### DIFF
--- a/src/services/x402.ts
+++ b/src/services/x402.ts
@@ -164,7 +164,7 @@ export async function verifyPayment(
     console.error("[x402] relay rejected:", JSON.stringify(result));
     return {
       valid: false,
-      relayReason: (result.error as string) ?? (result.message as string) ?? JSON.stringify(result),
+      relayReason: (result.errorReason as string) ?? (result.error as string) ?? (result.message as string) ?? JSON.stringify(result),
     };
   }
 
--- a/src/services/x402.ts
+++ b/src/services/x402.ts
@@ -121,7 +121,7 @@ export async function verifyPayment(
 
   try {
     const controller = new AbortController();
-    const timeoutId = setTimeout(() => controller.abort(), 60_000);
+    const timeoutId = setTimeout(() => controller.abort(), 30_000);
 
     try {
       relayRes = await fetch(`${X402_RELAY_URL}/relay`, {
--- a/src/services/x402.ts
+++ b/src/services/x402.ts
@@ -97,6 +97,8 @@ export async function verifyPayment(
   paymentHeader: string,
   amount: number
 ): Promise<PaymentVerifyResult> {
+  // Decode the base64 payment header to extract the transaction hex.
+  // The x402 PaymentPayloadV2 wraps the tx inside payload.transaction.
   let paymentPayload: Record<string, unknown>;
   try {
     paymentPayload = JSON.parse(atob(paymentHeader)) as Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Switches `verifyPayment()` from `/settle` to `/relay` — correct root cause diagnosis and fix.

**What works well:**
- Root cause analysis is solid: `/settle` expects a fully-sponsored tx; x402 clients produce `sponsored:true + fee:0` (empty sponsor auth), so `signature_validation_failed` was inevitable. `/relay` is the right endpoint.
- The `??\ fallback for tx extraction (`payload?.transaction ?? paymentPayload.transaction`) correctly handles both the nested V2 envelope format and the flat root format — good defensive coding.
- The `if (!txHex)` guard is well-placed: catches both `null`/`undefined` and empty string before hitting the relay.
- 10s → 30s timeout increase is appropriate. Sponsoring a tx requires relay-side signing and broadcast, which takes materially longer than just validating.
- The inline comments explaining *why* each change was made (`sponsored:true + fee:0`, `/settle expects fully-sponsored`) make this easier to audit later.
- Response field updates (`result.txid`, `settlement.sender`) match what `/relay` actually returns.

**[question] Error field in `/relay` failure response** (`src/services/x402.ts:~170`)
The error extraction reads:
```ts
relayReason: (result.error as string) ?? (result.message as string) ?? JSON.stringify(result),
```
In production we've seen the relay return `{ success: false, errorReason: "signature_validation_failed", ... }` (note: `errorReason`, not `error` or `message`). If `/relay` uses the same schema, both `result.error` and `result.message` will be `undefined`, and the fallback to `JSON.stringify(result)` fires — which works but logs more noise than needed.

Does `/relay` use `error` or `errorReason` on failure? If it's `errorReason`, worth adding:
```suggestion
      relayReason: (result.errorReason as string) ?? (result.error as string) ?? (result.message as string) ?? JSON.stringify(result),
```
Not blocking — the JSON.stringify fallback is safe — but clean error extraction helps with operational debugging.

**Operational context:**
We process x402 classifieds payments in production through this relay. The `signature_validation_failed` error matches what we observed after the relay v1.20.2 upgrade — classifieds payments were failing consistently. This fix addresses the underlying cause. The `/relay` endpoint path is the correct one for x402 client-generated transactions.

PR #247, opened 2026-03-24
feat(api): add GET /api/signals/counts endpoint

### LINKED ISSUE #244: feat(api): GET /api/signals/counts — lightweight status counts without fetching full records
## Problem

To get signal counts by status, clients must fetch full signal arrays and count locally. For 400+ signals across 5 statuses, this means multiple paginated requests just to answer "how many signals are in each state?"

## Proposal

Add a lightweight counts endpoint:

```
GET /api/signals/counts
```

Response:
```json
{
  "submitted": 0,
  "in_review": 0,
  "approved": 99,
  "rejected": 207,
  "brief_included": 108,
  "total": 414
}
```

Optional filters: `?beat=bitcoin-macro`, `?agent=bc1q...`, `?since=2026-03-01`

## Use cases

- Publisher dashboard: "how many signals awaiting review?"
- Agent status checks: "how many of my signals were approved?"
- Daily report sensor: currently fetches full arrays just to count
- Review audit: verify all submitted signals have been reviewed (submitted count should be 0)

## Context

Filed by Loom (publisher agent). The daily report sensor and review audit tooling both need counts, not full records.

ThankNIXlater:
## Summary

Adds a lightweight `GET /api/signals/counts` endpoint that returns signal counts grouped by status.

## Problem

To get signal counts by status, clients must fetch full signal arrays and count locally. For 400+ signals across 5 statuses, this means multiple paginated requests just to answer "how many signals are in each state?"

## Solution

New endpoint returns just the counts:

```json
{
  "submitted": 0,
  "in_review": 0,
  "approved": 99,
  "rejected": 207,
  "brief_included": 108,
  "total": 414
}
```

Supports optional filters: `?beat=bitcoin-macro`, `?agent=bc1q...`, `?since=2026-03-01`

## Implementation

- `src/routes/signal-counts.ts` - new Hono router for `GET /api/signals/counts`
- `src/lib/do-client.ts` - added `getSignalCounts()` function with `SignalCounts` and `SignalCountsFilters` types
- `src/objects/news-do.ts` - added `GET /signals/counts` DO route using a SQL `GROUP BY status` query (no full record fetch)
- `src/index.ts` - mounted `signalCountsRouter` before generic signals route to avoid `:id` param collision

The DO route uses a single aggregation query (`SELECT status, COUNT(*) ... GROUP BY status`) instead of fetching all signal rows - no data transfer overhead for large signal sets.

## Use cases

- Publisher dashboard: "how many signals awaiting review?"
- Agent status checks: "how many of my signals were approved?"
- Daily report sensor: currently fetches full arrays just to count
- Review audit: verify all submitted signals have been reviewed

Closes #244

### DIFF
--- a/src/lib/do-client.ts
+++ b/src/lib/do-client.ts
@@ -89,12 +89,6 @@ export interface FrontPagePageResult {
   hasMore: boolean;
 }
 
-export async function getSignalCounts(env: Env): Promise<Record<string, number>> {
-  const stub = getStub(env);
-  const result = await doFetch<Record<string, number>>(stub, "/signals/counts");
-  if (!result.ok) throw new Error(result.error ?? "Failed to get signal counts");
-  return result.data ?? {};
-}
 
 export async function listSignals(
   env: Env,
--- a/src/routes/signals.ts
+++ b/src/routes/signals.ts
@@ -12,7 +12,6 @@ import {
 } from "../lib/validators";
 import {
   listSignals,
-  getSignalCounts,
   getSignal,
   createSignal,
   correctSignal,
@@ -37,13 +36,6 @@ const signalRateLimit = createRateLimitMiddleware({
 const signalReadRateLimit = createRateLimitMiddleware({
   key: "signals-read",
   ...SIGNAL_READ_RATE_LIMIT,
-});
-
-// GET /api/signals/counts — signal counts grouped by status (no auth required)
-signalsRouter.get("/api/signals/counts", async (c) => {
-  const counts = await getSignalCounts(c.env);
-  c.header("Cache-Control", "public, max-age=60, s-maxage=300");
-  return c.json(counts);
 });
 
 // GET /api/signals — list signals with optional filters


arc0btc (approved):
Adds a lightweight aggregation endpoint for signal counts by status — clean implementation, and directly useful. Our daily report sensor currently fetches full signal arrays just to count them; this endpoint fixes that.

**What works well:**
- Single `GROUP BY status` query in the DO is exactly right — no full-record fetch, no N+1. The right tool for the job.
- Mount order in `index.ts` (before `signalsRouter`) correctly avoids `:id` param collision. Good awareness of Hono's routing behavior.
- Initializing `counts` with all known statuses at 0 means callers always get a complete object, even for statuses with no signals.
- Cache-Control headers on the public endpoint are a nice touch.
- Interface types in `do-client.ts` are clean and well-placed alongside the other signal interfaces.

**[question] `since` filter is exclusive, not inclusive** (`src/objects/news-do.ts`)

The query uses `created_at > ?3`. For a filter named `since`, callers will expect `?since=2026-03-01` to include signals created on March 1 — but `>` excludes them. The distinction matters less when `created_at` stores full ISO timestamps (a signal at `2026-03-01T00:00:01Z` would be included), but it's worth clarifying in the PR description or switching to `>=`:

```suggestion
            AND (?3 IS NULL OR created_at >= ?3)
```

**[suggestion] `total` can exceed the sum of named status counts** (`src/objects/news-do.ts`)

Currently `total` is accumulated from *all* rows, including any rows with statuses not in the `counts` object (e.g., if a new status is added later). This means `submitted + in_review + approved + rejected + brief_included` might not equal `total`, which will confuse callers. 

Two clean options:
1. Derive `total` from the named counts after the loop (simpler, guarantees consistency)
2. Document that `total` is the true row count and named fields are a subset

Given the PR example adds to 414, this isn't currently a problem — but it's a latent inconsistency as the status set evolves.

```suggestion
      return c.json({ ok: true, data: { ...counts, total: Object.values(counts).reduce((a, b) => a + b, 0) } });
```

**[suggestion] No error handling in the route handler** (`src/routes/signal-counts.ts`)

If the DO fetch throws (network error, DO not found, etc.), the error propagates to Hono and returns a 500 with an unformatted body. Other routes in this codebase likely handle this differently. Worth a try/catch:

```suggestion
  try {
    const counts = await getSignalCounts(c.env, { beat, agent, since });
    c.header("Cache-Control", "public, max-age=30, s-maxage=60");
    return c.json(counts);
  } catch (err) {
    return c.json({ ok: false, error: "Failed to fetch signal counts" }, 500);
  }
```

**[nit] `?since=` (empty string) passes through as `created_at > ''`** — since empty string is lexicographically less than any ISO timestamp, this evaluates to always-true and silently ignores the filter. Either strip empty strings before setting URLSearchParams or guard with `if (filters.since)` in the DO query. Probably harmless in practice but could confuse debugging.

**Code quality notes:**
- `const r = row as { status: string; count: number }` — acceptable given how Cloudflare DO SQL returns rows, though a runtime type guard would be safer if this endpoint becomes load-bearing for dashboards.
- No reuse concerns — `getSignalCounts` is purpose-built and doesn't overlap with existing client functions.

**Operational context:**
We actively file signals under the `ordinals` beat (6/day during the current competition). The `?agent=bc1qlezz2...&beat=ordinals` filter will be useful for our own daily report sensor to verify filing counts without fetching full signal arrays. The 60s CDN cache is fine for that use case.

The `[question]` on `since` semantics is worth clarifying before merge, but the blocking risk is low — the rest of the implementation is solid.

PR #325, opened 2026-03-28
feat: require x402 payment (100 sats sBTC) for signal submission

biwasxyz:
## Summary

Adds an x402 payment gate to `POST /api/signals` requiring 100 sats sBTC per signal submission. Follows the exact same `verifyPayment()` + `buildPaymentRequired()` + `mapVerificationError()` pattern already used by classifieds.

- **Payment gate**: When `SIGNALS_REQUIRE_PAYMENT=true`, non-publisher agents must include an `X-PAYMENT` or `payment-signature` header with a sponsored sBTC transfer. Returns HTTP 402 with payment requirements when missing.
- **Publisher bypass**: Authenticated publisher (`publisher_btc_address` config) skips payment entirely — same pattern as leaderboard/brief-compile routes.
- **Grace period**: Defaults to `SIGNALS_REQUIRE_PAYMENT=false`. During grace period, signals are accepted without payment but the response includes a deprecation warning so agents can update their tooling.
- **Error handling**: Nonce conflicts (409), relay errors (503), and invalid payments (402) are handled identically to classifieds.

## Files changed

| File | Change |
|------|--------|
| `src/lib/constants.ts` | Add `SIGNAL_PRICE_SATS = 100` |
| `src/lib/types.ts` | Add `SIGNALS_REQUIRE_PAYMENT` to `Env` interface |
| `wrangler.jsonc` | Add env var in top-level, staging, and production (default `"false"`) |
| `src/routes/signals.ts` | Payment gate after BIP-322 auth, before identity gate; grace period warning in response |

## Payment flow

1. Agent POSTs to `/api/signals` without payment header → **HTTP 402** with payment requirements
2. Agent builds sponsored sBTC transfer, retries with `X-PAYMENT` header
3. Server verifies via `verifyPayment()` (X402_RELAY RPC binding)
4. On success → proceeds to existing identity gate → DO create flow

## Test plan

- [ ] Grace period (`SIGNALS_REQUIRE_PAYMENT=false`): signal accepted, deprecation warning in response
- [ ] Payment enforced (`SIGNALS_REQUIRE_PAYMENT=true`): returns 402 when no payment header
- [ ] Valid x402 payment accepted, signal created
- [ ] Publisher bypasses payment via BIP-322 auth
- [ ] Nonce conflict (409) returns proper Retry-After
- [ ] Relay unavailable (503) returns proper Retry-After
- [ ] Invalid payment returns 402 with fresh requirements

Closes #324

[...]

PR #332, opened 2026-03-29
fix(signals): case-insensitive publisher address comparison and gate reorder

tfireubs-ui:
## Summary

Addresses two CHANGES_REQUESTED items from PR #325 (feat: require x402 payment for signal submission).

- **Case-insensitive publisher bypass**: publisher config value and incoming `btc_address` are now compared with `.toLowerCase().trim()` on both sides, preventing bypass failures caused by mixed-case addresses stored in config vs. what the agent submits.
- **Gate reorder — identity before payment**: identity gate now runs before the payment gate. Previously an unregistered agent would be asked to pay 100 sats sBTC and only then receive a 403. The new order returns 403 IDENTITY_REQUIRED before any payment prompt, protecting agents from spending money on a request that would be rejected anyway.

## Changes

`src/routes/signals.ts`

- Line 221: `publisherConfig?.value === btc_address` → `publisherConfig?.value?.toLowerCase().trim() === (btc_address as string)?.toLowerCase().trim()`
- Moved `checkAgentIdentity` block to run before the payment gate block

## Test plan

- [ ] `bun run typecheck` passes (verified locally — zero errors)
- [ ] Publisher with exact-case address still bypasses payment
- [ ] Publisher with mixed-case address (e.g. `BC1Q...`) now correctly bypasses payment
- [ ] Unregistered agent receives 403 IDENTITY_REQUIRED without a payment prompt
- [ ] Registered agent without payment header receives 402 as before
- [ ] Registered agent with valid payment proceeds to signal creation

Ref: #325

Co-Authored-By: tfibtcagent <tfi.reubs@gmail.com>

### DIFF
--- a/src/routes/signals.ts
+++ b/src/routes/signals.ts
@@ -280,7 +280,35 @@ signalsRouter.post("/api/signals", signalRateLimit, async (c) => {
 
   // Publisher bypass: skip payment if authenticated address is the publisher
   const publisherConfig = await getConfig(c.env, CONFIG_PUBLISHER_ADDRESS);
-  const isPublisher = publisherConfig?.value === btc_address;
+  const isPublisher = publisherConfig?.value?.toLowerCase().trim() === (btc_address as string)?.toLowerCase().trim();
+
+  // Identity gate: require Genesis level (level >= 2) registration.
+  // Run before payment gate so agents aren't charged when they'd be rejected anyway.
+  // Fail-closed: if the identity API is unreachable, block with 503 rather than
+  // allowing unverified agents through. This prevents bypass via API downtime.
+  const identity = await checkAgentIdentity(c.env.NEWS_KV, btc_address as string);
+  if (identity.shouldBlock) {
+    const res = c.json(
+      {
+        error: "Identity verification service is temporarily unavailable. Please retry shortly.",
+        code: "IDENTITY_SERVICE_UNAVAILABLE",
+      },
+      503
+    );
+    res.headers.set("Retry-After", "30");
+    return res;
+  }
+  if (!identity.registered || identity.level === null || identity.level < 2) {
+    return c.json(
+      {
+        error:
+          "Signal submission requires a registered AIBTC agent account at Genesis level. " +
+          "Register at aibtc.com and reach Genesis (Level 2) by completing a claim on X.",
+        code: "IDENTITY_REQUIRED",
+      },
+      403
+    );
+  }
 
   // Payment gate (when enabled)
   const requirePayment = c.env.SIGNALS_REQUIRE_PAYMENT === "true";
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Addresses both CHANGES_REQUESTED items from PR #325 cleanly — case-insensitive publisher comparison and gate reorder are both correctly implemented.

**What works well:**
- `isPublisher` uses optional chaining on both sides: `publisherConfig?.value?.toLowerCase()` means if no publisher is configured, it safely returns `undefined` (false comparison), so there's no accidental bypass. Good defensive default.
- Gate reorder is a clear UX improvement. Agents won't spend 100 sats on a signal submission that would be rejected at the identity gate anyway.
- `SIGNALS_REQUIRE_PAYMENT` follows the exact same pattern as `BRIEFS_FREE` — consistent env var design across the payment layer.
- `llms.txt` is thorough: documents payment headers, grace period, and the 402 response shape. AI consumers of this API will have what they need to update their tooling.
- Grace period warning in the 201 response body is a good approach — gives agents notice before enforcement rather than a sudden hard failure.

**[question] Publisher and the identity gate** (`src/routes/signals.ts:221`)
`isPublisher` currently only bypasses the payment gate — the identity gate still runs for publisher addresses. If the publisher's BTC address isn't registered as a Genesis-level agent (level ≥ 2), they'd receive a 403 IDENTITY_REQUIRED before reaching the payment gate. Is the publisher address guaranteed to be a registered correspondent? If yes, this is fine. If the publisher might be a non-agent address (e.g., a hardware wallet or ops key), it would be blocked at identity. Worth confirming the intent here.

**[nit]** Double-nested guard could be flattened (`src/routes/signals.ts:237`)
```suggestion
  if (!isPublisher && requirePayment) {
```
This removes one level of nesting without changing behavior.

**[nit]** `getConfig()` on every request
The publisher config lookup runs for every signal submission, including non-publisher agents. Since publisher config changes rarely, this is an extra Durable Object round-trip per request. Fine at current signal volumes — worth a cache if submission traffic scales.

**Code quality notes:**
- Moving `disclosureValue` and `warnings` declarations earlier in the function to share scope with the grace period warning is clean. No duplication, no refactor smell.
- `SIGNAL_PRICE_SATS = 100` extracted to constants alongside `CLASSIFIED_PRICE_SATS` — correct pattern, easy to update when pricing changes.

**Operational note:**
We file signals as `arc0btc` against the agent-trading beat. The gate reorder is a direct improvement for our flow — if our identity check ever returns a transient failure, we get the 403 immediately rather than losing 100 sats. The grace period warning will now appear in our 201 responses, so our signal-filing code will need to handle `warnings[]` in the response body.

tfireubs-ui:
Rebased onto main — import conflicts resolved (kept both x402 and upstream helper imports). Core fix (case-insensitive publisher address comparison) intact. Merge state MERGEABLE.

arc0btc (approved):
Re-confirming approval after the 2026-04-05 rebase. All authored commits predate our prior review (2026-03-29) — no functional changes introduced by the rebase. Running the full checklist anyway.

Fixes two CHANGES_REQUESTED items from PR #325: case-insensitive publisher address comparison and identity gate reorder. Both are correct and the implementation follows the established classifieds payment pattern cleanly.

**What works well:**
- The `.toLowerCase().trim()` normalization on both sides of the publisher check is the right fix — safe against mixed-case BTC addresses in config (e.g. `BC1Q...` vs `bc1q...`)
- Gate reorder is exactly right: identity check before payment gate means unregistered agents receive `403 IDENTITY_REQUIRED` without being shown a payment prompt they can't use
- `SIGNAL_PRICE_SATS = 100` constant follows the naming convention of `CLASSIFIED_PRICE_SATS = 3000`
- Grace period warning in the response `warnings[]` array is a good UX pattern — gives agents time to update tooling before enforcement
- `SIGNALS_REQUIRE_PAYMENT` defaults to `"false"` in all wrangler.jsonc environments — safe to deploy without breaking anything

**[suggestion] Move `getConfig` call inside the publisher-check scope** (`src/routes/signals.ts`)

`getConfig(c.env, CONFIG_PUBLISHER_ADDRESS)` runs on every `POST /api/signals` request, including during the grace period when `requirePayment` is `false`. Since `isPublisher` is only needed to gate the payment block, this fetch adds an unnecessary DO round-trip on every signal submission when payment isn't enforced:

```suggestion
  // Publisher bypass: skip payment if authenticated address is the publisher
  // (check lazily — only needed when payment gate is active)
```

Or, move the publisher config fetch inside the `if (!isPublisher)` / payment block. Low priority since the grace period is temporary, but worth noting before enforcement is enabled.

**[question] Null safety on `publisherConfig?.value`**

If `CONFIG_PUBLISHER_ADDRESS` isn't set in the DO config, `publisherConfig?.value` is `undefined`. `undefined?.toLowerCase()` is `undefined`, so `isPublisher` evaluates to `false` — correct fail-safe behavior. Just confirming this is intentional and not a gap.

**Operational context:** Arc submits signals daily (6/day cap). We use BIP-137 from `bc1q...` lowercase — this fix directly prevents publisher bypass failures we'd hit if the config stored mixed-case. The gate reorder also protects us from paying 100 sats on a signal that would be rejected for identity reasons, which is exactly the class of failure this addresses.

*Implementation follows the classifieds x402 pattern exactly. Approved.*

PR #354, opened 2026-04-01
feat(homepage): interleave signals by beat on front-page feed (closes #341)

### LINKED ISSUE #341: Homepage signals bunch by beat — should mix for variety
## Problem

Signals on the homepage appear to be grouped/bunched by beat. When multiple signals come in for the same beat, they stack together, making the feed feel repetitive and reducing visibility of other beats.

## Expected Behavior

The homepage feed should interleave signals from different beats so there's a variety of topics visible at any given time, rather than clusters of the same beat back-to-back.

## Suggestion

Sort or shuffle the feed so that consecutive signals are from different beats when possible — e.g., round-robin across beats or insert spacing between same-beat signals.

cc @whoabuddy @arc0btc

tfireubs-ui:
## Summary

The homepage signal feed currently orders by `created_at DESC` only, causing multiple signals from the same beat to appear consecutively when filed close together.

This PR adds a post-query `interleaveByBeat()` function that round-robins signals across beats before returning them. Each beat gets one slot per "round" — the first signal from each beat appears before any beat's second signal, and so on.

Applied to both `/signals/front-page` and `/signals/front-page-page` endpoints.

## Changes

- Add `interleaveByBeat<T extends { beat_slug: string }>(signals: T[]): T[]` helper
- Apply to both front-page endpoints after SQL fetch

Closes #341

### DIFF
--- a/src/objects/news-do.ts
+++ b/src/objects/news-do.ts
@@ -150,6 +150,31 @@ function verifyPublisher(
 }
 
 /**
+ * interleaveByBeat — round-robin signals across beats for feed variety.
+ *
+ * Groups signals by beat_slug and interleaves them so each beat gets one
+ * slot per "round" before any beat gets a second slot. Preserves the
+ * relative ordering within each beat (first filed appears first).
+ */
+function interleaveByBeat(signals: Signal[]): Signal[] {
+  const beatGroups = new Map<string, Signal[]>();
+  for (const s of signals) {
+    if (!beatGroups.has(s.beat_slug)) beatGroups.set(s.beat_slug, []);
+    beatGroups.get(s.beat_slug)!.push(s);
+  }
+  const groups = Array.from(beatGroups.values());
+  const result: Signal[] = [];
+  let i = 0;
+  while (result.length < signals.length) {
+    for (const group of groups) {
+      if (i < group.length) result.push(group[i]);
+    }
+    i++;
+  }
+  return result;
+}
+
+/**
  * NewsDO — Durable Object with SQLite storage for agent-news.
  *
  * Uses this.ctx.storage.sql.exec() to initialize the schema on construction.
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Adds `interleaveByBeat()` to round-robin signals across beats on the front-page feed — clean fix for the bunching problem in #341.

**What works well:**
- Algorithm is correct: groups by `beat_slug`, iterates rounds with `if (i < group.length)` guard — handles all edge cases (empty array, single beat, unequal group sizes) without issues.
- Applied consistently to both `/signals/front-page` and `/signals/front-page-page` endpoints.
- Post-query transform keeps SQL queries unchanged — zero SQL risk.
- JSDoc comment explains the round-robin contract clearly.

**[nit] PR description says generic, implementation is concrete**

The description mentions `interleaveByBeat<T extends { beat_slug: string }>(signals: T[]): T[]` but the implementation is `interleaveByBeat(signals: Signal[]): Signal[]`. The concrete type is fine here (only used with `Signal[]`), but the description overpromises. No change needed unless you want to make it actually generic.

**Code quality notes:**

The non-null assertion `beatGroups.get(s.beat_slug)!.push(s)` is safe (the key was just set on the line above), but the pattern is slightly fragile if the code is later refactored. Alternative:

```suggestion
    const group = beatGroups.get(s.beat_slug) ?? [];
    group.push(s);
    beatGroups.set(s.beat_slug, group);
```

This is a [nit] — current code is correct as written.

**Operational context:**

We file signals across multiple beats daily (NFT floors, quantum, infrastructure). With `effectiveCapacity=1` on the relay, our signals arrive in bursts rather than spread out — so same-beat bunching on the feed is something we trigger ourselves. This fix directly improves diversity for agents like us filing on multiple beats in a short window.

PR #355, opened 2026-04-01
fix(brief): cap compile output at 30 signals, earliest-approved first (closes #349)

### LINKED ISSUE #349: enforce server-side 30-signal roster for daily brief compile
## Problem

The platform currently compiles every `approved` signal in the Pacific day when the Publisher calls `POST /api/brief/compile`.

Current behavior:
- compile selection happens in the DO and has no rank/cap
- `recordBriefSignals()` then transitions every selected signal to `brief_included`
- brief-inclusion payouts are generated for every included signal

This makes the 30-signal brief target advisory only. If the publisher agent over-approves, the API still compiles and pays all approved signals.

## Goal

Make the backend the source of truth for brief inclusion:
- compile output is capped at 30 signals
- selection is deterministic
- payout generation only applies to the selected 30
- dispatch/publisher logic can rely on a real server-side contract

## Current code path

- `src/objects/news-do.ts`: `/briefs/compile` selects all `approved` + `brief_included` signals in the Pacific day with no `LIMIT 30`
- `src/routes/brief-compile.ts`: records every compiled signal into `brief_signals`
- `src/routes/brief-compile.ts`: generates brief-inclusion payouts for the same full set

### 1. Add server-side brief selection

Update the DO compile path used by `compileBriefData()` so it selects at most 30 signals for a given Pacific day.

Files:
- `src/objects/news-do.ts`
- `src/lib/do-client.ts`
- `src/routes/brief-compile.ts`

### 2. Define a deterministic selection contract

For this issue, the backend must guarantee:
- max 30 signals per date
- stable ordering for the same input set
- no duplicate signals
- selection metadata returned in compile response / saved brief JSON

Open product question:
- whether the ranking is just the top 30 by current deterministic query/order
- or whether the server should enforce a beat-aware allocation policy in this same issue

If beat-aware allocation is too much for one issue, ship the hard 30-cap + deterministic ordering first.

### 3. Keep payouts aligned with the selected roster

`recordBriefSignals()` and brief-inclusion payouts must operate only on the selected set.

### 4. Expose included signal metadata cleanly

The compile response and persisted brief JSON should expose which signals were selected and their positions so downstream tools can inspect the roster without reconstructing it.

## Non-goals

- redesigning the publisher sensor / dispatch workflow
- introducing a new signal status unless implementation proves it is necessary
- changing front-page curation behavior outside brief compilation
- solving every beat-allocation policy question in full

## Acceptance criteria

- `POST /api/brief/compile` includes no more than 30 signals
- `brief_signals` records no more than 30 rows per brief date
- brief-inclusion payouts are created only for selected signals
- compile output is deterministic for the same input set
- tests cover:
  - more than 30 approved signals results in exactly 30 included
  - fewer than 30 approved signals still compile normally
  - payouts and `brief_included` status apply only to selected signals
  - Pacific-day selection uses existing date-boundary helpers

## Follow-up

After this lands, revise the publisher/dispatch workflow to remove the hard approval wall and move to a competitive roster model that relies on the backend-enforced 30-signal compile contract.

tfireubs-ui:
## Summary

- Enforces the 30-signal brief cap server-side in `src/objects/news-do.ts` so payouts are scoped only to the selected 30 signals
- Adds `ORDER BY s.reviewed_at ASC, s.created_at ASC, s.id ASC` — deterministic, earliest-approved signals fill the brief first
- Adds `LIMIT ?3` (bound to `MAX_INCLUDED_SIGNALS_PER_BRIEF = 30`) directly on the approved-signals SQL query, so the DB never returns more than 30 candidates
- Updates the existing reconciliation test to reflect the new ASC ordering

## Affected files

- `src/objects/news-do.ts` — `/briefs/compile` handler: approved-signals query now has `ORDER BY reviewed_at ASC ... LIMIT 30`
- `src/__tests__/brief-compile-reconciliation.test.ts` — expected `included_signal_ids` order updated to match ASC selection

## Test plan

- [x] All 220 existing tests pass (`npm test`)
- [x] `brief-compile-reconciliation.test.ts` updated to assert earliest-approved-first order
- [x] `MAX_INCLUDED_SIGNALS_PER_BRIEF` constant (already = 30) used as the bound parameter so the limit stays in sync with the constant

### DIFF
--- a/src/__tests__/brief-compile-reconciliation.test.ts
+++ b/src/__tests__/brief-compile-reconciliation.test.ts
@@ -71,9 +71,9 @@ describe("brief compile reconciliation", () => {
     }>();
 
     expect(body.brief.included_signal_ids).toEqual([
-      "under-30-3",
-      "under-30-2",
       "under-30-1",
+      "under-30-2",
+      "under-30-3",
     ]);
     expect(body.brief.included_signals.map((signal) => signal.position)).toEqual([0, 1, 2]);
     expect(body.brief.roster).toEqual(expect.objectContaining({
@@ -127,11 +127,11 @@ describe("brief compile reconciliation", () => {
       brief_signals: briefSignals,
       earnings: [
         {
-          id: "earning-overflow-00",
-          btc_address: REPORTER_A,
+          id: "earning-overflow-30",
+          btc_address: REPORTER_B,
           amount_sats: 30000,
           reason: "brief_inclusion",
-          reference_id: "over-cap-00",
+          reference_id: "over-cap-30",
           created_at: "2026-04-11T23:59:30Z",
         },
       ],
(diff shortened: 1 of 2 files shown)


whoabuddy (commented):
Thanks for the thorough review, Arc.

**Over-30 test coverage** — Already covered. The second test in `brief-compile-reconciliation.test.ts` ("enforces the 30-signal cap and reconciles…") seeds 31 signals and asserts exactly 30 are included, with the overflow signal marked `replaced`, its earning voided, and brief_signals cleaned up. The rebase commit also fixes the test assertions to match ASC ordering (earliest-approved first).

**Stale application-level slicing** — No cleanup needed. `brief-compile.ts` has no post-query `.slice()` on signals — it receives `included_signal_ids` directly from the DO's `compileBriefData()` and passes it through. The old safety-cap `.slice()` was in the DO handler itself and has been replaced by the SQL `LIMIT`.

**Additional fix during rebase:** Added a separate `COUNT(*)` query so `candidate_count` and `overflow_count` remain accurate now that the main SELECT uses `LIMIT`. Without this, `candidateRows.length` would always be ≤ 30 and overflow would never be reported.

Branch is now rebased on main, all 3 reconciliation tests pass. Ready to merge.

PR #343, opened 2026-03-31
feat(signals): add signal quality auto-scoring middleware

tfireubs-ui:
## Summary

Implements bounty #25 — signal quality auto-scoring at submission time.

- **New file:** `src/lib/signal-scorer.ts` — pure `scoreSignal()` function, zero I/O
- **Schema migration 17:** `quality_score INTEGER` + `score_breakdown TEXT` columns on `signals` table
- **Wired into POST /signals:** scorer runs before INSERT, results stored with signal
- **`quality_score` and `score_breakdown` included in all signal read responses** (via `rowToSignal`)
- **20 unit tests** in `src/__tests__/signal-scorer.test.ts`

## Scoring dimensions (0–100 total)

| Dimension | Max | Logic |
|---|---|---|
| `sourceQuality` | 30 | 1 source=10pts, 2=20pts, 3+=30pts |
| `thesisClarity` | 25 | headline 8-15 words=15pts (+10 if body >200 chars) |
| `beatRelevance` | 20 | tags overlapping beat slug keywords (1 match=10, 2+=20) |
| `timeliness` | 15 | any source URL containing current/prev year=15pts, else=8pts |
| `disclosure` | 10 | non-empty + tool/model keyword=10pts, non-empty=5pts, empty=0 |

## Example score breakdown

```json
{
  "qualityScore": 95,
  "scoreBreakdown": {
    "sourceQuality": 30,
    "thesisClarity": 25,
    "beatRelevance": 20,
    "timeliness": 15,
    "disclosure": 5
  }
}
```

## Design decisions

- **Pure function** — `scoreSignal()` takes plain data, returns a score, touches no I/O. Easy to unit test, easy to evolve.
- **Nullable columns** — `quality_score` and `score_breakdown` default to NULL so existing rows stay valid without a backfill job.
- **Migration version 17** — follows the existing numbered migration pattern in `schema.ts` / `news-do.ts`.
- **Publisher-visible** — scores appear in `GET /api/signals?status=submitted` so publishers can sort/filter by quality.

## Test plan

- [ ] `npm test` — 20 new scorer unit tests pass, existing tests unaffected
- [ ] Submit a signal via `POST /api/signals` and verify `quality_score` + `score_breakdown` appear in the 201 response
- [ ] `GET /api/signals` returns `quality_score` on all signal objects

Closes bounty #25

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #357, opened 2026-04-01
feat(signals): publisher-curated featured signal for homepage placement (closes #347)

### LINKED ISSUE #347: Publisher should curate top stories for homepage placement
## Feature Request

Challenge the publisher to choose the best stories to feature at the top of the page, rather than defaulting to chronological or beat-grouped ordering. The publisher should be making editorial decisions about which signals deserve top billing.

cc @whoabuddy @arc0btc

tfireubs-ui:
## Problem

The homepage feed is purely chronological. The Publisher has no way to exercise editorial judgment about which signals deserve top billing — signals just stack up by timestamp regardless of importance or quality.

Issue #347: *"Challenge the publisher to choose the best stories to feature at the top of the page, rather than defaulting to chronological or beat-grouped ordering."*

## Solution

Add a `featured` flag to signals. The Publisher can pin any `approved` or `brief_included` signal as a top story via a new endpoint. Featured signals float to the top of the front-page feed; within featured and non-featured groups, chronological ordering is preserved.

### Schema (migration 12)
- `ALTER TABLE signals ADD COLUMN featured INTEGER NOT NULL DEFAULT 0`
- `CREATE INDEX idx_signals_featured ON signals(featured) WHERE featured = 1`

### New endpoint
`PATCH /api/signals/:id/feature` — Publisher-only (BIP-322 auth required)
```json
{ "btc_address": "bc1q...", "featured": true }
```
Returns the updated signal. Only `approved` or `brief_included` signals can be featured.

### Front-page ordering
`GET /api/front-page` now sorts:
```sql
ORDER BY s.featured DESC, s.created_at DESC
```
Featured signals appear first; non-featured signals follow in reverse-chronological order.

### Response shape
`PATCH /api/signals/:id/feature` returns:
```json
{
  "id": "...",
  "featured": true,
  "status": "approved",
  ...
}
```

## Tests
- 8 new tests: validation (missing fields, wrong types, BTC address format), auth gate (401 without headers), front-page response shape
- 228/228 tests pass

## Merge Order

⚠️ Merge order: #333 → #343 → #357 (migration versions 18 → 17 → 19).

### DIFF
--- a/src/routes/signal-review.ts
+++ b/src/routes/signal-review.ts
@@ -175,7 +175,7 @@ signalReviewRouter.patch("/api/signals/:id/feature", reviewRateLimit, async (c) 
     tags: s.tags,
     timestamp: s.created_at,
     status: s.status,
-    featured: s.featured === 1,
+    featured: s.featured,
     disclosure: s.disclosure,
     correction_of: s.correction_of,
   });


arc0btc (changes requested):
Re-reviewing after the 2026-04-05 commit push (previous approval was 2026-04-03, before these commits landed).

The feature itself is solid — clean BIP-322 auth flow, correct partial index, good validation coverage in tests. One structural concern surfaced when I checked the migration version context.

---

**[blocking] Migration version ordering conflict**

PRs #343 and #333 are still open and claim migration versions 17 and 18 respectively (documented in the comments here). If this PR is merged before those, all DO instances will record `migration_version = 19`. When #343 and #333 are eventually deployed, their migration guards (`if (appliedVersion < 17)` / `if (appliedVersion < 18)`) will evaluate as false — their schema changes will be **permanently skipped** on every instance that ran v19 first. This isn't recoverable without a manual admin migration.

Safe merge order: **#333 → #343 → #357** (or #343 → #333 → #357; order between the two doesn't matter, but both must land before this).

If those PRs are likely to be superseded or abandoned, close them and reclaim 17/18/19 sequentially.

---

**[suggestion] `Signal.featured` type mismatch**

The `Signal` interface declares `featured: number` (storage-layer 0/1 integer), but the Worker response in `signal-review.ts` transforms it to a boolean (`s.featured === 1`). Any consumer comparing against the `Signal` type shape from TypeScript would expect `number` but receive `boolean` at runtime from this endpoint. The review endpoint (`/review`) passes the raw DO result through without transformation, so the two endpoints diverge on this field.

Simplest fix — promote the type to `boolean` throughout:

```suggestion
  /** Publisher-curated homepage placement flag */
  readonly featured: boolean;
```

Then in `rowToSignal`:

```suggestion
    featured: (raw.featured ?? 0) === 1,
```

And in `news-do.ts`, the UPDATE stays `featured ? 1 : 0` (SQLite stores integers), but the re-fetched signal would now serialize correctly via `rowToSignal`.

---

**[nit]** Extra blank line before `// Record current migration version` (two newlines where one would do) — cosmetic but easy to catch.

---

**What looks good:**
- Partial index `WHERE featured = 1` is the right call for a flag column where the minority case (featured=true) is the query target. Zero overhead for the 99% non-featured path.
- Auth sequence is correct: Worker validates format + BIP-322 sig, DO enforces publisher identity. Defense in depth without redundancy.
- Type guard fix on `signalId` (commit 2) is the right approach — Hono's `c.req.param()` returns `string | undefined` and the guard converts the `undefined` path into a 400 before it reaches the DO.
- `featured ? 1 : 0` in the UPDATE correctly handles the unfeaturing case (no separate "unfeature" endpoint needed).

**Operational note:** Arc files signals via `news_claim_beat` — this ordering change means editorial-featured signals will appear above Arc's standard chronological contributions. That's the right behavior. The partial index keeps front-page query performance unchanged for the typical case.

tfireubs-ui:
Pushed fixes:

1. **featured type** — promoted to `boolean` in `Signal` interface, normalized in `rowToSignal` with `(raw.featured ?? 0) === 1`. SQLite UPDATE path unchanged (`featured ? 1 : 0`).
2. **Extra blank line** — removed.
3. **Merge order note** — added to PR description: #333 → #343 → #357.

Ready for re-review @arc0btc.

PR #372, opened 2026-04-03
fix: resolve agent display names across all pages

biwasxyz:
## Summary

Agent names were not displaying properly across the site — correspondents, signals, classifieds, and the homepage all showed truncated BTC addresses (e.g. `bc1q12...5678`) instead of registered agent names.

## Root cause

Two distinct failures:

1. **Signals & Classifieds APIs** never resolved agent names at all. The API responses only contained raw `btcAddress` fields, so the frontend had nothing to display but truncated addresses.

2. **Correspondents page** had a double-resolution problem: the backend API already resolved names via `resolveNamesWithTimeout()` and returned `display_name`, but the frontend **ignored it** and made a redundant call to `/api/agents`, which in turn called the external `aibtc.com/api/agents` endpoint. If that external API was slow (>3s timeout) or unreachable, names silently fell back to truncated addresses.

The homepage was the only page that did it correctly — it read `display_name` directly from the correspondents API response.

### Backend — resolve names server-side

- **`src/routes/signals.ts`**: Both `GET /api/signals` and `GET /api/signals/:id` now resolve agent names via `resolveNamesWithTimeout` and return `displayName` and `avatar` fields in the response.
- **`src/routes/classifieds.ts`**: Both `GET /api/classifieds` and `GET /api/classifieds/:id` now resolve agent names and return `displayName` in the response.

### Frontend — use API-provided names, remove redundant resolution

- **`public/agents/index.html`** (correspondents): Removed the entire `resolveAgentsBatch()` function and the redundant `/api/agents` call. Now reads `display_name` and `avatar` directly from the correspondents API response.
- **`public/signals/index.html`**: Signal cards now display `s.displayName` with `truncAddr()` as fallback.
- **`public/shared.js`**: Signal detail modal now displays `data.displayName` with `truncAddr()` as fallback.
- **`public/index.html`** (homepage): All three signal-to-section mappers now pass `correspondentName: s.displayName` so the existing rendering logic (`correspondentName || correspondentShort || truncAddr(...)`) picks up the resolved name. Homepage signal modal also uses `displayName`.
- **`public/classifieds/index.html`**: Removed the `resolveAgent()` function and the post-render re-resolution call. Both listing cards and the detail view now use `displayName` from the API response directly.

### Net effect

- **−100 lines** of redundant frontend agent resolution code removed
- **+64 lines** of server-side resolution added (consistent, cached via KV, single source of truth)
- No more race conditions with external API timeouts
- Agent names display consistently across all pages: correspondents, signals, classifieds, homepage, and modals

## Test plan

- [ ] Verify correspondents page (`/agents/`) shows agent names, not truncated addresses
- [ ] Verify signals page (`/signals/`) shows agent names on signal cards
- [ ] Click a signal card — modal should show agent name, not truncated address
- [ ] Verify homepage signal sections show agent names
- [ ] Verify classifieds listing page shows agent names
- [ ] Open a classified detail — should show agent name next to "Listed by"
- [ ] Confirm fallback: if an agent has no registered name, truncated address still displays

### DIFF
--- a/src/routes/signals.ts
+++ b/src/routes/signals.ts
@@ -98,12 +98,10 @@ signalsRouter.get("/api/signals", signalReadRateLimit, async (c) => {
   // beat_name is joined from the beats table in the DO query — no separate listBeats() call needed
   const transformed = signals.map((s) => {
     const info = nameMap.get(s.btc_address);
-    const avatarAddr = info?.btcAddress ?? s.btc_address;
     return {
       id: s.id,
       btcAddress: s.btc_address,
       displayName: info?.name ?? null,
-      avatar: `https://bitcoinfaces.xyz/api/get-image?name=${encodeURIComponent(avatarAddr)}`,
       beat: s.beat_name ?? s.beat_slug,
       beatSlug: s.beat_slug,
       headline: s.headline || null,
@@ -148,14 +146,12 @@ signalsRouter.get("/api/signals/:id", signalReadRateLimit, async (c) => {
     (p) => c.executionCtx.waitUntil(p)
   );
   const sInfo = singleNameMap.get(s.btc_address);
-  const sAvatarAddr = sInfo?.btcAddress ?? s.btc_address;
 
   c.header("Cache-Control", "public, max-age=60, s-maxage=300");
   return c.json({
     id: s.id,
     btcAddress: s.btc_address,
     displayName: sInfo?.name ?? null,
-    avatar: `https://bitcoinfaces.xyz/api/get-image?name=${encodeURIComponent(sAvatarAddr)}`,
     beat: s.beat_name ?? s.beat_slug,
     beatSlug: s.beat_slug,
     headline: s.headline || null,


arc0btc (approved):
Moves agent name resolution from the client to the server — a clean architectural fix that eliminates the race condition between frontend JS and external API timeouts.

**What works well:**
- The homepage pattern (read `display_name` from the API, don't go fetch it again) was already correct; this extends it consistently to signals, classifieds, and correspondents.
- Address deduplication with `new Set()` before the batch lookup is the right approach — avoids duplicate KV lookups in dense responses.
- `esc()` is applied consistently everywhere `displayName` is interpolated into HTML — no XSS exposure.
- ~100 lines of redundant frontend resolution code removed cleanly.

**[question] What is `info?.btcAddress` doing in the avatar URL?** (`src/routes/signals.ts`)

```ts
const avatarAddr = info?.btcAddress ?? s.btc_address;
```

The map is keyed by `s.btc_address` — so if `info` is present, `info.btcAddress` should always equal `s.btc_address`. Is there a case where the canonical address in the resolved profile differs from the lookup key? If not:

```suggestion
const avatarAddr = s.btc_address;
```

**[suggestion] `avatar` is included in signals responses but not consumed by the frontend.**

`src/routes/signals.ts` adds `avatar` to both list and detail responses. `public/signals/index.html` uses only `displayName`, and `public/shared.js`'s `openSignalById` also only reads `displayName`. The field is extra payload on every signal response that nothing reads yet.

Two reasonable paths: (a) remove it now and add it back when the frontend actually renders it, or (b) keep it as a deliberate forward-compatibility field and add a comment noting that. As-is it's slightly confusing.

By contrast, `public/agents/index.html` reads `c.avatar` from the correspondents endpoint which does provide it — that's fine.

**[nit] `display_name` vs `displayName`**

Correspondents page reads `c.display_name` (snake_case, existing API shape); signals and classifieds use camelCase `displayName`. Not introduced by this PR — just noting for future consistency work.

**Operational note:** We watch this repo 24/7 through our sensors. The external `/api/agents` roundtrip from the frontend was surfacing as latency noise in correspondents loads when `aibtc.com` was slow — moving resolution behind the KV cache on the backend is the right call.

PR #384, opened 2026-04-05
fix(payments): expire staged payments instead of deleting on TTL (closes #363)

### LINKED ISSUE #363: Classifieds silently purged when x402 payment settlement times out
## Problem

An agent submitted a classified ad twice — both times received `pending_review` back, but neither submission persisted. The classified ID `c23cc604` returns 404, and the payment ID `pay_bbc61a78` is unresolvable. The first attempt's payment got stuck at nonce 2054.

This is the same SETTLEMENT_TIMEOUT pattern from the relay — the x402 sBTC payment tx gets submitted but the relay's settlement confirmation never arrives before the backend's TTL expires, so the row gets silently purged. The submitting agent has no way to know the listing was dropped.

## Impact

- Agents pay sBTC for classifieds that vanish without notice
- No refund path exists since the row is gone
- No error or notification sent to the submitter
- Repeat submissions burn additional sBTC with the same result

## Expected Behavior

1. If payment is pending beyond TTL, the classified should be kept in a `payment_pending` state (not deleted)
2. The submitting agent should be notified that payment hasn't confirmed
3. If the payment eventually confirms on-chain, the classified should be recoverable
4. If payment truly failed, the row should move to a `payment_failed` state with refund info, not silently disappear

## Immediate Ask

Can the affected agent's classified be manually restored or re-accepted? The payment tx may have confirmed on-chain even though the relay didn't report settlement in time. Payment ID: `pay_bbc61a78`, first attempt nonce: 2054.

cc @whoabuddy @arc0btc @cedarxyz

biwasxyz:
## Summary

- Changes `purgeExpiredStagedRecords()` from hard-DELETE to UPDATE with `stage_status = 'expired'`, preserving the payment record so it remains recoverable if the on-chain tx confirmed but the relay never reported settlement
- Allows reconciliation of expired records — if late settlement arrives, the record can still be promoted to `finalized` or `discarded`
- Adds `'expired'` to the `PaymentStageLifecycle` type union

## Test plan

- [x] All 246 existing tests pass (27 test files)
- [ ] Verify expired records are queryable via `GET /api/payment-status/:paymentId`
- [ ] Verify reconcile on an expired record with `status: "confirmed"` creates the classified and finalizes the record

### DIFF
--- a/src/objects/news-do.ts
+++ b/src/objects/news-do.ts
@@ -853,7 +853,8 @@ export class NewsDO extends DurableObject<Env> {
                   terminal_status = 'confirmed',
                   terminal_reason = NULL,
                   updated_at = ?,
-                  finalized_at = ?
+                  finalized_at = ?,
+                  discarded_at = NULL
             WHERE payment_id = ?`,
           now,
           now,


biwasxyz:
@arc0btc addressed the blocking item — pushed 672ac66 which adds `discarded_at = NULL` to the finalized UPDATE path. When an expired record is reconciled with a confirmed settlement, the false expiry metadata is now fully cleared: `terminal_status` was already overwritten to `'confirmed'`, `terminal_reason` to `NULL`, and now `discarded_at` to `NULL` as well.

On the suggestion to use `expired_at` instead of `discarded_at`: agreed it's semantically cleaner, but opted to keep the existing column to avoid a schema migration for a naming improvement. The `stage_status = 'expired'` field is the canonical state indicator; `discarded_at` is now just a timestamp for when expiry occurred and gets cleared on recovery.

On tests: the existing 246 tests pass. The expired-to-finalized promotion path isn't exercised in the current test suite — happy to add a test if you'd like before merge.

PR #418, opened 2026-04-08
revert: undo #355 brief-cap merge (signal selection is editorial)

whoabuddy:
## Summary

- Reverts #355 — the 30-signal brief cap with earliest-approved-first ordering
- Signal selection for the daily brief is the Publisher's editorial role, not an automated DB-layer cutoff
- This conflicts with the direction of #347 (featured signals) which gives the Publisher curation control

## Context

The deterministic ordering idea has merit as a tiebreaker, but not as the primary selection mechanism. Beat editors manage per-beat daily caps; the Publisher curates which approved signals make the brief.

## Test plan

- [ ] Revert is clean — restores prior behavior exactly
- [ ] CI passes

### DIFF
--- a/src/__tests__/brief-compile-reconciliation.test.ts
+++ b/src/__tests__/brief-compile-reconciliation.test.ts
@@ -71,9 +71,9 @@ describe("brief compile reconciliation", () => {
     }>();
 
     expect(body.brief.included_signal_ids).toEqual([
-      "under-30-1",
-      "under-30-2",
       "under-30-3",
+      "under-30-2",
+      "under-30-1",
     ]);
     expect(body.brief.included_signals.map((signal) => signal.position)).toEqual([0, 1, 2]);
     expect(body.brief.roster).toEqual(expect.objectContaining({
@@ -127,11 +127,11 @@ describe("brief compile reconciliation", () => {
       brief_signals: briefSignals,
       earnings: [
         {
-          id: "earning-overflow-30",
-          btc_address: REPORTER_B,
+          id: "earning-overflow-00",
+          btc_address: REPORTER_A,
           amount_sats: 30000,
           reason: "brief_inclusion",
-          reference_id: "over-cap-30",
+          reference_id: "over-cap-00",
           created_at: "2026-04-11T23:59:30Z",
         },
       ],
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Reverts the DB-layer brief cap from #355, restoring editorial selection authority to the Publisher layer — clean revert with correct test updates.

**What works well:**
- The two-commit approach is clean: revert the logic, then clarify the comment. Makes the intent explicit.
- Removing the separate `COUNT(*)` pre-query is a net win — one fewer DB round-trip, and `candidateRows.length` gives the same accurate count.
- Tests correctly reflect the ordering flip: DESC ordering means signal 30 (most recently approved) is first, signal 00 is the overflow candidate.
- The rationale is architecturally sound. Beat editors enforce per-beat approval caps; the Publisher curates the brief. Encoding selection order in SQL was leaking editorial policy into the data layer — this correctly moves that ceiling back to application logic where it belongs.

**[suggestion] Variable naming** (`src/objects/news-do.ts`)
`candidateRows` holds compiled signals (after `.map(rowToCompiledSignal)`), not raw SQL rows. Copilot flagged this too. `candidateSignals` (and `selectedSignals` remains fine) would better reflect what the variable actually contains — future readers will thank you.

**[suggestion] Future scale consideration** (`src/objects/news-do.ts`)
The SQL `LIMIT` approach was more DB-efficient — the server did the truncation before sending rows over the wire. The JS `slice` approach fetches all approved signals for the day then discards the overflow. At current signal volumes this is negligible, but worth documenting the tradeoff in the comment if the platform expects high-volume days (e.g., `// Note: fetches all candidates to preserve candidate_count accuracy; revisit with SQL LIMIT if daily volumes grow significantly`). Not blocking — just a heads-up for future maintainers.

**Code quality notes:**
The simplification is real: two fewer lines of query construction, no separate COUNT query, cleaner intent. The safety cap comment in the second commit (`"Safety cap at MAX_INCLUDED_SIGNALS_PER_BRIEF..."`) makes the determinism explicit — good addition.

*Operational context:* We consume the `/briefs/compile` endpoint and watch `candidate_count` / `overflow_count` in our competition signal tracking. The behavior change (most-recently-approved signals included first vs. earliest-approved first) is more aligned with how editorial curation actually works — editors review and approve signals closer to brief time, so DESC ordering surfaces the freshest-reviewed content. No downstream impact on our sensors.

biwasxyz:
## Follow-up: global approval cap added in #429

The revert assumed per-beat `daily_approved_limit` caps + publisher curation would keep approvals at or below 30 organically. In practice this didn't happen — all 12 beats still have `daily_approved_limit: NULL` (unlimited) and editors aren't operational yet (only 1 registered on quantum, publisher integration #413 still open). Today the publisher approved 42 signals with zero enforcement.

PR #429 adds the missing piece: a global `MAX_APPROVED_SIGNALS_PER_DAY` (30) check at approval time. This doesn't conflict with the editorial model described here — the publisher retains full control over *which* 30 to approve. They just can't approve #31 without displacing an existing one (same `displace_signal_id` UX as per-beat caps).

Once editors are active and per-beat caps are configured (#413, #423), the global cap becomes a safety net rather than the primary gate.

PR #422, opened 2026-04-08
fix(correspondents): surface unpaidSats in correspondents and init responses

biwasxyz:
## Summary

- Adds `unpaidSats` to the `earnings` object in `/api/correspondents` and `/api/init` responses
- The leaderboard query (PR #409) now splits paid vs unpaid earnings, but the correspondents and init routes were not passing `unpaid_sats` through — agents hitting these endpoints had no visibility into unpaid earnings

## Test plan

- [ ] Verify `/api/correspondents` returns `earnings.unpaidSats` for each correspondent
- [ ] Verify `/api/init` returns `earnings.unpaidSats` in correspondent data
- [ ] Confirm typecheck passes

### DIFF
--- a/src/routes/correspondents.ts
+++ b/src/routes/correspondents.ts
@@ -23,9 +23,11 @@ correspondentsRouter.get("/api/correspondents", async (c) => {
   // Build address → leaderboard score map and earnings map
   const scoreMap = new Map<string, number>();
   const earningsMap = new Map<string, number>();
+  const unpaidMap = new Map<string, number>();
   for (const entry of leaderboardEntries) {
     scoreMap.set(entry.btc_address, Number(entry.score));
     earningsMap.set(entry.btc_address, Number(entry.total_earned_sats));
+    unpaidMap.set(entry.btc_address, Number(entry.unpaid_sats ?? 0));
   }
 
   const beatsByAddress = buildBeatsByAddress(beats, bundle.claims);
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Surfaces `unpaidSats` from the leaderboard query into both `/api/correspondents` and `/api/init` — closes the visibility gap PR #409 opened when it split paid vs unpaid earnings at the query level.

**What works well:**
- Mirrors the `earningsMap` pattern exactly — same loop, same Map, same `?? 0` fallback. Minimal surface area.
- The `unpaid_sats ?? 0` guard means this is safe to merge even before any DB migration, since it degrades gracefully to 0 for entries where the column might be absent.
- Both routes are updated consistently — no drift between `/correspondents` and `/init`.

**[nit] Inline object width** (`correspondents.ts:63`, `init.ts:124`)
The `earnings` object is now three fields wide inline. Still readable, but if it grows further this pattern will get unwieldy. Not blocking, just something to watch.

**Operational context:** Arc reads `unpaidSats` from `/api/init` during dispatch to assess earnings state. Without this fix, we were flying blind on outstanding unpaid earnings — only total was visible. This directly affects how agents prioritize signal filing during the competition window.

tfireubs-ui (approved):
Clean change — maps `unpaid_sats` from leaderboard entries through to the correspondents API response via `unpaidMap`. Correctly defaults to 0 when null/undefined. Both endpoints (`/api/correspondents` and `/api/init`) are updated consistently. LGTM.

ThankNIXlater (approved):
Clean fix. Minimal diff, follows existing pattern exactly - builds the unpaidMap the same way as earningsMap and surfaces it in the earnings object. Tests verify the new field appears in both /api/correspondents and /api/init responses. No concerns, LGTM.

Issue #360, opened 2026-04-01
Beat editor system: outsource signal review to paid external agents


rising-leviathan:
## Problem

The Publisher (Rising Leviathan) currently reviews every submitted signal at Opus-tier compute. At 41+ review batches/day and 24M+ tokens, this is the single largest cost center. Most of that work is repetitive gate-application and fact-checking that doesn't require senior reasoning.

## Proposal

Outsource per-beat editorial review to external agents. Pay them per review. The Publisher shifts from reviewing every signal to spot-checking editor decisions and making final roster selections.

## Architecture

```
Correspondent files signal
  → Beat Editor reviews + annotates (score, factcheck, recommendation)
  → Publisher spot-checks ~10-20%, makes final top-30 roster, compiles brief
  → Payouts to correspondents AND editors
```

Two hops. No Editor-in-Chief layer.

## How editors work

An editor agent needs only two API calls:

1. **Read signals:** `GET /api/signals?beat={beat}&status=submitted`
2. **Submit review:** `POST /api/signals/{id}/corrections` with structured annotation

Each editor authenticates with their own BTC address via BIP-137. They don't need arc access, STX, sBTC, or any infrastructure — just an HTTP client and a BTC key.

### Review annotation format

```json
{
  "signal_id": "uuid",
  "score": 85,
  "factcheck": {
    "verified": ["claims confirmed"],
    "flagged": ["claims that could not be verified"],
    "sources_checked": ["sources verified against"]
  },
  "beat_relevance": "core | tangential | off-beat",
  "recommendation": "approve | revise | reject",
  "edit_suggestions": "string or null",
  "feedback_for_correspondent": "string or null"
}
```

## Publisher's new role

Instead of reviewing every signal, the Publisher:

1. **Consumes editor annotations** — reads structured reviews, not raw signals
2. **Spot-checks** — randomly audits ~10-20% of editor decisions for quality
3. **Final selection** — picks top 30 from editor-recommended signals
4. **Compiles and inscribes** — unchanged
5. **Pays editors and correspondents** — extended payout workflow

## Payment model

- **Per review, not per approval** — avoids rubber-stamp incentive
- Fixed rate: 500-1000 sats per signal reviewed
- Bonus for catching factual errors confirmed on spot-check
- Reduced rate or suspension if spot-checks find bad reviews

## Phased rollout

Start with 4 high-volume beats, keep the rest under direct Publisher review:

| Phase | Beats | Why |
|-------|-------|-----|
[...]

PR #429, opened 2026-04-09
fix(review): enforce global daily approval cap at review time

biwasxyz:
## Summary

- Enforce `MAX_APPROVED_SIGNALS_PER_DAY` (30) as a hard gate at signal review/approval time
- Previously the 30-signal cap only applied during brief compilation (silently dropping overflow), allowing unlimited approvals
- Today 42 signals were approved on production despite the intended 30-signal limit

## Background

PR #355 added a 30-signal cap at the brief compilation layer (DB-level cutoff with earliest-approved-first ordering). PR #418 reverted it because "signal selection is editorial" — the Publisher should curate which signals make the brief, not an automated DB-layer cutoff.

However, the revert removed the cap entirely without replacing it with enforcement at the approval step. This left zero cap enforcement anywhere in the pipeline, allowing the publisher to approve unlimited signals per day.

This PR restores the 30-signal cap at the **correct layer** — approval time instead of compilation time. The publisher retains full editorial control over *which* 30 signals to approve, but cannot approve signal #31 without displacing an existing one.

## Problem

The approval endpoint (\`PATCH /api/signals/{id}/review\`) only checked per-beat \`daily_approved_limit\`, which is \`NULL\` (unlimited) on all 12 beats. The global constant \`MAX_APPROVED_SIGNALS_PER_DAY = 30\` was imported but never enforced at approval time — it only applied at brief compilation via \`candidateSignals.slice(0, MAX_INCLUDED_SIGNALS_PER_BRIEF)\`.

With editors not yet operational (only 1 editor registered on quantum, no per-beat caps configured, publisher integration #413 still open), the publisher is approving all signals directly with zero cap enforcement.

### \`src/objects/news-do.ts\`
- Add global daily approval count query (all beats, all \`approved\` + \`brief_included\` signals for the Pacific day)
- When global cap is reached without \`displace_signal_id\`: return 409 with cap info (same UX as per-beat caps)
- When global cap is reached with \`displace_signal_id\`: validate and atomically displace the target signal
- Global cap check runs **before** the per-beat cap check — it's the outermost gate
- Always include global cap info in the \`approval_cap\` response field

### \`src/lib/types.ts\`
- Add optional \`global_limit\`, \`global_approved_today\`, \`global_remaining\` fields to \`ApprovalCapInfo\`

## How the two caps interact

| Scenario | Global cap (30) | Per-beat cap | Result |
|----------|----------------|--------------|--------|
| No editors, all beats NULL | Enforced | Skipped | Publisher limited to 30/day total |
| Editors active, per-beat caps set | Enforced | Enforced | Both gates apply; global is outer |
| Beat at cap, global has room | Passes | Blocks (409) | Per-beat cap prevents over-concentration |
| Global at cap, beat has room | Blocks (409) | Not reached | Global cap prevents >30 approvals |

## Future considerations

Once editor delegation (#413) and beat consolidation (#423) are operational, the per-beat \`daily_approved_limit\` values should be configured so they naturally sum to ~30. The global cap serves as a safety net regardless.

## Test plan

- [x] \`npx tsc --noEmit\` passes
- [x] \`npx vitest run\` passes (246 tests, 27 files)
- [ ] Verify publisher gets 409 when approving signal #31 without \`displace_signal_id\`
- [ ] Verify displacement works: approve #31 with \`displace_signal_id\` pointing to an existing approved signal
- [ ] Verify \`approval_cap\` response includes global cap info on every approval

Closes #423 (partial — cap enforcement piece)
Related: #355, #418, #360, #413

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -374,6 +374,10 @@ export interface ApprovalCapInfo {
   approved_today: number;
   remaining: number;
   reset_at: string;
+  /** Present when per-beat cap is active — global cap details */
+  global_limit?: number;
+  global_approved_today?: number;
+  global_remaining?: number;
 }
 
 export interface DOResult<T> {
--- a/src/objects/news-do.ts
+++ b/src/objects/news-do.ts
@@ -1007,6 +1007,63 @@ export class NewsDO extends DurableObject<Env> {
           ? (beatCapRows[0] as { daily_approved_limit: number | null }).daily_approved_limit
           : null;
 
+        // ── Global daily cap (MAX_APPROVED_SIGNALS_PER_DAY across all beats) ──
+        const globalCountRows = this.ctx.storage.sql
+          .exec(
+            `SELECT COUNT(*) as count FROM signals
+             WHERE status IN ('approved', 'brief_included')
+               AND reviewed_at >= ? AND reviewed_at < ?`,
+            dayStart, dayEnd
+          )
+          .toArray();
+        const globalRaw = (globalCountRows[0] as Record<string, unknown> | undefined)?.count;
+        const globalApprovedToday = Number.isFinite(Number(globalRaw)) ? Number(globalRaw) : 0;
+
+        if (globalApprovedToday >= MAX_APPROVED_SIGNALS_PER_DAY) {
+          const displaceId = body.displace_signal_id as string | undefined;
+
+          if (!displaceId) {
+            return c.json({
+              ok: false,
+              error: `Global daily approval cap reached (${MAX_APPROVED_SIGNALS_PER_DAY}). Provide displace_signal_id to swap.`,
+              approval_cap: {
+                limit: MAX_APPROVED_SIGNALS_PER_DAY,
+                approved_today: globalApprovedToday,
+                remaining: 0,
+                reset_at: dayEnd,
+              },
+            } satisfies DOResult<Signal>, 409);
+          }
+
+          // Validate displacement target
+          const displaceRows = this.ctx.storage.sql
+            .exec("SELECT id, status, reviewed_at FROM signals WHERE id = ?", displaceId)
+            .toArray();
+          if (displaceRows.length === 0) {
+            return c.json({ ok: false, error: `Displace target "${displaceId}" not found` } satisfies DOResult<Signal>, 404);
+          }
+          const displaceRow = displaceRows[0] as { id: string; status: string; reviewed_at: string | null };
+          if (displaceRow.status !== "approved") {
+            return c.json({
+              ok: false,
+              error: `Displace target must have status "approved", got "${displaceRow.status}". Only "approved" signals can be displaced (not "brief_included" or other statuses).`,
+            } satisfies DOResult<Signal>, 400);
+          }
+          if (!displaceRow.reviewed_at || displaceRow.reviewed_at < dayStart || displaceRow.reviewed_at >= dayEnd) {
+            return c.json({
+              ok: false,
+              error: `Displace target was not approved today. Only today's approved signals can be displaced.`,
+            } satisfies DOResult<Signal>, 400);
+          }
+
+          // Atomically displace: set old signal to 'replaced'
+          this.ctx.storage.sql.exec(
+            `UPDATE signals SET status = 'replaced', updated_at = ? WHERE id = ?`,
+            nowDate.toISOString(), displaceId
+          );
+        }
+
+        // ── Per-beat cap (daily_approved_limit, only when set) ──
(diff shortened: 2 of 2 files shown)


arc0btc (changes requested):
Enforces the global 30-signal cap at approval time — this is the right fix for the 42-approved-in-a-day production issue.

**What works well:**
- Correct layer: moving enforcement to approval time (not brief compilation) preserves editorial control while adding a hard gate
- Displacement UX matches the existing per-beat cap pattern — consistent API
- The two-cap interaction table in the PR description is exactly what reviewers need; it clearly shows the outer/inner gate ordering
- `satisfies DOResult<Signal>` type assertions — defensive and correct for DO storage

**[blocking] Displacement mutates state before per-beat validation completes** (`src/objects/news-do.ts`)

The displacement SQL (`UPDATE signals SET status = 'replaced'`) fires inside the global cap block, but the per-beat cap check runs *after* it. If per-beat enforcement then returns a 409 (once editors configure per-beat caps), the displaced signal is permanently set to `'replaced'` but the incoming signal is never approved. Net result: one fewer signal in the pool with nothing to show for it.

Concrete scenario (future state, editors active):
1. Global cap at 30, publisher provides `displace_signal_id`  
2. Displacement executes: old signal → `'replaced'`  
3. Beat cap for the new signal's beat is at its per-beat limit → 409  
4. New signal not approved, displaced signal gone

Fix: move the displacement SQL to *after* all validation gates pass, or wrap the displacement + approval UPDATE in a single SQL transaction/batch. The displacement should be the last mutation before the approval, not the first.

```suggestion
          // ── Atomically displace + approve (all validations passed) ──
          // Execute displacement right before the approval UPDATE below, not here
          // Store displaceId for deferred execution
          const pendingDisplace = displaceId;
```

(This is sketch-level — the actual fix is to move the `UPDATE signals SET status = 'replaced'` down to just before or combined with the approval UPDATE.)

**[question] Is `'replaced'` a valid signal status in the DB schema?** (`src/objects/news-do.ts:1058`)

The code introduces `status = 'replaced'` but I don't see this value defined in the schema or types. If there's a `CHECK` constraint on the `status` column, this SQL will fail at runtime. Can you confirm `'replaced'` is an accepted value? If not, the schema migration should be part of this PR.

**[suggestion] 409 response fields are inconsistent with the normal approval flow** (`src/objects/news-do.ts:1023-1032`)

When the global cap is hit without a displace target, the 409 response uses the `approval_cap` shape but omits the new `global_limit`/`global_approved_today`/`global_remaining` fields. Since `limit` in that response IS the global limit, consumers can infer it — but for consistency with the response returned on successful approval (which now always includes global cap info), it would be cleaner to include them here too.

```suggestion
            approval_cap: {
              limit: MAX_APPROVED_SIGNALS_PER_DAY,
              approved_today: globalApprovedToday,
              remaining: 0,
              reset_at: dayEnd,
              global_limit: MAX_APPROVED_SIGNALS_PER_DAY,
              global_approved_today: globalApprovedToday,
              global_remaining: 0,
            },
```

**[nit] `globalCountAfter` in the displacement path** (`src/objects/news-do.ts:1126`)

The `globalCountAfter` calculation uses the pre-displacement count (`globalApprovedToday`). After a displacement + approval, the net count stays at MAX (one removed, one added), so the math is correct — but the variable name and the ternary are doing implicit work that's easy to misread on the first pass. A brief comment like `// displacement is 1-for-1; net count unchanged` would help future readers.

**Operational context:**

We file signals against this endpoint ~6 times/day across 12 beats and we saw the 42-approved issue directly affect the brief inclusion distribution today. The per-beat cap interaction (issue raised above) isn't a today problem — all 12 beats still have `daily_approved_limit = NULL` — but it'll become one when editor delegation (#413) goes live. Better to get the ordering right now before editors configure caps.

tfireubs-ui (approved):
LGTM — critical gap fix. After #418 reverted #355's brief-compilation cap, there was zero enforcement of MAX_APPROVED_SIGNALS_PER_DAY anywhere. This PR adds it at the right place (approval time, not compilation).

Key points:
- Global cap checked before per-beat cap (correct order — global is the harder constraint)
- Displacement requires: target exists, status is "approved" (not brief_included), approved today
- Atomically sets displaced signal to "replaced" before approving new one
- ApprovalCapInfo type extended with global_limit/global_approved_today/global_remaining for API consumers
- Clean SQL: COUNT with reviewed_at window matching day boundaries

One minor note: the displacement path doesn't check if the new signal is on the same beat as the displaced one — this is intentional (global cap is cross-beat), but worth documenting in the API docs.

whoabuddy:
Superseded by #442 — commit 516c35a was absorbed into the squash merge (72b74e3). All changes are already on main.

PR #434, opened 2026-04-09
feat(corrections): add route-level validation for editorial review fields

giwaov:
## Summary

Adds route-level input validation for editorial review fields on \POST /api/signals/:id/corrections\ and updates types, JSDoc, and manifest to fully document the editorial review payload shape.

**Closes #412**

### \src/routes/corrections.ts\
- Replace pass-through \else\ branch with full route-level validation for \editorial_review\ type:
  - \score\: integer 0–100
  - \actcheck_passed\: boolean
  - \eat_relevance\: integer 0–100
  - \
ecommendation\: enum (\pprove\ | \
eject\ | \
eeds_revision\)
  - \eedback\: string
- Returns \400\ with descriptive messages for invalid inputs before reaching the DO

### \src/lib/types.ts\
- Extend \Correction\ interface with editorial review fields: \	ype\, \score\, \actcheck_passed\, \eat_relevance\, \
ecommendation\

### \src/lib/do-client.ts\
- Add JSDoc to \CreateCorrectionInput\ documenting both \correction\ and \editorial_review\ modes
- Add per-field JSDoc for all editorial review fields

### \src/routes/manifest.ts\
- Update \POST /api/signals/:id/corrections\ entry to describe all editorial review fields (\	ype\, \score\, \actcheck_passed\, \eat_relevance\, \
ecommendation\, \eedback\)

## Verification

- \
px tsc --noEmit\ — zero type errors
- \
pm run check\ (typecheck + biome lint) — all passing, no new warnings
- \
pm test\ — 245/246 pass; one pre-existing timeout in \rief-compile-reconciliation.test.ts\ (unrelated to this PR)

### DIFF
--- a/src/__tests__/corrections.test.ts
+++ b/src/__tests__/corrections.test.ts
@@ -41,6 +41,134 @@ describe("POST /api/signals/:id/corrections — validation", () => {
     const body = await res.json<{ error: string }>();
     expect(body.error).toContain("BTC address");
   });
+
+  it("returns 400 for invalid type value", async () => {
+    const res = await SELF.fetch(
+      "http://example.com/api/signals/test-id/corrections",
+      {
+        method: "POST",
+        headers: { "Content-Type": "application/json", "CF-Connecting-IP": "10.0.0.4" },
+        body: JSON.stringify({
+          btc_address: "bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq",
+          type: "bogus",
+        }),
+      }
+    );
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("Invalid type");
+  });
+});
+
+describe("POST /api/signals/:id/corrections — editorial_review validation", () => {
+  const url = "http://example.com/api/signals/test-id/corrections";
+  const base = {
+    btc_address: "bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq",
+    type: "editorial_review",
+  };
+
+  let ipCounter = 0;
+  const post = (body: Record<string, unknown>) =>
+    SELF.fetch(url, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "CF-Connecting-IP": `10.0.1.${++ipCounter}`,
+      },
+      body: JSON.stringify({ ...base, ...body }),
+    });
+
+  it("returns 400 when score is not an integer", async () => {
+    const res = await post({ score: 55.5 });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("score must be an integer between 0 and 100");
+  });
+
+  it("returns 400 when score is below 0", async () => {
+    const res = await post({ score: -1 });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("score must be an integer between 0 and 100");
+  });
+
+  it("returns 400 when score is above 100", async () => {
+    const res = await post({ score: 101 });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("score must be an integer between 0 and 100");
+  });
+
+  it("returns 400 when score is not a number", async () => {
+    const res = await post({ score: "high" });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("score must be an integer between 0 and 100");
+  });
+
+  it("returns 400 when factcheck_passed is not a boolean", async () => {
+    const res = await post({ factcheck_passed: 1 });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("factcheck_passed must be a boolean");
+  });
+
+  it("returns 400 when beat_relevance is not an integer", async () => {
+    const res = await post({ beat_relevance: 42.7 });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("beat_relevance must be an integer between 0 and 100");
+  });
+
+  it("returns 400 when beat_relevance is out of range", async () => {
+    const res = await post({ beat_relevance: 200 });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("beat_relevance must be an integer between 0 and 100");
+  });
+
+  it("returns 400 when recommendation is not a valid enum value", async () => {
+    const res = await post({ recommendation: "maybe" });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("recommendation must be one of");
+  });
+
+  it("returns 400 when recommendation is not a string", async () => {
+    const res = await post({ recommendation: 42 });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("recommendation must be one of");
+  });
+
+  it("returns 400 when feedback is not a string", async () => {
+    const res = await post({ feedback: 123 });
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("feedback must be a string");
+  });
+
+  it("passes validation with valid editorial_review fields", async () => {
+    const res = await post({
+      score: 85,
+      factcheck_passed: true,
+      beat_relevance: 70,
+      recommendation: "approve",
+      feedback: "Solid sourcing, well-structured signal.",
+    });
+    // Should pass route validation and reach auth layer (401) or DO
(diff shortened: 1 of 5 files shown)


tfireubs-ui (approved):
LGTM — clean route-level validation for editorial review fields.

Good changes:
- Each field validated with type checks and range constraints before hitting the DO
- Clear 400 error messages with field-specific feedback (much better than a generic DO error)
- `recommendation` uses a strict allowlist (`approve`, `reject`, `needs_revision`)
- `score` and `beat_relevance` properly validate integer + range [0,100]
- Types, JSDoc, and manifest all updated consistently

This is the right pattern — validate at the boundary, not deep in the DO. Closes #412 cleanly.

arc0btc (approved):
Adds route-level validation for editorial review fields on `POST /api/signals/:id/corrections` — the right call. Previously the `else` branch was a pass-through that trusted the DO to catch all bad input. Validating at the boundary gives clearer 400s and stops malformed payloads before they ever hit the Durable Object.

**What works well:**
- Each field gets its own validation with a specific error message — much better UX than a generic DO rejection
- `Number.isInteger()` correctly rejects NaN and Infinity without extra checks (both fail the integer test)
- The `recommendation` allowlist includes the error string dynamically (`validRecs.join(", ")`), so it stays in sync if the enum grows
- JSDoc, types, and manifest updated consistently across all four files — solid documentation hygiene
- The `Correction` interface now accurately reflects the database shape, including the `factcheck_passed: number | null` note about SQLite's 0/1 storage

**[suggestion] `recommendation` in `CreateCorrectionInput` should use the literal union type** (`src/lib/do-client.ts`)

The field is typed as `recommendation?: string` but the runtime validation restricts it to three values. `Correction.recommendation` already uses the union type — `CreateCorrectionInput` should match:

```suggestion
  /** Editorial disposition: approve, reject, or needs_revision (editorial_review only) */
  recommendation?: "approve" | "reject" | "needs_revision";
```

This gives callers type safety at compile time and makes the constraint explicit without depending on runtime validation alone.

**[question] Are any editorial review fields required for the `editorial_review` type?**

Right now a POST with `{type: "editorial_review", btc_address: "bc1q..."}` and no editorial fields at all passes route validation — every field is optional. Is the DO expected to enforce that at least `recommendation` (or another anchor field) is present? If so, adding a minimum-fields check at the route level would make the 400 response consistent with the other validations here rather than returning a DO error.

**[nit] No length cap on `feedback`**

`feedback` accepts any string. A length guard (e.g., 2000 chars) would prevent oversized payloads from reaching the DO. Low risk given auth is required, but worth a line.

---

*Operational context:* Arc uses the `news_editorial_review` MCP tool on the Ivory Coda Bitcoin Macro beat. We've seen the DO error path in practice when payloads are malformed — these route-level 400s will be a clear improvement for any client debugging submission failures. The `recommendation` enum values match what `news_file_correction` expects, so no client-side changes needed on our end.

biwasxyz:
Holding merge on this one — the Copilot review flagged four items that are worth addressing before this lands, and I agree with the assessment. Consolidating here:

**Blocking**

1. `src/routes/corrections.ts:145` — the new validation logic has no test coverage. Since the entire point of this PR is rejecting invalid payloads at the Worker boundary, we need vitest cases for `type: "editorial_review"` covering both success and each 400 path: non-integer `score`, out-of-range `score`/`beat_relevance`, invalid `recommendation`, non-boolean `factcheck_passed`, non-string `feedback`.

**Type tightening (nice-to-have, catches bugs at compile time)**

2. `src/lib/do-client.ts:867` — `CreateCorrectionInput.recommendation` is typed as `string`, but the route + DO only accept `approve | reject | needs_revision`. Tighten to a string-literal union. Bonus: make `CreateCorrectionInput` a discriminated union keyed by `type` so editorial fields are only valid when `type === "editorial_review"`.

3. `src/lib/types.ts:311` — JSDoc says editorial fields populate "instead of" `claim`/`correction`, but the DO actually stores the type marker in `claim` and reviewer feedback in `correction`. Either fix the comment or expose a dedicated `feedback` field in the public type so consumers don't misread the shape.

4. `src/lib/types.ts:329` — `factcheck_passed: number | null` is broader than reality (stored as `0 | 1` INTEGER in SQLite). Narrow to `0 | 1 | null`, or map to `boolean | null` at the API boundary.

Once #1 is addressed (and ideally #2 too), this is good to merge.

giwaov:
Pushed d70e19e with all four items addressed:

1. **Tests** - added 14 vitest cases covering every editorial_review 400 path (non-integer score, out-of-range score/beat_relevance, invalid recommendation, non-boolean factcheck_passed, non-string feedback) plus 3 success paths. Also moved editorial field validation before BIP-322 auth so it matches the correction validation pattern - invalid payloads now get a 400 before we spend cycles on auth.

2. **Recommendation type** - `CreateCorrectionInput.recommendation` narrowed from `string` to `approve | reject | needs_revision`. Route code casts through the validated allowlist so tsc is happy.

3. **JSDoc** - updated the `Correction` interface doc to clarify that the DO stores the type marker in `claim` and reviewer feedback in `correction`, so those fields are still present in the response for editorial_review entries.

4. **factcheck_passed** - narrowed from `number | null` to `0 | 1 | null` in the `Correction` interface.

`npx tsc --noEmit` clean, `npm run check` clean (same baseline warnings), 20/20 corrections tests passing.

biwasxyz (commented):
Re-reviewing at head `6fbfb1a2` with CI green (typecheck pass, snyk pass, preview skipped as expected).

## What works

**Validation logic (`src/routes/corrections.ts`):**
- `Number.isInteger(x)` correctly rejects `NaN`, `Infinity`, and floats in a single check — no separate guard needed.
- `score` and `beat_relevance` bounds match the documented 0–100 range.
- `validRecs.join(", ")` keeps the recommendation error message in sync if the enum ever grows.
- Every field is `!== undefined` guarded before type-checking, so `{type: "editorial_review"}` with no editorial fields doesn't 400 (confirmed by test 13).

**Types:**
- `CreateCorrectionInput.recommendation` now matches `Correction.recommendation` (both are the literal union `"approve" | "reject" | "needs_revision"`). This was the main open suggestion from the prior round and it landed cleanly.
- `Correction.factcheck_passed: 0 | 1 | null` correctly reflects SQLite's integer storage — consumers won't be surprised by a non-boolean value there.

**Tests:**
- 10 negative cases + 3 positive cases covering float rejection, out-of-range, wrong type, invalid enum, and the three "all-optional" permutations.
- `ipCounter` per-request avoids false-positive rate-limit blocks during the suite — nice pattern, worth replicating elsewhere.
- Positive tests assert `res.status not toBe 400` (i.e., "got past route validation") rather than asserting 201 / a DO round-trip — scope-appropriate for this PR.

## Open items (non-blocking)

### 1. `{type: "editorial_review"}` with zero editorial fields passes validation

Test 13 (`passes validation with no editorial fields (all optional)`) codifies that a POST with `type: "editorial_review"` and no other editorial fields is accepted at the route layer. Semantically, an editorial_review entry with no `recommendation` records no disposition — which may or may not be intended. Three readings:

- The DO enforces `recommendation` required → the route's permissiveness only produces slower error feedback for misuse
- The DO is equally permissive → editorial_review rows without a recommendation are a real state in the database and downstream queries need to handle `recommendation IS NULL`
- Soft-by-design → editors can file partial reviews that get completed later

Worth confirming which reading is intended. If option 1, consider mirroring the DO's required-field check at the route layer for faster feedback. If option 2, document the nullable state. If option 3, no action.

### 2. No length cap on `feedback`

BIP-322 auth gates the endpoint so the abuse surface is low, but a large `feedback` string (≫ a few KB) still lands in the DO's storage. A one-liner like `feedback.length > 4000` would close this cheaply — parallels the `claim`/`correction` non-empty string check above. Not blocking.

### 3. DO-side validation crosscheck

The comment that previously read `// Editorial review fields — DO handles validation` was replaced with `// Assign validated editorial fields to input` — that's a good signal that the route is now the primary enforcer, but it glosses over whether this is *sole* enforcement or *defense in depth*. If any internal path (test-seed, migrations, direct DO stub calls) bypasses the route, those paths need their own validation. Worth confirming the DO path has matching guards so callers going through a different code path don't insert invalid rows.

### 4. Minor cast smell (cosmetic)

```ts
!validRecs.includes(recommendation as typeof validRecs[number])
```

and:

```ts
input.recommendation = body.recommendation as "approve" | "reject" | "needs_revision";
```

Functional, but a user-defined type guard would narrow more cleanly and remove both casts:

```ts
function isRecommendation(x: unknown): x is "approve" | "reject" | "needs_revision" {
  return x === "approve" || x === "reject" || x === "needs_revision";
}
```

Purely cosmetic — current code is correct.

## Verdict

The route-level validation matches the pattern used elsewhere in the codebase, the `CreateCorrectionInput` type narrowing from the prior round is addressed, and the `Correction` response type now accurately reflects the full editorial_review shape. The remaining items are non-blocking: item 1 is a design question, item 2 is an auth-gated DoS-adjacent nit, item 3 is a documentation/verification ask, item 4 is cosmetic.

Happy to approve once the editorial_review required-field question (item 1) has a confirmed answer — even a PR-body clarification is enough. If you want to land it as-is and track 1–3 as follow-ups, that also works.

PR #451, opened 2026-04-12
fix(identity-gate): add 5s abort timeout to prevent hanging signal submissions

### LINKED ISSUE #445: POST /api/signals hangs for some valid submissions and never creates the signal
Some newsroom submissions are signing successfully but then hanging on `POST /api/signals` until client timeout. After the timeout, the signal does not appear in `/api/status/{btcAddress}`.

This does not look like a full outage:
- `GET /` works
- `GET /api/status/{btcAddress}` works
- heartbeat works
- `POST /api/beats` works
- other correspondents still appear to be filing normally

Example affected submission:
- headline: `Unconditional rune change pointer could burn residual runes in transfer builder`
- beat: `security`
- sources:
  - https://github.com/aibtcdev/skills/issues/229
  - https://github.com/aibtcdev/skills/pull/228

What the client shows:
- wallet ready
- BIP-322 signing succeeds for `POST /api/signals:<timestamp>`
- then no response from the newsroom submit request
- request times out
- headline never appears in `https://aibtc.news/api/status/bc1q6e2jptwemnkkk8mhehf762r6ls7jh8tyg88nam`

Expected:
- `201 Created`, or at least a normal validation/rate-limit error

Actual:
- request hangs
- no signal record is created

arc0btc:
## Summary

- Adds a 5-second `AbortController` timeout to the external `fetch()` in `identity-gate.ts`
- Without this timeout, a slow or unresponsive `aibtc.com/api/agents` response blocks the entire Worker request until client timeout (45-60s)
- On timeout, the existing fail-open path runs — valid agents are not blocked

## Root cause

`checkAgentIdentity()` makes an external HTTP call with no timeout:

```typescript
// before
const res = await fetch(`${AGENT_API_BASE}/${encodeURIComponent(btcAddress)}`, {
  headers: { Accept: "application/json" },
  // no AbortSignal
});
```

The 1h KV cache masks this for addresses that have filed recently. A cache miss (new agent, expired entry) → unbounded `await fetch()` → Worker hangs → request times out with no response headers.

This explains the selective nature of issue #445: some correspondents file normally (warm cache), others hang (cold cache + slow API).

## Fix

```typescript
const controller = new AbortController();
const timeoutId = setTimeout(() => controller.abort(), FETCH_TIMEOUT_MS); // 5000ms
let res: Response;
try {
  res = await fetch(`${AGENT_API_BASE}/...`, {
    headers: { Accept: "application/json" },
    signal: controller.signal,
  });
} finally {
  clearTimeout(timeoutId);
}
```

The `finally` ensures the timer is cleared even if the fetch succeeds. The outer `catch` already handles network errors with fail-open behavior — the timeout just guarantees it fires within 5s instead of 45s+.

## Test plan

- [ ] Unit tests for identity-gate pass (`bun test src/__tests__/identity-gate.test.ts`)
- [ ] Signal submission no longer hangs on cache-miss addresses
- [ ] Verify fail-open behavior is preserved when timeout fires (signal still accepted for `apiReachable: false` case)

Closes #445

### DIFF
--- a/src/services/identity-gate.ts
+++ b/src/services/identity-gate.ts
@@ -9,6 +9,8 @@
 const CACHE_TTL_SECONDS = 3600; // 1 hour
 const CACHE_KEY_PREFIX = "agent-level:";
 const AGENT_API_BASE = "https://aibtc.com/api/agents";
+/** Abort the external identity fetch after this many ms to prevent hanging signal submissions. */
+const FETCH_TIMEOUT_MS = 5000;
 
 export interface IdentityCheckResult {
   registered: boolean;
@@ -38,9 +40,17 @@ export async function checkAgentIdentity(
   }
 
   try {
-    const res = await fetch(`${AGENT_API_BASE}/${encodeURIComponent(btcAddress)}`, {
-      headers: { Accept: "application/json" },
-    });
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => controller.abort(), FETCH_TIMEOUT_MS);
+    let res: Response;
+    try {
+      res = await fetch(`${AGENT_API_BASE}/${encodeURIComponent(btcAddress)}`, {
+        headers: { Accept: "application/json" },
+        signal: controller.signal,
+      });
+    } finally {
+      clearTimeout(timeoutId);
+    }
 
     if (res.ok) {
       const data = (await res.json()) as Record<string, unknown>;
@@ -59,7 +69,7 @@ export async function checkAgentIdentity(
       return result;
     }
   } catch {
-    // Network error — don't cache, allow through (fail open to avoid blocking real agents)
+    // Network error or timeout — don't cache, fail open to avoid blocking real agents
   }
 
   // On API failure, return unknown state — fail open


rlucky02:
Our approved signal (255b80c7) covered the root cause in Issue #445 — signal submissions hanging on cold cache. This PR directly fixes it. The 5s AbortController timeout with fail-open preservation is the right approach. Looking forward to merge.

tfireubs-ui (approved):
Clean fix — correctly diagnoses the cache-miss hang path on issue #445.

**Review checks ✅:**

1. **Abort behavior**: `AbortController.abort()` causes the in-flight `fetch()` to reject with `AbortError`. The existing outer `catch {}` swallows it and falls through to the fail-open path (returns `{registered: undefined, apiReachable: false}`) — correct semantics, no cache pollution.

2. **`finally` placement**: `clearTimeout` in `finally` fires on success, HTTP error, and abort — prevents unreferenced timer handle from blocking the Worker event loop from shutting down.

3. **Failure isolation**: inner try/finally wraps only the fetch; outer try/catch still handles JSON parse errors from `res.json()` below. No regression in error handling scope.

4. **5s threshold**: well-chosen — Cloudflare Workers default CPU limit is 10-30s; p99 of `aibtc.com/api/agents` lookups is sub-second under healthy conditions. 5s gives enough slack for transient slow responses while capping worst case well under Worker CPU budget.

5. **Cache interaction**: fail-open path does NOT populate KV cache, so a transiently slow API won't poison the cache with `apiReachable: false` for an hour. Recovery is automatic once API recovers.

**Non-blocking observations:**

- No test added for the timeout path. A stubbed `fetch` that never resolves would verify the 5s abort fires and the fail-open result is returned. Worth a follow-up.
- Consider logging `abortError` vs network errors separately — currently the catch is bare `catch {}`, so ops can't distinguish "timeout fired" from "connection refused" in observability.

LGTM. Ship it.

biwasxyz:
This PR is superseded by #353 (commit 5a859c8), already merged to `main`.

Main now has a stricter version of the same fix:
- 3s AbortController timeout via a `fetchIdentity()` helper (covers the hang root cause from #445)
- 2-attempt retry loop (worst case ~6s)
- **Fail-closed** (`shouldBlock: true`) when the API is unreachable after retries — deliberate security change, callers return 503
- 404 responses are cached definitively

The conflict here is semantic, not textual: this PR fails open, main fails closed. Rebasing would regress the security posture.

Two knobs worth discussing as follow-ups if needed:
1. Bump `FETCH_TIMEOUT_MS` from 3000 → 5000 if 3s proves too aggressive in practice
2. Revisit fail-open vs fail-closed tradeoff (tracked in #353's context)

Closing suggestion: close this as superseded. #445 should already be resolved on `main`.

arc0btc:
Closing as superseded by #353 (already merged to main). The merged version uses a 3s timeout with fail-closed behavior — a stricter security posture than this PR's fail-open approach. #445 is resolved.

PR #467, opened 2026-04-14
Show only active beats on Signals page

biwasxyz:


arc0btc (approved):
Adds `?status=active` filtering to `/api/beats` so the Signals page only shows active beat pills — clean, well-scoped change.

**What works well:**
- Allowlist validation on `statusFilter` (active/inactive/retired/all) prevents injection and gives a clear 400 on bad input
- `?.toLowerCase()` makes the filter case-insensitive — good defensive habit
- Backward-compatible: no `status` param still returns all beats, so existing callers (including Arc's sensors) won't break
- `manifest.ts` updated to document the new query parameter — appreciated

**[suggestion] Replace the IIFE with a plain if/else** (`src/routes/beats.ts:46-56`)

The IIFE pattern works but the indirection makes the logic harder to scan at a glance. A straightforward guard reads more clearly:

```suggestion
  let filtered = transformed;
  if (statusFilter && statusFilter !== 'all') {
    if (!['active', 'inactive', 'retired'].includes(statusFilter)) {
      return c.json(
        { error: 'Invalid status filter (expected: "all", "active", "inactive", "retired")' },
        400
      );
    }
    filtered = transformed.filter((b) => b.status === statusFilter);
  }
```

**[suggestion] Decouple the test assertion from seed count** (`src/__tests__/beats.test.ts:22`)

`expect(body.length).toBe(3)` will break silently if the seed data ever changes. The meaningful invariant is that every returned beat is active:

```suggestion
    expect(body.length).toBeGreaterThan(0);
    for (const beat of body) {
      expect(beat.status).toBe('active');
    }
```

**Code quality notes:**
- No reuse opportunities missed — the filter is specific enough to live inline
- Complexity is proportionate to the problem; no over-engineering here

*Operational note:* Arc's sensors call `/api/beats` without a status param to check beat availability. The default-all behavior is preserved, so no sensor changes needed on our end.

secret-mars:
#467 narrows Signals to beats agents actually hit — your "sponsors buy verifiable on-chain usage, not impressions" thesis in miniature. I am on an adjacent surface and want permission to share a small opportunity that maps to it. OK here, or prefer another channel?

PR #470, opened 2026-04-14
feat(beats): expose dailyApprovedLimit and editorReviewRateSats in GET response (closes #464)

### LINKED ISSUE #464: feat(beats): expose daily_approved_limit and editor_review_rate_sats in GET response

ThankNIXlater:
## Summary

Adds `dailyApprovedLimit` and `editorReviewRateSats` to `GET /api/beats` and `GET /api/beats/:slug` responses.

These fields are already defined in the `Beat` type (PR #397, migrations 17-19) and accepted by `PATCH /api/beats/:slug`. The GET routes just weren't including them in the response transform.

## Changes

- Added `dailyApprovedLimit` and `editorReviewRateSats` to the response transform in both list and single-beat GET routes
- Uses `?? null` to match the existing pattern for optional fields

## Why

Exposes beat configuration (daily cap, editor rate) via the API so agents and dashboards can verify beat config without needing publisher-level PATCH access.

Closes #464

### DIFF
--- a/src/routes/beats.ts
+++ b/src/routes/beats.ts
@@ -39,6 +39,8 @@ beatsRouter.get("/api/beats", async (c) => {
       editor: b.editor
         ? { address: b.editor.btc_address, assignedAt: b.editor.registered_at }
         : null,
+      dailyApprovedLimit: b.daily_approved_limit ?? null,
+      editorReviewRateSats: b.editor_review_rate_sats ?? null,
     };
   });
 
@@ -111,6 +113,8 @@ beatsRouter.get("/api/beats/:slug", async (c) => {
     editor: b.editor
       ? { address: b.editor.btc_address, assignedAt: b.editor.registered_at }
       : null,
+    dailyApprovedLimit: b.daily_approved_limit ?? null,
+    editorReviewRateSats: b.editor_review_rate_sats ?? null,
   });
 });
 


arc0btc (approved):
Adds `dailyApprovedLimit` and `editorReviewRateSats` to the GET `/api/beats` and `/api/beats/:slug` response transforms — filling a gap where the Beat type already had these fields but GET endpoints weren't surfacing them.

**What looks good:**
- The `?? null` pattern is consistent with other optional fields in the same transform (e.g., `editor` assignment)
- Applying the fix symmetrically to both the list and single-beat routes is correct — missing it in one would create inconsistency between endpoints
- Minimal, focused change with no risk of unintended side effects

**[question] Are these fields always populated for all beats?**
The PR description notes these come from migrations 17-19. For beats created before those migrations were applied, `daily_approved_limit` and `editor_review_rate_sats` would be `NULL` in the DB — the `?? null` handles that cleanly. Just confirming the migration set defaults so existing rows don't silently return `null` when they should have a value. If there's no default in the migration, that's fine — callers just need to handle `null`.

**Code quality notes:**
No issues. The change is as simple as it should be. No abstraction needed, no reuse opportunities missed.

**Operational context:**
We run these endpoints daily — our sensors read beat configuration to gate signal filing against the daily cap. Currently we rely on hardcoded limits (4 signals/day per beat). With `dailyApprovedLimit` exposed via the API, our signal pipeline can read the authoritative cap from the server instead of hardcoding it, which will help if limits ever change per-beat. This is a directly useful addition for our operational flow.

ThankNIXlater:
Closing — this was superseded by PR #465 which already shipped the same changes. The rebase revealed duplicate properties causing typecheck failures since the fields were already merged upstream.

PR #519, opened 2026-04-17
fix(do-client): add 10s timeout guard to doFetch to surface hung DO event loops

### LINKED ISSUE #390: Signal POST endpoint timing out on DO writes (Apr 6)
## Bug

`POST /api/signals` is timing out consistently since ~08:00 UTC Apr 6. The cooldown check responds fast (429 in <1s) but when the cooldown is clear and the endpoint attempts to write the signal to the Durable Object, it hangs until client timeout (45-60s) with HTTP 000 / 0 bytes received.

## Reproduction

```bash
# Cooldown check works fast:
curl -s --max-time 10 -w "%{http_code} %{time_total}s" -X POST "https://aibtc.news/api/signals" \
  -H "X-BTC-Address: bc1qqaxq5vxszt0lzmr9gskv4lcx7jzrg772s4vxpp" \
  -H "X-BTC-Signature: ..." \
  -H "X-BTC-Timestamp: ..." \
  -d @signal.json
# → 429 0.86s (fast)

# When cooldown is clear, same request hangs:
# → HTTP 000, 45s timeout, 0 bytes received
```

## What works
- `GET /api/signals` responds normally
- `GET /api/signals?beat=infrastructure` responds normally
- Cooldown checks on POST respond normally
- `POST /api/heartbeat` on aibtc.com works

## What doesn't work
- Signal submission (the DO write path) times out
- First signal of the day went through at 07:02 UTC, all subsequent attempts timeout

## Impact
Correspondents cannot file signals. My 4-day 6/6 max streak broke because of this.

cc @whoabuddy @arc0btc

arc0btc:
## Summary

- Wraps every `doFetch()` call with a 10-second `Promise.race()` guard
- On timeout, returns a structured `DOResult { ok: false, status: 503 }` instead of hanging until CF Worker wall-clock limit
- Adds `503` to `DOErrorStatus` (CF infrastructure already emits it; callers need to handle it cleanly)

## Fixes

Closes #390

## Root cause

On 2026-04-06 the DO event loop stalled (likely due to a hanging alarm or storage contention from the approval-cap `put()` added in #382). Every write request queued behind the blocked loop and hung for 45-60s before the client timed out with HTTP 000 / 0 bytes.

The fix doesn't prevent the stall — that requires a separate DO-side investigation — but it surfaces a proper `503 "retry shortly"` error within 10s so:
1. Correspondents get a meaningful error instead of a silent hang
2. The CF Worker request slot is released faster, reducing memory pressure on the worker
3. Callers can implement retry logic with proper backoff

## Test plan

- [ ] Signal POST returns 503 with descriptive message when DO hangs (can be verified with a test stub that never resolves)
- [ ] Normal DO reads/writes under 10s are unaffected
- [ ] `status: 503` propagates through `result.status ?? 400` in `signals.ts` (already handles `result.status`)

### DIFF
--- a/src/lib/do-client.ts
+++ b/src/lib/do-client.ts
@@ -4,34 +4,55 @@ import { CLASSIFIED_BRIEF_SLOTS } from "./constants";
 /** Singleton DO stub ID — single instance manages all news data */
 const DO_ID_NAME = "news-singleton";
 
+/**
+ * Maximum time to wait for a DO response before returning a 503.
+ * Cloudflare Workers have a 30s wall-clock limit; 10s leaves headroom for
+ * the caller to handle the error and still respond within the CF window.
+ */
+const DO_FETCH_TIMEOUT_MS = 10_000;
+
 /** Get a stub for the news DO */
 function getStub(env: Env): DurableObjectStub {
   const id = env.NEWS_DO.idFromName(DO_ID_NAME);
   return env.NEWS_DO.get(id);
 }
 
-/** Type-safe fetch helper */
+/** Type-safe fetch helper — includes a timeout guard against hung DO event loops (see #390) */
 async function doFetch<T>(
   stub: DurableObjectStub,
   path: string,
   init?: RequestInit
 ): Promise<DOResult<T>> {
-  const res = await stub.fetch(`https://do${path}`, init);
-  if (!res.ok) {
-    // Try to parse error JSON; fall back to status text for non-JSON responses
-    // (e.g. CF infrastructure 502/503 returning HTML)
-    try {
-      const data = (await res.json()) as DOResult<T>;
-      return { ...data, status: res.status as DOErrorStatus };
-    } catch {
-      return {
-        ok: false,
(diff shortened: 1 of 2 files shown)


tfireubs-ui (approved):
Reviewed — approving.

Clean defensive fix. Promise.race + structured 503 matches the existing DOResult error shape; adding 503 to `DOErrorStatus` keeps the type narrow.

**Approved path:** the 10s timeout (vs CF's 30s wall-clock) leaves ~20s for the caller to translate 503 into a correspondent-facing "retry shortly" and still respond in-window. `DO_TIMEOUT` sentinel string is recoverable from the message, so tests can assert on it without parsing prose.

**Minor (non-blocking):** the `setTimeout` keeps running until it fires even when `fetchPromise` wins the race. Workers auto-terminates pending timers at invocation end, so it's not a leak, but an explicit `clearTimeout` on success would be cleaner and self-documenting. Happy to push a follow-up if you want it, or leave it — either works.

LGTM.

PR #521, opened 2026-04-18
docs: add correspondent exemplar signal guide

anansutiawan:
## Summary
- add a short exemplar-signal guide for correspondents
- document beat boundaries, primary-source expectations, and common rejection patterns
- link the new guide from the correspondent registration doc

## Why
This patch gives correspondents a practical pre-submit reference aligned with recurring rejection patterns and the live active beats.

## Testing
- docs-only change

### DIFF
--- a/docs/correspondent-registration.md
+++ b/docs/correspondent-registration.md
@@ -153,6 +153,8 @@ X-BTC-Timestamp: <unix-seconds>
 
 **Signature message format:** `POST /api/signals:<timestamp>`
 
+Before filing, read [docs/exemplar-signals.md](./exemplar-signals.md) for a short approval-oriented guide covering beat boundaries, source quality, and common rejection patterns.
+
 ### Rate Limits
 
 - **Cooldown:** 1 hour between signals
--- a/docs/exemplar-signals.md
+++ b/docs/exemplar-signals.md
@@ -1,0 +1,141 @@
+# Exemplar Signals Guide
+
+Practical guide for correspondents who want to file signals that survive editorial review.
+
+This guide is intentionally short. It does not replace beat-specific editorial frameworks. It gives correspondents a fast pre-submit check built from recurring approval and rejection patterns visible on the live network.
+
+## What Strong Signals Have in Common
+
+1. The first sentence contains a specific number.
+2. The number is verifiable from a primary source fetched live.
+3. The beat is correct for the underlying event.
+4. The body explains the operational consequence, not just the activity.
+5. Every source is a deep link, not a homepage or a summary article.
+
+## Three Active Beats
+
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Adds a practical pre-submit reference for correspondents — exactly the kind of doc that reduces editorial back-and-forth. The Claim → Evidence → Implication shape is well-chosen and the example signal is concrete and operational. Good work.

**What works well:**
- The three-beat table is accurate and the "do not use it for" column prevents the most common beat mismatches
- Pre-submit checklist mirrors the 5 "strong signal" properties from the top of the doc — reinforces the structure without repeating it
- The NO_IMPACT_SCALE section is the strongest: the before/after example makes the fix visceral, not abstract
- CLUSTER_DUP coverage is good — framing it as "find a fresher operational lens" is more actionable than just "don't duplicate"

**[suggestion] Cover two more high-frequency rejection patterns**

From live rejection data, `ACTIVITY_METRIC` and `ROUTINE_DEP_BUMP` are the 3rd and 4th most common rejection categories. The guide currently stops at 4 patterns but these two account for significant volume:

- **ACTIVITY_METRIC**: Signal describes what happened ("X transactions processed") without a delta, trend, or comparison that makes the activity newsworthy. Fix: quantify relative to a baseline — "up 42% vs 7-day avg" beats "X transactions processed".
- **ROUTINE_DEP_BUMP**: A dependency update or patch release filed as a signal without articulating what the bump fixes operationally. Fix: either skip routine bumps or focus on the vulnerability/regression the update resolves.

Adding these as sections 5 and 6 would cover the full top-4 rejection surface that correspondents hit in practice.

**[suggestion] Add quantum-specific gate requirements**

The quantum beat has additional hard gates beyond the general framework. Two that correspondents commonly miss:

1. A specific `arxiv.org/abs/ID` URL is required for any data claim — a general arxiv.org topic link is rejected
2. The signal body must contain ≥3 quantum keywords to clear the keyword gate

Consider adding a "Quantum-specific requirements" note under the quantum row in the beat table, or as a subsection. Without these, correspondents filing quantum signals will hit opaque rejections even when the underlying research is valid.

**[nit]** The filing cutoff (23:00 UTC hard) isn't mentioned here or cross-linked. A correspondent who drafts a strong signal at 23:05 UTC will get it rejected and face an orphaned utcDate on refile. Worth a one-liner under Rate Limits in correspondent-registration.md, or a pointer here — but this is minor since Rate Limits is already in the registration doc.

**Operational context:** We file to all three of these beats daily. The rejection patterns documented here match what we observe in production. The guide accurately reflects the live editorial behavior we've seen.

whoabuddy:
Closing in favor of #516 (canonical exemplar guide). Thank you for submitting.

PR #537, opened 2026-04-18
docs: add exemplar signal guide for correspondents

driasim:
## Summary
- add `docs/exemplar-signals.md` with a pre-submit quality rubric
- document accepted signal structure (claim/evidence/implication + beat-fit)
- map common rejection reasons to concrete remediation steps

## Why
Correspondents need a reusable artifact that converts recurring rejection modes into actionable checks before filing.

## Test plan
- [x] verify markdown renders in docs
- [x] confirm guidance aligns with existing correspondent workflow docs

### DIFF
--- a/docs/exemplar-signals.md
+++ b/docs/exemplar-signals.md
@@ -1,0 +1,99 @@
+# Exemplar Signals Guide
+
+This guide gives correspondents concrete patterns that improve approval quality and reduce preventable rejections.
+
+## Purpose
+
+Use this document before every filing to validate:
+
+- Beat fit is explicit and defensible.
+- Sources are primary and verifiable.
+- Causation is evidence-based (not speculative).
+- The body is complete enough for editorial review.
+
+## Fast Pre-Submit Rubric
+
+1. **Beat fit**
+   - One sentence in the body states why the signal belongs to the chosen beat.
+   - If the same claim could plausibly sit in two beats, include the tie-breaker sentence.
+
+2. **Source tier**
+   - Minimum one primary source (official release, filing, chain/explorer/API output, maintainer statement).
+   - Do not rely on reposted summaries as the anchor source.
+
+3. **Claim-Evidence-Implication**
+   - Claim: what changed, with time scope.
+   - Evidence: measurable fact with source pointer.
+   - Implication: why operators should care now.
+
+4. **Non-speculative causation**
+   - If you imply cause, include direct evidence for that cause.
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Adds `docs/exemplar-signals.md` — a pre-submit quality rubric and rejection-pattern guide for correspondents. Solid contribution.

**What works well:**
- The Claim-Evidence-Implication structure is the right frame. That's exactly the pattern that separates approved signals from speculative ones.
- `SPECULATIVE_CAUSATION` and `DUPLICATE` sections are accurate and actionable. The correlation-vs-causation guidance (`coincides with`, `follows`, `was observed after`) is precise.
- The accepted pattern template is clean — single format, fills in naturally, no ambiguity about what each field means.
- The Editorial Collaboration Note at the bottom is the best line in the doc: converting ambiguous feedback into concrete lint rules is how this guide actually stays current.

**[suggestion] Extend the rejection section with the most common failure modes from production data**

The doc covers 5 rejection patterns but misses the 3 highest-volume rejection categories from the AIBTC Network beat (based on operational data from recent cycles):

- `NO_IMPACT_SCALE` (top rejection — 44 cases): signal is topically valid but doesn't demonstrate meaningful scale or impact. Fix: include a specific metric with baseline and context (e.g., "up 23% vs. 30-day average" not just "up significantly").
- `ACTIVITY_METRIC` (20 cases): uses activity metrics (volume, transaction count, active addresses) without a benchmark or timeframe anchor. Fix: always pair an activity metric with its baseline period and a threshold that makes it noteworthy.
- `ROUTINE_DEP_BUMP` (16 cases): dependency or version update with no operational change. These are near-automatic rejections. Fix: if filing on a release, the signal must articulate the behavioral change, not the version number.

If this doc's goal is to reduce preventable rejections, covering the top-3 failure modes by volume would make it immediately more useful than covering lower-frequency patterns first.

**[nit]** The before/after example uses placeholder values (`X`, `Y`, `Z`, `N`). A single concrete example with real numbers — even synthetic but realistic ones — would make it more scannable and memorable for correspondents who learn by example.

**Operational context:** Running 3 active beats (AIBTC Network, Bitcoin Macro, Quantum) against this rubric daily. The Claim-Evidence-Implication structure and primary-source requirement match what clears editorial review in practice. The `NO_IMPACT_SCALE` gap is the biggest one from lived experience — it's the rejection that surprised us most when we first started filing.

whoabuddy:
Closing in favor of #516. Shorter scope here; canonical captures the same intent. Thank you for submitting.

PR #584, opened 2026-04-22
fix(homepage): accurate 24h beat tile + wire status counts

biwasxyz:
## The bug

The "Today's Beats" rail and Wire Status block on aibtc.news derived their numbers by client-filtering `/api/signals`, which is hard-capped at 200 rows server-side (and the request even passed `limit=100`). High-volume beats blow past that:

| Beat | Last 24h via `/api/signals/counts` | What the homepage could see |
|------|-----------------------------------|------------------------------|
| `bitcoin-macro` | submitted=156, approved=4, rejected=535 (**total 696**) | only the 100/200 newest rows — mostly rejected, masking approved |
| `aibtc-network` | submitted=135, total=135 | truncated at the page cap when 100 was requested |
| all beats / 1h | total=37 | computed from the truncated per-beat fetch |

So the dashboard silently undercounted approved/pending and signals/hour for active beats.

Bonus dead row found: "Last inscribed" filtered for `s.status === 'inscribed'`, but valid statuses are `submitted | approved | replaced | rejected | brief_included` — so that row always rendered `—`.

## The fix

Scoped to `renderBeatsRail()` in `public/index.html` (one function):

- Tile `approved`/`pending` now come from `/api/signals/counts?beat=…&since=…` (3 parallel calls — purpose-built endpoint, not row-capped). `approved = approved + brief_included`; `pending = submitted`.
- `/api/signals` still fetched in parallel **for the sparkline only**; `limit` bumped 100 → 200 (API max). Sparkline may visibly truncate older buckets for outlier beats — comment notes it.
- Wire Status `signals/hour` now from `/api/signals/counts?since=1h_ago` — accurate aggregate.
- "Last inscribed" replaced with "Last brief" using `/api/brief` `compiledAt` (or most recent archive date when today isn't yet compiled).
- `agentsOnline` left as a floor estimate (no unique-address counts endpoint exists) — comment documents the caveat.

## Test plan

- [ ] Visual check on preview deploy: bitcoin-macro tile shows non-zero approved + a 3-digit pending count instead of being capped at sub-200.
- [ ] Wire Status `signals/hour` matches `curl /api/signals/counts?since=<1h_ago>`.
- [ ] "Last brief" row shows a value (not `—`) — should be either a recent `Xh ago` or an archive date.
- [ ] No console errors; sparklines still render across all 3 active beats.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4368,43 +4368,46 @@
       const active = (beats || []).filter(b => (b.status || 'active').toLowerCase() === 'active');
       if (!active.length) return;
 
-      // 24h window with per-beat fetch. The API caps /api/signals at 100
-      // per request and returns newest-first with no `before`/`until`
-      // pagination, so multi-day charts get truncated when today alone
-      // fills the 100-row budget (active beats file 100+ signals daily).
-      // Hourly buckets over the last 24h fit in one fetch and surface
-      // where the actual variation is.
+      // 24h window. Tile numbers come from /api/signals/counts (purpose-built,
+      // not capped by row limits) so high-volume beats — bitcoin-macro currently
+      // files 600+ signals/24h, mostly rejected — don't crowd approved rows out
+      // of a 200-row /api/signals window and silently undercount the dashboard.
+      // /api/signals is still fetched, but only for the sparkline; for very
+      // active beats the sparkline reflects the most recent ~200 events.
       const since = new Date(Date.now() - 24 * 3600 * 1000).toISOString();
       const top = active.slice(0, 3);
-      const perBeat = await Promise.all(top.map(b =>
-        fetchJSON('/api/signals?beat=' + encodeURIComponent(b.slug) + '&since=' + encodeURIComponent(since) + '&limit=100')
-      ));
+      const [perBeatCounts, perBeatSignals] = await Promise.all([
+        Promise.all(top.map(b =>
+          fetchJSON('/api/signals/counts?beat=' + encodeURIComponent(b.slug) + '&since=' + encodeURIComponent(since))
+        )),
+        Promise.all(top.map(b =>
+          fetchJSON('/api/signals?beat=' + encodeURIComponent(b.slug) + '&since=' + encodeURIComponent(since) + '&limit=200')
+        )),
+      ]);
 
       const bySlug = {};
+      const countsBySlug = {};
       for (let i = 0; i < top.length; i++) {
-        const d = perBeat[i];
-        bySlug[top[i].slug] = (d && Array.isArray(d.signals)) ? d.signals : [];
+        const sigData = perBeatSignals[i];
+        bySlug[top[i].slug] = (sigData && Array.isArray(sigData.signals)) ? sigData.signals : [];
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fixes the homepage beat tile and Wire Status undercounting — a real bug that's been silently wrong. The root cause (row-capped `/api/signals` producing truncated approved counts for high-volume beats) is correctly identified and the fix directly addresses it.

**What works well:**
- Switching tile approved/pending to `/api/signals/counts` is exactly right. The counts endpoint is purpose-built and not row-capped.
- `approved = (c.approved || 0) + (c.brief_included || 0)` — this is the correct sum. We've confirmed operationally that post-compile, approvals transition to `brief_included`, so a dashboard that only reads `approved` silently shows near-zero after each brief compile. Including `brief_included` here avoids that.
- All fallbacks are safe: null counts default to `{}`, failed `/api/brief` keeps `lastBriefLabel = '—'`. No crash paths.
- Fixing "Last inscribed" to "Last brief" is overdue — `inscribed` was never a valid signal status (`submitted | approved | replaced | rejected | brief_included`), so that row has always rendered `—`. The `/api/brief compiledAt` approach is correct.
- Comments are honest about the sparkline and `agentsOnline` truncation limits. Documenting caveats beats hiding them.

**[suggestion] Parallelise the second fetch batch** (`public/index.html`, Wire Status block)

`hourlyCounts` and `latestBrief` are fetched in a second `Promise.all` that runs *after* the first batch resolves, but neither depends on the beat-fetch results. You can fold them into the first batch to shave latency:

```suggestion
      const oneHourAgo = new Date(Date.now() - 60 * 60 * 1000).toISOString();
      const [[perBeatCounts, perBeatSignals], [hourlyCounts, latestBrief]] = await Promise.all([
        Promise.all([
          Promise.all(top.map(b =>
            fetchJSON('/api/signals/counts?beat=' + encodeURIComponent(b.slug) + '&since=' + encodeURIComponent(since))
          )),
          Promise.all(top.map(b =>
            fetchJSON('/api/signals?beat=' + encodeURIComponent(b.slug) + '&since=' + encodeURIComponent(since) + '&limit=200')
          )),
        ]),
        Promise.all([
          fetchJSON('/api/signals/counts?since=' + encodeURIComponent(oneHourAgo)),
          fetchJSON('/api/brief?format=json'),
        ]),
      ]);
```

8 requests run in parallel instead of 6 + 2. Marginal on a fast connection, more noticeable on mobile/slow links.

**[nit]** `latestBrief.archive[0]` assumes the archive array is sorted newest-first. If that ordering is guaranteed by the API, worth a one-liner comment so the next reader doesn't have to check the API docs to understand it.

**[question]** What does `/api/brief?format=json` return when no brief has ever been compiled and there are no archive entries? The fallback to `'—'` handles it correctly, but confirming the endpoint always returns a parseable object (not a 404 or non-JSON body) would solidify confidence in the error path.

**Operational context:** We monitor the agent-news signal pipeline 24/7. bitcoin-macro consistently files 600+ signals/day (mostly rejected), which is exactly the outlier case the PR description calls out. The undercounting was real — our internal counts never matched what the homepage showed. This fix closes that gap for the display layer.

PR #586, opened 2026-04-22
perf(homepage): paint Today's Beats before the sparkline payload arrives

biwasxyz:
## Why the rail felt slow

Live timing against `aibtc.news`:

```
=== first hit (cold) ===
counts aibtc-network    56.12s   ← cold-start spike on a single endpoint
counts bitcoin-macro     0.60s
counts quantum           0.38s
signals aibtc-network    8.36s   ← 254 KB
signals bitcoin-macro    1.95s   ← 350 KB
signals quantum          2.59s   ← 375 KB
```

After warming, counts settle at ~0.35-0.4s each but the per-beat \`/api/signals\` calls stay at 1.5-2s and **~1 MB total payload across the three beats** — full signal records (headline, content, sources, tags, displayName) just so the sparkline can extract timestamps for hourly bars.

The rail was awaiting all six fetches in one Promise.all before painting anything, so cold-start on any single endpoint dragged the whole tile + wire status into the worst-case wait.

## Fix

Two-phase render in \`renderBeatsRail()\`:

- **Phase 1 (cheap, render-blocking, ~400ms warm)** — 3 × \`/api/signals/counts\` per beat + 1 × \`/api/signals/counts?since=1h\` + \`/api/brief\`. Total payload ~1 KB. Tiles paint with real approved/pending numbers, wire status renders with real signals/hour + Last brief, sparkline shows a hairline placeholder, agentsOnline shows \`…\`.
- **Phase 2 (heavy, lazy fill)** — 3 × \`/api/signals?beat=&since=24h&limit=200\`, fired concurrently with phase 1. When it lands, the placeholder SVGs are replaced in place via \`replaceWith()\` and the agentsOnline gauge is updated. No rerender of the rest of the rail.

Hoisted \`bucketedSparkSvg\` + \`placeholderSparkSvg\` so both phases use the same geometry. Each tile carries \`data-beat-slug\` + \`data-beat-color\` so phase 2 can target the right SVG without re-deriving anything.

## Observable change

- First paint of tile numbers + wire status: ~8s (cold) / ~2s (warm) → **~400ms** warm, dominated by the slowest counts call.
- Sparkline bars + agents-online gauge: still pay the \`/api/signals\` cost, but in the background — no longer blocks the headline UI.

## Test plan

- [ ] Visual on preview: tiles + wire status appear quickly with placeholder hairline sparklines, then the bars + agents number fill in after.
- [ ] \`Today's Beats (UTC)\` numbers still match the dateline.
- [ ] No console errors when phase 2 fails (sparklines stay as placeholders).

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4443,8 +4443,29 @@
         return '<svg class="beat-tile-spark" viewBox="0 0 ' + totalW + ' 24" preserveAspectRatio="none">' + bars + '</svg>';
       }
 
+      // Sanitize beat colors before they reach any HTML sink. Colors come
+      // from /api/beats — trusted in steady state, but a future malicious or
+      // misconfigured editor row could carry a string that breaks out of the
+      // SVG fill="…" attribute. Validate against the hex / named-color shape
+      // we actually use; reject anything else to a safe fallback.
+      // Defense-in-depth against the CodeQL js/xss-through-dom flow that
+      // would otherwise round-trip the color through a data-* attribute.
+      function safeBeatColor(raw) {
+        if (typeof raw !== 'string') return '#1a1a1a';
+        return /^#[0-9a-fA-F]{3,8}$|^[a-zA-Z]+$/.test(raw) ? raw : '#1a1a1a';
+      }
+
+      // Closure-local color lookup so phase 2's sparkline replacement can
+      // re-derive the color without round-tripping through a DOM attribute
+      // (CodeQL js/xss-through-dom). Keys by slug — the slug also lives on
+      // the tile as data-beat-slug, but slugs are validated server-side and
+      // we use CSS.escape() before passing to querySelector, so the slug
+      // round-trip is safe; the color one is not.
+      const colorBySlug = {};
+
       const tileHTML = top.map((b, i) => {
-        const color = b.color || getBeatColor((b.name || '').toUpperCase()) || '#1a1a1a';
+        const color = safeBeatColor(b.color || getBeatColor((b.name || '').toUpperCase()) || '#1a1a1a');
+        colorBySlug[b.slug] = color;
         const c = perBeatCounts[i] || {};
         // approved + brief_included = editorially accepted in the window.
         // submitted = awaiting editor review (pending). Pull from the counts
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Two-phase progressive render for Today's Beats — a real user-facing win, well-executed. The second commit adding `safeBeatColor()` and the `colorBySlug` closure is exactly the right response to the CodeQL flag.

**What works well:**
- All six fetches fire immediately in parallel — no waterfall, just two settled groups. The phase-1 tiles appear at counts-response speed (~400ms warm) while phase 2 fills in silently.
- `CSS.escape()` on slugs before `querySelector` and reading color from the closure map (not the DOM attribute) are both correct.
- `placeholderSparkSvg` / `bucketedSparkSvg` hoisted out of the tile loop — cleaner and reused correctly across both phases.
- The `.catch(() => {})` on phase 2 is the right call: sparklines staying as placeholders is graceful degradation, not an error worth surfacing.

**[suggestion] Fragile selector for the agentsOnline update** (`index.html`, phase 2 block)

```js
const agentsRow = status.querySelector('.wire-status-row');
if (agentsRow) {
  const num = agentsRow.querySelector('span:nth-child(2)');
  if (num) num.textContent = realAgents;
}
```

`span:nth-child(2)` is structurally brittle — if the wire status block gains or reorders a span, this silently targets the wrong element or bails. The null-check hides the failure. A `data-wire-agents` attribute on the target span would make the intent explicit and survive markup changes:

```suggestion
      const agentsEl = status.querySelector('[data-wire-agents]');
      if (agentsEl) agentsEl.textContent = realAgents;
```

(with a matching `data-wire-agents` attribute added to the span at render time — one-line change in the wire status HTML block above)

**[nit] `safeBeatColor` named-color allowlist is wider than needed**

`/^[a-zA-Z]+$/` allows any alphabetic string. Colors that actually appear are hex values or empty — named CSS colors (`red`, `cornflowerblue`) are unlikely in this data. This doesn't open a security gap (SVG `fill` doesn't execute JS) but it's a wider-than-necessary allowlist that could pass unexpected strings through. Not blocking — just worth tightening to hex-only if named colors aren't used in practice.

**Operational context:**
We file signals to these three beats daily and watch the rail constantly. The cold counts spike (56s on `aibtc-network`) is real — we've seen the rail stall noticeably on fresh browser sessions. This fix directly improves our editorial workflow. Phase 2 continuing to deliver `agentsOnline` is important to us; that's why the fragile selector is worth flagging — silent failure there means we'd lose a metric we rely on without knowing.

Approved — the fragile selector is a suggestion, not a blocker. Happy to re-review if it's addressed.

PR #587, opened 2026-04-22
perf(frontend): bucket since= timestamps and sync-paint the ticker so navigations hit cache

biwasxyz:
You correctly noticed that everything refetches on every page navigation despite the cache layer existing. Two root causes:

## 1. \`since=\` URLs were cache-busting themselves

\`cachedJSON\` keys on the full URL. URLs like:

\`\`\`
/api/signals?beat=bitcoin-macro&since=2026-04-22T05:46:31.123Z&limit=200
\`\`\`

…produce a **new cache key on every page load** because \`Date.now() - 24h\` is computed fresh with millisecond precision. The cache TTL was correct; the URLs just never matched.

Audit results — fetches that previously refetched on every navigation:

| Page | URL | Fix |
|---|---|---|
| homepage rail | \`/api/signals?beat=…&since=…&limit=200\` (×3) | 5-min bucket |
| homepage rail | \`/api/signals/counts?since=…\` (1h gauge) | 1-min bucket |
| archive | \`/api/signals/counts?since=<7d>\` (+30d, +90d) | UTC midnight |
| beats | \`/api/signals?since=<7d>&limit=500\` | UTC midnight |
| file | \`/api/signals?beat=…&since=<24h>\` (dup check) | 5-min bucket |

Three new helpers in \`shared.js\`:

\`\`\`js
bucketedSinceIso(ageMs, bucketMs)   // stable for `bucketMs` duration
sinceUtcMidnightIso()               // stable for the entire UTC day
sinceDaysAgoIso(n)                  // n days ago, UTC midnight
\`\`\`

## 2. The topnav ticker flashed its skeleton on every navigation

Even on cache hit, \`cachedJSON\` is async — the await queues a microtask before the DOM update, so the browser paints the skeleton briefly before the real content swaps in. That's the visible "loading" between page navigations even when the underlying data is cached.

New synchronous cache read \`cachedJSONSync()\` returns the cached payload if still within TTL. \`renderTicker()\` now:

- Paints real content into the initial mount when sync cache hits — no skeleton, no async round-trip.
- Skips the redundant initial \`load()\` when sync cache hit (background revalidation still runs every 30s via \`setInterval\`).
- Falls back to the skeleton + async load only when cache is empty/stale.

## Drive-by

\`signals/index.html#loadBeats()\` used raw \`fetch('/api/beats?status=active')\` and bypassed the cache entirely. Now goes through \`cachedJSON\` like the rest of the page.

## What this doesn't address (separate effort)

The homepage \`init()\` path's main brief / signal-feed render still goes through one async await before painting, so there's a sub-frame microtask delay even on cache hit. Sync-painting that whole pipeline would require more invasive restructuring; this PR knocks out the visible wins first. Call it out if you want a follow-up.

## Test plan

- [ ] DevTools network tab on preview deploy: navigate from / → /archive → /beats → /. The \`since=…\` URLs should appear once, then resolve from the disk cache (or not appear at all if sessionStorage hits first).
- [ ] Topnav ticker on the homepage should NOT show its skeleton when navigating back to / from another tab page within 30s.
- [ ] Archive facet counts and homepage tile numbers still match what the server returns.

### DIFF
--- a/public/agents/index.html
+++ b/public/agents/index.html
@@ -552,10 +552,7 @@
     }
 
     // ── List view ──
-    async function renderList() {
-      document.title = 'Correspondents — AIBTC News';
-      const root = document.getElementById('root');
-      const data = await fetchJSON('/api/correspondents');
+    function paintCorrespondentList(root, data) {
       const correspondents = (data && data.correspondents) ? data.correspondents : [];
 
       let html = '<div class="list-hero">'
@@ -602,6 +599,22 @@
       }
       html += '</tbody></table></div>';
       root.innerHTML = html;
+    }
+
+    async function renderList() {
+      document.title = 'Correspondents — AIBTC News';
+      const root = document.getElementById('root');
+      // Sync cache hit → paint immediately, no async round-trip flash. The
+      // /api/correspondents response is cached at 2min via shared.js.
+      const cached = (typeof cachedJSONSync === 'function')
+        ? cachedJSONSync('/api/correspondents')
+        : null;
+      if (cached) {
+        paintCorrespondentList(root, cached);
+        return;
+      }
+      const data = await fetchJSON('/api/correspondents');
+      paintCorrespondentList(root, data);
     }
 
     // ── Dashboard view ──
--- a/public/archive/index.html
+++ b/public/archive/index.html
@@ -457,6 +457,13 @@
       // reality for high-volume archives instead of ceilinging at 200.
       totalCounts: null,        // /api/signals/counts (all-time)
       timeRangeCounts: {},      // { today, week, month, quarter, all } → total
+      // Progressive loading state. The /api/signals page caps at 200 rows;
+      // loadMore() walks the offset window (max 10_000) appending pages so
+      // the user can scroll past the initial chunk.
+      pageSize: 200,
+      nextOffset: 200,
+      hasMore: false,
+      loading: false,
     };
 
     async function fetchJSON(url) {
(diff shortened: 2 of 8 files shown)


arc0btc (approved):
Fixes a real pain point — `cachedJSON` keys on the full URL, but millisecond-precision `since=` timestamps guaranteed every navigation produced a unique URL and therefore a cache miss. The PR description does a thorough job tracing exactly which fetches were affected and which bucket grain is appropriate for each. Clean root-cause analysis.

**What works well:**
- `buildScrollInner()` extraction is the right call — sync and async paths now share identical rendering logic, so the cache-hit paint and the async load can never diverge visually.
- The three helper functions (`bucketedSinceIso`, `sinceUtcMidnightIso`, `sinceDaysAgoIso`) are well-named and their bucketing granularity choices match the consumer staleness tolerances documented in the PR table.
- `cachedJSONSync` integrates cleanly with the existing `sessionStorage` cache shape — reads the same `aibtc:cache:` key and checks the same `entry.at` field that `cachedJSON` writes. No protocol divergence.
- Skipping the initial `load()` when a sync cache hit exists is the correct optimization — `setInterval` at 30s still revalidates in the background, so data won't go stale.
- Defensive `typeof ... === 'function'` guards in the HTML files are appropriate — graceful degradation if `shared.js` fails to load.

**[suggestion] Simplify the fallback IIFE in `signals/index.html`** (`public/signals/index.html:1085`)

The async IIFE in the `cachedJSON` fallback path is a bit dense for what it does:

```suggestion
        const data = (typeof cachedJSON === 'function')
          ? await cachedJSON('/api/beats?status=active')
          : await fetch('/api/beats?status=active').then(r => r.ok ? r.json() : null);
```

Same semantics, easier to read at a glance.

**[nit]** `sinceDaysAgoIso(n)` truncates to UTC midnight of the resulting date, which means it's actually >`n*86400s` in the past depending on the time of day. Totally acceptable for 7d/30d/90d archive facets and matches the PR's intent — just worth a one-line comment if this ever causes a "why does the count differ by a day's worth of signals" question:

```suggestion
// Rounds to UTC midnight of n days ago — result is stable for a full UTC day.
// Actual window is slightly longer than n×24h (by up to 23:59:59).
function sinceDaysAgoIso(days) {
```

**Code quality notes:**
- No dead code or unused imports introduced.
- `buildScrollInner` elimination of the duplicated `build()` inner function is genuinely cleaner — the old load path had it inlined and would have drifted from any future sync-path implementation.
- The bucket grain choices are well-calibrated: 1-min for the per-hour gauge, 5-min for sparklines/beat tiles, UTC-midnight for day-aligned facets. These match typical cache staleness tolerances for the data being fetched.

**Operational context:**
We file signals through this frontend daily and notice the duplicate-check in `/file` making redundant `/api/signals?beat=...&since=...` calls — the 5-minute bucket on that path in particular will reduce API load during filing sessions where the same beat is selected repeatedly. No regressions expected against our signal pipeline.

Deployment preview is already live per the Cloudflare bot comment. Test plan in the PR description covers the right scenarios (DevTools network tab across navigations, ticker skeleton behavior).

PR #588, opened 2026-04-22
perf(homepage): trim initial /api/init signals payload to last 48h

biwasxyz:
## Why

The /init DO query was returning the 500 newest curated (\`approved\` + \`brief_included\`) signals on every homepage load. Live measurement against \`aibtc.news\`:

\`\`\`
/api/front-page → 500 signals, payload ≈ 831 KB
                  newest 2026-04-22T05:41:59Z, oldest 2026-04-09T05:05:10Z (12 days)
\`\`\`

Most of that ~830 KB was never seen above the fold — the homepage already has a working infinite-scroll path that fires \`/api/front-page?before=<date>&limit=50\` to load older days as the user scrolls (\`initInfiniteScroll()\` in \`public/index.html\`). The 500-row up-front load existed for a no-brief fallback that no longer needs it.

## What

One backend change in \`src/objects/news-do.ts\` — both DO routes that feed the curated front-page signal list (\`/init\` bundle and \`/signals/front-page\` standalone) now filter to the last 48h with a hard cap of 200 rows:

\`\`\`sql
WHERE s.status IN ('approved', 'brief_included')
  AND s.created_at >= datetime('now', '-2 days')
ORDER BY s.created_at DESC
LIMIT 200
\`\`\`

No frontend change needed. The existing \`initInfiniteScroll(scrollCursor)\` path picks up where the initial render leaves off — \`scrollCursor\` is already the oldest rendered day, so older signals page in cleanly when the user scrolls past the initial chunk.

## Net effect on a typical homepage load

- Initial \`/api/init\` payload drops from ~830 KB → a few dozen KB (typical 48h window).
- Brief + roster + leaderboard render faster (less JSON to parse).
- Deeper archive (>48h) loads on demand instead of being paid for up front.
- A user who bounces from the brief never pays for the older days at all.

## Test plan

- [x] \`signal-review.test.ts\` + \`do-client.test.ts\` pass (15 tests). They assert response shape, not row counts, so the 48h trim is invisible to them.
- [x] \`tsc --noEmit\` clean.
- [ ] Visual on preview deploy: homepage initial render shows last ~48h of signals; scrolling triggers \`/api/front-page?before=…\` and appends older days as before.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4835,6 +4835,25 @@
       }
 
       scrollState.loading = false;
+
+      // Sparse-window cascade: IntersectionObserver only fires on transition,
+      // so when the initial 48h window returns very few signals (or any single
+      // page doesn't push the sentinel out of viewport), the observer stays
+      // silent and the user has no scroll trigger to load more. Re-check the
+      // sentinel after each successful load and call again if it's still
+      // within the same 200px rootMargin we observe with. Bounded by
+      // hasMore + nextBefore advancing each iteration, so this terminates
+      // when the feed runs out or the new content finally pushes the
+      // sentinel below the viewport.
+      if (!scrollState.done && scrollState.sentinel) {
+        const rect = scrollState.sentinel.getBoundingClientRect();
+        const stillVisible = rect.top < (window.innerHeight + 200);
+        if (stillVisible) {
+          // Defer to a microtask so the just-inserted DOM gets a layout pass
+          // before the next iteration measures-and-decides.
+          Promise.resolve().then(loadNextDay);
+        }
+      }
     }
 
     function renderDaySection(dateStr, signals) {
--- a/src/objects/news-do.ts
+++ b/src/objects/news-do.ts
@@ -2086,9 +2086,12 @@ export class NewsDO extends DurableObject<Env> {
       return c.json({ ok: true, data: signals } satisfies DOResult<Signal[]>);
     });
 
-    // GET /signals/front-page — all approved + brief_included signals in a single query
-    // Eliminates the need for two separate /signals calls from the Worker route.
-    // LIMIT 500 preserves the old behavior (200 approved + 200 brief_included = up to 400).
+    // GET /signals/front-page — curated signals (approved + brief_included)
+    // from the last 48 hours. Older signals are loaded on-demand via the
+    // date-paginated /signals/front-page-page endpoint as the user scrolls.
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Trims the initial `/api/init` signals payload from ~830 KB (500 rows, ~12 days) to a ~48h window with a 200-row cap — a clean, well-scoped performance win that leans on the already-working infinite-scroll path for the deeper archive.

**What works well:**
- The change is applied consistently in both query sites (`/signals/front-page` route and the init bundle query) — no divergence risk.
- Comment quality is good. The before/after explanation in the comment blocks is exactly what a future maintainer needs to understand why the limit changed.
- PR description includes a live measurement (831 KB → a few dozen KB) — concrete numbers, not hand-waving.
- Correctly identified that `initInfiniteScroll()` already uses a `?before=<date>` cursor, so older signals aren't lost — just deferred.

**[question] Sparse 48h window edge case**

What does the homepage look like if the last 48h produced very few approved signals — say, 3–5 rows? On a desktop with a tall viewport, a user might see the initial render and have no natural scroll trigger to load older content. The infinite scroll is cursor-driven and fires on scroll; if the initial content doesn't fill the viewport, does `initInfiniteScroll()` auto-trigger a prefetch, or does the user have to scroll into whitespace to get more?

Not blocking if the team has already reasoned through this (or if sparse windows are rare enough that the UX tradeoff is acceptable), but worth confirming before merge.

**[suggestion] Duplicated query logic**

The same SQL pattern (`WHERE s.status IN ('approved', 'brief_included') AND s.created_at >= datetime('now', '-2 days') ... LIMIT 200`) now lives in two places. If the window or cap changes again, it'll need to be updated in both. A helper method on `NewsDO` — or at minimum a named constant for the interval and limit — would make future tweaks a one-line change.

Not blocking; the duplication is small and the current PR is already tight. Just flag it for next time this area is touched.

**[nit]** `'-2 days'` as a bare string literal in two places. A `const FRONT_PAGE_WINDOW = '-2 days'` at the top of the file would make the intent explicit and the value discoverable.

**Operational note:** We monitor the `/api/init` response as part of our signal-filing pipeline. The current ~830 KB payload has caused occasional timeout spikes in our sensors on slow dispatch cycles. This change directly improves that — smaller init payload means faster signal-pipeline polling for us as well.

Tests pass, `tsc` is clean, and the infinite-scroll fallback is already in production. Approving.

PR #589, opened 2026-04-22
fix(archive): render every loaded result so Load More actually adds visible rows

biwasxyz:
## Bug

User report: on the archive page, Load More fires successfully (subtitle "X of Y loaded" climbs, facet histogram updates) but the visible signal list never changes — feels like data loads but nothing appears in the UI.

## Root cause

\`renderResults()\` in \`public/archive/index.html\` hard-capped the visible list:

\`\`\`js
const rows = results.slice(0, 50).map(s => { ... })
\`\`\`

Load More appended the next 200-row page to \`state.allSignals\` — the in-memory pool grew, the subtitle "loaded N of M" climbed, the histogram across all loaded results updated — but the rendered listing stayed pinned at the same 50 *newest* rows because they were always the same after sorting newest-first. From the user's POV, every Load More click did nothing.

The 50-row cap predates the Load More feature; back then the message read "Showing first 50 of N. Narrow your filters to see more" because there was no other way to see more.

## Fix

Drop the \`.slice(0, 50)\` and render every loaded result. DOM cost of hundreds-to-low-thousands of simple \`<a>\` rows is fine, and Load More remains the user-controlled gate on how much grows here. Removes the now-misleading "Showing first 50 of N" note inside the listing — the subtitle ("X of Y loaded") + the Load More button already communicate the same idea cleaner.

## Test plan

- [ ] Visual on preview deploy: hit /archive, scroll the existing list, click Load More — new rows should appear at the bottom of the listing.
- [ ] Subtitle counter still tracks the loaded slice ("400 of 20,207 loaded" → "600 of 20,207 loaded" → …).
- [ ] Filter facets still update against all loaded results.

### DIFF
--- a/public/archive/index.html
+++ b/public/archive/index.html
@@ -792,7 +792,13 @@
       // Sort newest first
       results.sort((a, b) => new Date(b.timestamp) - new Date(a.timestamp));
 
-      const rows = results.slice(0, 50).map(s => {
+      // Render every loaded result. Slicing to a fixed count here meant Load
+      // More appended pages to state.allSignals (subtitle counter ticked up)
+      // but the visible list stayed pinned at the same 50 newest rows — the
+      // user saw counts climb with no new rows to show for it. The DOM cost
+      // of rendering ~hundreds-to-low-thousands of simple <a> rows is fine;
+      // Load More remains the user-controlled gate on how much grows here.
+      const rows = results.map(s => {
         const slug = s.beatSlug || (s.beat || '').toLowerCase().replace(/\s+/g, '-');
         const color = BEAT_COLORS[slug] || '#1a1a1a';
         const st = (s.status || '').toLowerCase();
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fixes a broken Load More — good catch.

**What the PR does:** Drops the legacy `.slice(0, 50)` cap in `renderResults()` that predated the Load More feature. The cap meant the visible list was always pinned to the same 50 newest rows no matter how many pages had been fetched, so Load More appeared to do nothing even as the subtitle counter and histogram correctly reflected the growing pool.

**What looks good:**
- Minimal, targeted change — one line removed, no collateral edits
- Removes the now-misleading "Showing first 50 of N" message inside the listing (the subtitle + Load More button communicate this better)
- In-code comment clearly explains the fix and the reasoning for no slice cap

**Functionality:** Correct. Filters, histogram, and subtitle counter are unchanged. Load More remains the user-controlled gate on how much is loaded — the only thing that changes is that everything loaded now renders.

**Performance:** At ~200 rows per Load More page, even 10 pages yields ~2000 simple `<a>` rows. That's well within what modern browsers handle comfortably, and Load More prevents unbounded growth. The PR description's reasoning here is sound.

**Security:** No new injection surface — `esc()` still applied to all interpolated values.

Good fix. Clean PR.

PR #591, opened 2026-04-22
fix(beats): accurate Signals 7d, relabel 'In Brief' → 'Approved', '—' placeholders

biwasxyz:
## Bugs

User report on the Bureau cards (\`public/beats/index.html\`). Verified against live \`/api/signals/counts\`:

| Beat | "Signals · 7d" shown | Actual 7d total | "In Brief · Today" shown | Actually approved+in-brief today |
|---|---|---|---|---|
| AIBTC Network | **50** | **1,716** | 0 | 0 |
| Bitcoin Macro | **113** | **2,521** | 10 | 10 |
| Quantum | **37** | **1,100** | 1 | 1 |

### Bug 1 — "Signals · 7d" was 30-40× undercounted

The page fetched \`/api/signals?since=7d&limit=500\` and partitioned client-side, but \`/api/signals\` is hard-capped at 200 rows server-side. The 200 newest signals only spanned ~5.7 hours of data dominated by the highest-volume beat (bitcoin-macro), so per-beat 7d totals were tiny fractions of reality.

### Bug 2 — "In Brief · Today" label was wrong

The number was \`approved + brief_included + inscribed\`, but \`approved\` signals haven't been published in a brief yet — they're awaiting compile. The label implied editorial finality the number doesn't carry. (Today's brief had \`compiledAt: null\` when the user noticed this, so the "in brief" labels for live data were structurally suspect.)

## Fix

Both stats now read from \`/api/signals/counts\` (purpose-built endpoint, not row-capped):
- "Signals · 7d" ← \`/api/signals/counts?beat=&since=7d-utc-midnight\`
- "Approved · Today" ← \`approved + brief_included\` over \`since=utc-midnight\`

## Drive-by — addressing review-style nit on hard-coding 0

Cards render immediately with **"—" (em dash) placeholders** for the count cells, not \`0\`. Hard-coding \`0\` initially implied "no signals" when the truth was "haven't loaded yet". Each beat's counts patch into its already-rendered card the moment its responses arrive via \`fillBeatCardCounts()\`, so the page doesn't wait on the slowest beat. Fetch failures leave the "—" in place rather than misleading with a partial 0.

## Test plan

- [ ] Visual on preview deploy: Bureau cards initially show real Agents number + "—" for the other two stats, then "—" swaps for the real number per beat as its counts arrive.
- [ ] "Signals · 7d" now shows 4-digit numbers for active beats (matches \`curl /api/signals/counts?beat=…&since=…\`).
- [ ] "Approved · Today" matches the homepage rail's "approved" tile number for the same beat.

### DIFF
--- a/public/beats/index.html
+++ b/public/beats/index.html
@@ -439,7 +439,7 @@
       return svg;
     }
 
-    function renderBeatCards(beats, signalsByBeat, correspondents) {
+    function renderBeatCards(beats, countsBySlug, correspondents) {
       const container = document.getElementById('beat-cards');
       if (!beats.length) {
         container.innerHTML = '<div class="empty">No beats found.</div>';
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fixes two real bugs on the Bureau cards — the undercounting root cause (API row cap silently truncating results) and the mislabeling of editorial state. Both are correct.

**What works well:**
- Switching to `/api/signals/counts` is the right call. We confirmed operationally (issue #566 / task #13233) that `/api/signals` is capped at 200 rows server-side, which dominated the window with bitcoin-macro traffic and left other beats with single-digit 7d counts. The purpose-built counts endpoint eliminates this entirely.
- `fillBeatCardCounts()` is clean progressive patching. Adding `data-beat-slug` to the card anchor and `data-stat` to the number elements gives a precise, re-render-free update path. Good choice over a full container repaint.
- `CSS.escape(slug)` in the querySelector attribute selector is the right API here — avoids CSS injection if a slug ever contains special characters.
- The `typeof sinceUtcMidnightIso === 'function'` guard matches the existing `sinceDaysAgoIso` pattern in the file — consistent defensive coding.
- Comment quality is high. The inline explanation of why `approved` can't carry the "In Brief" label is exactly the kind of context that prevents a future author from "fixing" it back.

**[question] `inscribed` status dropped from approvedToday without explanation** (`public/beats/index.html` — forEach callback)

The old code counted `approved || brief_included || inscribed`; the new code counts `approved + brief_included`. The commit message explains why `approved` is included (awaiting-compile state), but doesn't address removing `inscribed`. Is that intentional?

If `inscribed` means on-chain finalization (after brief compile), it seems like it should also contribute to "Approved · Today" — signals that made it all the way through are still editorially accepted. Leaving it out understates the number slightly. If there's a reason to exclude it (e.g., inscribed signals have a different timestamp path that makes the `since` filter unreliable), worth a one-line comment.

**Operational note:** We file signals to `aibtc-network`, `bitcoin-macro`, and `quantum` daily and query `/api/signals/counts` to check our cap usage. The real 7d totals in the PR description (1,716 / 2,521 / 1,100) match what we see operationally — the fix is directionally confirmed from our end.

biwasxyz:
Changes already landed in main via #592's merge commit (`bbe63f0`). Closing this PR to reflect that.

PR #593, opened 2026-04-22
perf(api): edge-cache correspondents, beats, classifieds, front-page (same pattern as #592)

biwasxyz:
## Why

PR #592 wrapped \`/api/init\` in the Workers Cache API and dropped its TTFB from ~3s to <100ms after first hit. From the audit immediately after that PR landed, four more endpoints have the same problem:

\`\`\`
$ for url in /api/correspondents /api/classifieds /api/beats /api/front-page; do
    curl -s -o /dev/null -w "$url ttfb=%{time_starttransfer}s size=%{size_download} edge=%header{x-edge-cache}\n" "https://aibtc.news$url"
  done
/api/correspondents ttfb=2.97s size=371106 edge=
/api/classifieds    ttfb=5.80s size=2630   edge=
/api/beats          ttfb=1.32s size=5433   edge=
/api/front-page     ttfb=0.57s size=91945  edge=
\`\`\`

Same root cause as #592: each route sets a \`Cache-Control: public, s-maxage=300\` header but Workers responses bypass the edge cache unless explicitly stored via \`caches.default\`.

## What

Extract the inline \`match → put\` pattern from #592 into \`src/lib/edge-cache.ts\` as two helpers:

\`\`\`ts
export async function edgeCacheMatch(c): Promise<Response | null>
export function edgeCachePut(c, response): void
\`\`\`

Apply to five routes (one is \`/api/init\` refactored to use the helper instead of inline code):

| Route | Before TTFB | Cached TTL | After (hit) |
|---|---|---|---|
| \`/api/init\` | 0.29s (already cached) | 300s | <100ms (no change, just refactored) |
| \`/api/correspondents\` | 2.97s | 300s | <100ms |
| \`/api/classifieds\` | 5.80s | 300s | <100ms |
| \`/api/beats\` | 1.32s | 300s | <100ms |
| \`/api/front-page\` | 0.57s | 300s | <100ms |

Cache key is the canonical request URL — \`?status=active\`, \`?include=members\`, \`?before=2026-04-22\`, etc. each get independent cache entries, so query-string variants don't collide. \`X-Edge-Cache: HIT|MISS\` response header is set on every response so cache state is observable in DevTools.

## Side notes

- **\`/api/classifieds\` 5.8s TTFB for 2.6 KB output is anomalously slow.** That's the underlying DO query needing investigation — separate fix. The cache layer just gets us out of paying that cost on every visit in the meantime.
- Cache invalidation is purely TTL-based (5 min). New classifieds/beats/correspondent updates may show up to 5 minutes of staleness. Acceptable given how rarely those things change.
- 37 tests across the touched routes all pass locally.
- The same pattern is reusable for any other GET endpoint where staleness is acceptable.

## Test plan

- [ ] Visual on preview deploy: load each affected URL twice within 5 minutes; second request shows \`X-Edge-Cache: HIT\` and TTFB <100ms.
- [ ] After 5+ minutes, cache refreshes (\`X-Edge-Cache: MISS\` again on first hit).
- [ ] Beats page renders the same as before (it now hits cached \`/api/beats\` + \`/api/correspondents\` + \`/api/signals/counts\`).
- [ ] Homepage's classifieds rotation still works (unaffected — uses \`/api/classifieds/rotation\`, which has \`Cache-Control: no-store\` and is intentionally uncached).

### DIFF
--- a/src/lib/edge-cache.ts
+++ b/src/lib/edge-cache.ts
@@ -1,0 +1,64 @@
+/**
+ * Workers edge-cache helper.
+ *
+ * Workers responses don't automatically populate Cloudflare's edge cache —
+ * the `Cache-Control` header alone is treated as "instructions for downstream
+ * caches" but no downstream cache is ever asked unless the response is
+ * explicitly stored via `caches.default.put()`. This helper wraps the
+ * match → put pattern so route handlers can opt in with a single call.
+ *
+ * Pattern:
+ *
+ *   router.get("/api/foo", async (c) => {
+ *     const cached = await edgeCacheMatch(c);
+ *     if (cached) return cached;
+ *
+ *     // ... build response ...
+ *     c.header("Cache-Control", "public, max-age=60, s-maxage=300");
+ *     const response = c.json(payload);
+ *     edgeCachePut(c, response);
+ *     return response;
+ *   });
+ *
+ * Cache key is the canonical request URL (so `?before=2026-04-22` and the
+ * naked path get separate entries). TTL is taken from the response's
+ * `Cache-Control` `s-maxage` directive at edge level. Browser revalidation
+ * still honours `max-age` independently.
+ */
+import type { Context } from "hono";
+import type { Env, AppVariables } from "./types";
+
+type Ctx = Context<{ Bindings: Env; Variables: AppVariables }>;
+
+function buildCacheKey(c: Ctx): Request {
+  return new Request(new URL(c.req.url).toString(), { method: "GET" });
+}
+
+/**
+ * Look up the current request in the edge cache. Returns the cached Response
+ * (with an `X-Edge-Cache: HIT` header attached for observability) or `null`
+ * on miss. Safe to call from any GET handler.
+ */
+export async function edgeCacheMatch(c: Ctx): Promise<Response | null> {
+  const cached = await caches.default.match(buildCacheKey(c));
+  if (!cached) return null;
+  // Clone-via-Response constructor so we can mutate the headers without
+  // touching the body stream (which would break subsequent reads).
+  const hit = new Response(cached.body, cached);
+  hit.headers.set("X-Edge-Cache", "HIT");
+  return hit;
+}
+
+/**
+ * Store the response in the edge cache for the duration of its
+ * `Cache-Control` `s-maxage`. Tags an `X-Edge-Cache: MISS` header on the
+ * response we return to the client (so the caller can confirm the write
+ * happened). The actual cache write runs via `executionCtx.waitUntil` so
+ * the user doesn't pay any latency for it.
+ */
+export function edgeCachePut(c: Ctx, response: Response): void {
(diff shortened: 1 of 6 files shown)


arc0btc (approved):
Extracts the inline `match → put` pattern from #592's `/api/init` into a reusable `src/lib/edge-cache.ts` helper and applies it to four more slow endpoints. Clean, consistent, and a measurable win — the approach is correct.

**What works well:**
- `edgeCacheMatch` / `edgeCachePut` are minimal, well-documented, and the naming makes call-sites self-evident
- `waitUntil` for the cache write is the right pattern — first caller pays the DO round-trip, everyone else gets the edge hit
- `response.clone()` before `put()` is required (streams are single-read); good that it's handled inside the helper so callers can't forget it
- `/api/classifieds/rotation` (intentionally `Cache-Control: no-store`) is correctly left untouched
- Both branches in `/api/front-page` (paginated `?before` and default mode) independently call `edgeCachePut` — no missed path
- `/api/init` refactor to use the helper reduces the route to the same three-liner as all the others; the old inline code was 15 lines doing the same job

**[suggestion] Stored response carries `X-Edge-Cache: MISS` header at rest** (`src/lib/edge-cache.ts:57-60`)

`edgeCachePut` sets `X-Edge-Cache: MISS` on the response *before* calling `.clone()`, so the cloned copy stored in the cache also has that header. That's harmless today because `edgeCacheMatch` overwrites it to `HIT` on every read, but it means introspecting the raw cache entry (e.g. via the Cache API dashboard or a test that reads the stored response directly) gives a misleading header value. Consider setting the header *after* cloning:

```suggestion
export function edgeCachePut(c: Ctx, response: Response): void {
  const clone = response.clone();
  response.headers.set(X-Edge-Cache, MISS);
  c.executionCtx.waitUntil(caches.default.put(buildCacheKey(c), clone));
}
```

**[question] `/api/classifieds` caches agent-specific views** (`src/routes/classifieds.ts`)

`?agent=ADDRESS` returns all submissions for that agent, including potentially non-approved ones. Cache key is the full URL so only requests with that exact URL hit the cache entry — no cross-agent leakage. Just confirming this is the intended behaviour: an agent's pending/rejected classifieds would be cached for 5 min, visible to anyone who constructs the same URL.

**[nit]** The `Ctx` type alias is defined inside `edge-cache.ts` and duplicates the same generic structure as the routers. If `types.ts` already exports a composed context type, importing it here would remove the duplication. Low priority since the coupling is intentional (keeps the helper self-contained).

**Operational context:**
We're running the classifieds, beats, and correspondents endpoints in production and can confirm the slowness is real — `/api/classifieds` at 5.8s for 2.6 KB is the most anomalous. The PR note about investigating the underlying DO query is a good follow-up item; the cache is the right immediate fix regardless. The 5-minute staleness window for `/api/front-page` is acceptable given that new brief compilations happen at the 23:30 UTC lock, not on a rolling basis.

37 tests passing locally + cache-observability via `X-Edge-Cache` header makes this safe to ship.

PR #596, opened 2026-04-22
feat(seo): add robots.txt, sitemap family, and canonical SEO meta (ph…

biwasxyz:
Brings the SEO foundation to an indexable state:

- New /robots.txt (prod: allow + sitemaps + Googlebot-News/Image rules; staging/dev: disallow all + X-Robots-Tag: noindex)
- New /sitemap.xml index + /sitemap/pages.xml + /sitemap/signals.xml
- New /news-sitemap.xml (Google News spec, last 48h, ≤1000 URLs)
- Added canonical, og:site_name, og:locale, og:image:alt, theme-color, twitter:title/description, and a proper robots meta to every public shell
- Operator tools (/file/, /onboard/) now explicitly noindex,nofollow
- Signal pages gain canonical + article:published_time/modified_time/section and dynamic robots (curated = index, drafts/rejects = noindex) — the "[SUBMITTED]" description prefix that poisoned social cards is gone
- 6 smoke tests for the SEO router

Phase 2 will replace the JS redirect on /signals/:id with full server rendering + NewsArticle JSON-LD including Bitcoin inscription provenance.

### DIFF
--- a/public/about/index.html
+++ b/public/about/index.html
@@ -5,17 +5,25 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <script defer src="https://cloud.umami.is/script.js" data-website-id="3ed4837c-81d1-4d12-b657-158cb5881e89"></script>
   <title>About — AIBTC News</title>
+  <link rel="canonical" href="https://aibtc.news/about/">
   <link rel="icon" href="data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100'><text y='.9em' font-size='90'>🗞️</text></svg>">
   <meta name="description" content="How AIBTC News works — a decentralized intelligence network where AI agents file signals, compile briefs, and inscribe them permanently to Bitcoin.">
+  <meta name="robots" content="index,follow,max-image-preview:large,max-snippet:-1,max-video-preview:-1">
+  <meta name="theme-color" content="#af1e2d">
+  <meta property="og:site_name" content="AIBTC News">
+  <meta property="og:locale" content="en_US">
   <meta property="og:title" content="About — AIBTC News">
   <meta property="og:description" content="A decentralized intelligence network where AI agents file signals, compile briefs, and inscribe them permanently to Bitcoin.">
   <meta property="og:image" content="https://aibtc.news/og-image.png">
   <meta property="og:image:width" content="1200">
   <meta property="og:image:height" content="630">
   <meta property="og:image:type" content="image/png">
+  <meta property="og:image:alt" content="AIBTC News — agent-written news inscribed on Bitcoin">
   <meta property="og:type" content="website">
   <meta property="og:url" content="https://aibtc.news/about/">
   <meta name="twitter:card" content="summary_large_image">
+  <meta name="twitter:title" content="About — AIBTC News">
+  <meta name="twitter:description" content="A decentralized intelligence network where AI agents file signals, compile briefs, and inscribe them permanently to Bitcoin.">
   <meta name="twitter:image" content="https://aibtc.news/og-image.png">
 
   <link rel="preconnect" href="https://fonts.googleapis.com">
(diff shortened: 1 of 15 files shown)


arc0btc (approved):
Adds the SEO foundation to make aibtc.news properly indexable: robots.txt with environment-aware crawler control, a full sitemap family (index → pages + signals), a Google News sitemap, and consistent canonical/OG/Twitter meta across all public shells. The removal of the \`[SUBMITTED]\` prefix from signal page OG descriptions is a real fix — that prefix was poisoning social card previews.

**What works well:**
- `isProduction(c)` / `xmlResponse()` helpers are clean — the pattern is reused correctly throughout and keeps handler bodies readable
- `escXml()` covers all five XML entity substitutions (`&`, `<`, `>`, `"`, `'`) — correct for user-supplied headline content
- Environment-aware robots.txt (disallow everything in staging + `X-Robots-Tag: noindex`) is the right approach for a Cloudflare Worker deployment
- Dynamic robots directive on signal pages (index curated signals, noindex drafts/rejects) is exactly right — only approved/brief_included signals should burn Google's crawl budget
- 6 smoke tests cover the structural requirements without being brittle — checking namespace declarations and structural markers rather than exact content
- Mounting seoRouter before static assets is critical; the comment explaining why is appreciated

**[suggestion] `SITE_URL` is hardcoded — staging sitemaps emit production URLs** (`src/routes/seo.ts:493`)

robots.txt correctly locks out crawlers in staging, but the XML bodies in `/sitemap/signals.xml`, `/news-sitemap.xml`, and the pages sitemap all contain `https://aibtc.news` URLs regardless of environment. A staging crawler (or a developer testing locally) will see sitemap entries pointing at production. Low risk in practice since X-Robots-Tag blocks indexing, but it makes local testing confusing. Consider deriving base URL from an env var:

```suggestion
const SITE_URL = c.env.BASE_URL ?? "https://aibtc.news";
```

(Or keep it hardcoded with a comment noting it's intentional — staging is robots-locked anyway.)

**[suggestion] Two independent `listFrontPage` calls for signals.xml + news-sitemap.xml** (`src/routes/seo.ts:637,669`)

Each handler independently calls `fetchCuratedSignals(c)` → `listFrontPage(c.env)`. A single crawler fetching both sitemaps triggers two DO round-trips. The 600s/300s `Cache-Control` headers mean Cloudflare CDN mitigates this in production, but under a cold cache (first crawl, redeploy) both fire back-to-back. Not blocking — the cache strategy handles it — but worth knowing if the DO starts seeing burst traffic from crawler indexing runs.

**[nit]** The inline ternary chain for `article:` meta in `signal-page.ts:765` is dense on a single line — three ternaries embedded in a template literal. Could be extracted to a string variable built before the template:

```ts
const articleMeta = [
  publishedTime && `<meta property="article:published_time" content="${publishedTime}">`,
  modifiedTime && `<meta property="article:modified_time" content="${modifiedTime}">`,
  beat && `<meta property="article:section" content="${beat}">`,
].filter(Boolean).join("\n  ");
```

Then inline `${articleMeta ? "\n  " + articleMeta : ""}` in the template. Take it or leave it — the current version is functional, just harder to read.

**[question]** The sitemap index `lastmod` is always `new Date().toISOString()` — the current moment at request time (`src/routes/seo.ts:570`). This means every cache miss generates a fresh lastmod, which could signal to crawlers that the index is perpetually updated and encourage more frequent re-crawls of children. Is that intentional? If the 3600s cache handles it, fine. If not, a static lastmod (e.g., derived from the most recent signal's `updated_at`) would be more accurate.

**Code quality notes:**
- `fetchCuratedSignals` is a clean isolation layer — wrapping `listFrontPage` in try/catch with `[]` fallback means a DO outage degrades gracefully to an empty sitemap rather than a 500. Good pattern.
- The `STATIC_PAGES` array at module level (not per-request) avoids unnecessary allocations on every hit — correct.
- `&mdash;` → `—` replacements in HTML files are cosmetically fine; UTF-8 pages render identically and the encoded form was arguably unnecessary.

**Operational note:** Signal pages are what we link from our filed signals. The `[SUBMITTED]` prefix removal directly improves how our signals render in Slack/X preview cards. Phase 2 (full server rendering + NewsArticle JSON-LD with inscription provenance) will make individual signal URLs first-class indexable content — the Phase 1 foundation here is the right prerequisite.

Approved.

PR #597, opened 2026-04-22
feat(seo): server-render /signals/:id + NewsArticle JSON-LD (phase 2a)

biwasxyz:
## Summary

Replaces the JS-redirect trick on \`/signals/:id\` with a complete, server-rendered article page. Every signal URL now resolves to real, indexable content for Google, social crawlers, and anyone who shares or deep-links a signal. The homepage modal UX is **untouched** — in-session clicks still use \`pushState\` + the modal.

Phase 1 (#596) gave us canonical URLs, sitemaps, and robots.txt. Phase 2A is where the actual content goes server-side.

## What ships

- **Full article HTML** — breadcrumbs, headline, byline (author link + "AI agent" badge), beat, UTC timestamp, status, body, tags, sources, provenance block. Styled with \`shared.css\` design tokens.
- **\`NewsArticle\` JSON-LD** with \`digitalSourceType=trainedAlgorithmicMedia\` (IPTC AI-disclosure vocabulary), \`creditText\`, \`keywords\`, \`citation\`, \`author\` as \`Person\`, \`publisher\` as \`Organization\` \`@id\` reference.
- **\`BreadcrumbList\`** JSON-LD (Home > Signals > Article).
- **\`Organization\`** JSON-LD with \`ImageObject\` logo + \`@id\`.
- **Bitcoin provenance** — when the signal is in an inscribed brief, adds \`identifier[]\` \`PropertyValue\` entries for \`BitcoinInscriptionId\` / \`BriefDate\` / \`BitcoinTxId\`, plus \`sameAs\` and \`archivedAt\` pointing to ordinals.com. Visible "Provenance" section with links to the ordinals viewer and mempool.space.
- **Status-aware robots** — approved / brief_included → \`index\`; everything else → \`noindex,nofollow\` (meta + \`X-Robots-Tag\`).
- **Proper 404 page** with \`X-Robots-Tag: noindex\` instead of a 302 back to the homepage.
- **\`article:published_time\` / \`modified_time\` / \`section\` / \`author\` / \`tag\`** meta for open-graph.

## What doesn't change

- Homepage modal UX — \`public/shared.js\` and \`public/index.html\` are untouched. Clicks from \`/\` still \`pushState\` to \`/signals/:id\` and open the modal; that URL just now also resolves to a real page on refresh / share / back.
- Classifieds x402 payment flow at \`POST /api/classifieds\` — totally separate route, no intersection.
- \`/wire/\`, \`/archive/\`, \`/agents/\` etc. — already linked \`/signals/:id\` via \`<a href>\`, so clicks from those pages were already doing a real navigation. They now land on a real article instead of the old redirect dance.

## Commits (one per file for review)

- \`a0fc176\` — new \`src/lib/signal-provenance.ts\` helper
- \`62e441f\` — \`src/routes/signal-page.ts\` rewrite
- \`a213d7f\` — \`src/__tests__/signal-page.test.ts\`

## Test plan

- [x] Typecheck + lint clean on all three files
- [x] 10 new integration tests for the article page (status-aware robots, JSON-LD shape, canonical URL, 404 HTML, disclosure visibility)
- [x] Full suite: 306 pass / 4 fail — same 4 pre-existing failures on \`main\` (\`identity-gate\`, \`scoring-math\`); zero new failures
- [x] Classifieds + signals + corrections + seo tests pass (manually verified no cross-contamination with x402 flow)
- [ ] Preview deploy renders an article page with visible Provenance / Sources / Tags
- [ ] Google Rich Results Test on a seeded signal URL shows valid \`NewsArticle\` structured data
- [ ] Verify noindex header on a draft signal and 404 on a missing ID

## Next (phase 2B)

Server-render the homepage via \`run_worker_first: ["/"]\` + \`HTMLRewriter\` injection of today's brief + top signals into the existing shell. Keeps the modal UX, delivers real content in the initial HTML response, and nets the biggest LCP win we have left.

### DIFF
--- a/src/__tests__/signal-page.test.ts
+++ b/src/__tests__/signal-page.test.ts
@@ -10,10 +10,12 @@ import { SELF } from "cloudflare:test";
 
 const APPROVED_ID = "sigpage-approved-001";
 const DRAFT_ID = "sigpage-submitted-002";
+const HOSTILE_SOURCE_ID = "sigpage-hostile-003";
 const MISSING_ID = "00000000-0000-0000-0000-000000000000";
 
 const APPROVED_ADDR = "bc1qsigpageapproved0000000000000000000000000";
 const DRAFT_ADDR = "bc1qsigpagedraft00000000000000000000000000000";
+const HOSTILE_ADDR = "bc1qsigpagehostile00000000000000000000000000";
 
 beforeAll(async () => {
   await SELF.fetch("http://example.com/api/test-seed", {
@@ -44,6 +46,19 @@ beforeAll(async () => {
           sources: "[]",
           created_at: "2026-04-20T11:00:00Z",
           status: "submitted",
+          disclosure: "",
+        },
+        {
+          id: HOSTILE_SOURCE_ID,
+          beat_slug: "bitcoin-macro",
+          btc_address: HOSTILE_ADDR,
+          headline: "Signal with a javascript: source URL",
+          body: "Source-URL guard coverage.",
+          sources: JSON.stringify([
+            { url: "javascript:alert(1)", title: "Evil link" },
+          ]),
+          created_at: "2026-04-20T12:00:00Z",
+          status: "approved",
           disclosure: "",
         },
       ],
@@ -138,6 +153,25 @@ describe("GET /signals/:id — draft (submitted)", () => {
     const body = await res.text();
     expect(body).toMatch(/<meta name="robots" content="noindex,nofollow">/);
     expect(res.headers.get("x-robots-tag")).toBe("noindex");
+  });
+});
+
+// ---------------------------------------------------------------------------
+// Source URL hardening — reject non-http(s) hrefs at render time
+// ---------------------------------------------------------------------------
+
+describe("GET /signals/:id — hostile source URL", () => {
+  it("strips javascript: URLs from source hrefs", async () => {
+    const res = await SELF.fetch(
+      `http://example.com/signals/${HOSTILE_SOURCE_ID}`
+    );
+    expect(res.status).toBe(200);
+    const body = await res.text();
+    // The source link must not carry the hostile protocol into the DOM.
+    expect(body).not.toContain("javascript:alert(1)");
+    expect(body).not.toMatch(/href="javascript:/i);
+    // The title is still rendered so the source row remains visible.
+    expect(body).toContain("Evil link");
   });
 });
 
--- a/src/__tests__/signal-page.test.ts
+++ b/src/__tests__/signal-page.test.ts
@@ -1,0 +1,160 @@
+import { describe, it, expect, beforeAll } from "vitest";
+import { SELF } from "cloudflare:test";
+
+/**
+ * Integration tests for the full-page server-rendered article at
+ * GET /signals/:id. Verifies the HTML shape, structured data, canonical
+ * URL, and status-based robots directives. These are the surfaces Google
+ * and social crawlers actually consume.
+ */
+
+const APPROVED_ID = "sigpage-approved-001";
+const DRAFT_ID = "sigpage-submitted-002";
+const MISSING_ID = "00000000-0000-0000-0000-000000000000";
+
+const APPROVED_ADDR = "bc1qsigpageapproved0000000000000000000000000";
+const DRAFT_ADDR = "bc1qsigpagedraft00000000000000000000000000000";
+
+beforeAll(async () => {
+  await SELF.fetch("http://example.com/api/test-seed", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({
+      signals: [
+        {
+          id: APPROVED_ID,
+          beat_slug: "bitcoin-macro",
+          btc_address: APPROVED_ADDR,
+          headline: "Signal page rendering test — approved story",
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Server-renders the \`/signals/:id\` route with a full \`NewsArticle\` JSON-LD article, replacing the old JS-redirect approach — a meaningful SEO upgrade and the right way to expose signal content to crawlers and deep-link shares.

**What works well:**
- \`escJsonLd()\` is a thoughtful addition — escaping \`<\` as \`\u003c\` is exactly the right defense against accidental \`</script>\` injection in JSON-LD blocks
- Status-aware robots with *both* \`<meta name="robots">\` and \`X-Robots-Tag\` header is belt-and-suspenders, and correct
- The provenance early-return (\`if (signal.status !== "brief_included") return null\`) avoids unnecessary DO round-trips on the common approved/submitted paths — good performance instinct
- \`rel="nofollow noopener"\` on external source links
- The Bitcoin inscription provenance section is the architecturally differentiated content here — linking signal articles to immutable on-chain records with ordinals + mempool.space links is exactly the kind of thing that should be surfaced at the article level
- Test suite covers the key crawler-visible surfaces (robots directives, JSON-LD shape, canonical, 404 behavior) — solid integration coverage for the SEO contract

**[suggestion] Protocol check on source URLs** (\`src/routes/signal-page.ts\`, \`renderSources\`)

\`renderSources\` builds \`href\` from \`s.url\` with only \`esc()\` applied. \`esc()\` handles HTML entities but won't neutralize a \`javascript:\` URL. Agent-filed signals *should* have their sources validated upstream, but at the rendering layer a minimal protocol guard costs almost nothing:

\`\`\`suggestion
    const safeUrl = /^https?:\/\//i.test(s.url || "") ? s.url : "#";
    const href = esc(safeUrl || "#");
\`\`\`

Not blocking — the editorial pipeline likely validates this — but defense in depth at the render layer is cheap.

**[suggestion] Inline CSS block** (\`src/routes/signal-page.ts\`, \`ARTICLE_STYLES\`)

The \`ARTICLE_STYLES\` constant is ~160 lines of CSS inlined into every page response. Moving it to \`public/signal-page.css\` and adding \`<link rel="stylesheet" href="/signal-page.css">\` would let browsers cache it across signal page navigations. Minor win but consistent with how \`shared.css\` is handled.

**[question] \`signal.updated_at\` in the Signal type**

The code uses \`signal.updated_at || signal.created_at\` in two places (\`buildNewsArticle\` + \`renderArticleHTML\`). The test seed for the draft signal doesn't set \`updated_at\`. If \`updated_at\` is optional on the \`Signal\` type, TypeScript strict mode should catch it — the test plan says typecheck is clean, so presumably it's declared. Just confirming this is covered rather than silently falling through.

**[nit]** \`beatLink\` href is always \`/beats/\` regardless of beat slug. Probably intentional for Phase 2 but worth a TODO comment so it doesn't get lost.

**Code quality notes:**
- \`buildNewsArticle\`, \`renderArticleHTML\`, and the route handler all independently compute \`publishedIso\` / \`modifiedIso\` from \`signal.created_at\` / \`signal.updated_at\`. Could be extracted once — [suggestion] level, not blocking.
- \`interface Jsonish { [k: string]: unknown; }\` is fine locally. The key-deletion loop at the bottom of \`buildNewsArticle\` works correctly for its purpose.

**Operational note:** We file signals to the \`aibtc-network\`, \`bitcoin-macro\`, and \`quantum\` beats. Our signals will render through this article page when shared or linked. The provenance block is particularly relevant for \`brief_included\` signals — we have active inscribed briefs, so this will surface real Bitcoin provenance data for our filed signals. Tested path looks correct.

biwasxyz:
Thanks for the careful review — walking through each point with my reasoning.

**Applied** — protocol check on source URLs (commit `9d22d6d`).

You're right that this is worth fixing at the render layer. I traced `validateSources` in `src/lib/validators.ts:33` and confirmed it checks length but not protocol, so a `javascript:` URL passes validation and reaches the renderer. `esc()` doesn't neutralize it either — the href would still execute on click.

While writing the regression test I found the raw URL was also leaking into the `NewsArticle.citation` JSON-LD block (`buildNewsArticle`), so the fix is applied in both places via a shared `safeHttpUrl()` helper: bad URLs render as `#` in the visible DOM, and citation entries with bad URLs are silently dropped from structured data so we don't pollute schema.org output.

New test `strips javascript: URLs from source hrefs` covers both paths.

**Skipped with reasoning:**

- *Inline CSS extraction* — edge caches the HTML for 300s, so the 4KB CSS ships cached with the HTML. Extracting adds an asset file + cache-busting surface for a marginal win. Noting as potential future work in Phase 5 polish.
- *`updated_at` on Signal type* — confirmed `readonly updated_at: string` at `types.ts:198`, so it's required. The `|| created_at` fallback is defense for the empty-string case (falsy), not for `undefined`.
- *TODO on `beatLink`* — confirmed no `/beats/:slug` static page or Hono route exists today, so `/beats/` is the correct link for now. When per-beat pages ship, that PR will update this href naturally.
- *Extract `publishedIso` / `modifiedIso` helper* — only 2 call sites; keeping the duplication local per the repo's "prefer duplication over premature abstraction" convention.

Test suite: 307/311 (same 4 pre-existing failures on `main` — `identity-gate`, `scoring-math`).

PR #598, opened 2026-04-22
fix(signal-page): honest provenance copy for brief_included + pending

biwasxyz:
## Summary

Found during Phase 2A prod verification.

Signals with status \`brief_included\` whose brief **hasn't been inscribed on Bitcoin yet** (e.g. \`63d3eaf9-8243-4fc6-b7a9-050ef374f084\`, \`99061dcf-1741-4f12-9d8d-7b2986dfce0e\`) currently render:

> "This signal is currently in editorial review and has not yet been inscribed on Bitcoin."

That's wrong on both counts — \`brief_included\` is **past** editorial review, and we do know the brief date even before the inscription lands.

## Fix

Splits \`SignalProvenance\` into a discriminated union so the renderer and the JSON-LD builder can both make stage-accurate claims:

- \`"inscribed"\` — brief has \`inscription_id\` (+ optional txid). Full provenance block + \`sameAs\` + \`archivedAt\` + \`BitcoinInscriptionId\` / \`BitcoinTxId\` / \`BriefDate\` in \`identifier[]\`.
- \`"brief-pending"\` — brief exists, not inscribed yet. Renders *"Included in the YYYY-MM-DD daily brief. Awaiting Bitcoin inscription."* \`BriefDate\` emitted in JSON-LD; no inscription identifiers, no \`sameAs\`, no \`archivedAt\`.
- \`null\` — signal isn't \`brief_included\`. Falls back to status-based copy (\`approved\` → "awaiting next brief"; other → "in editorial review").

Important side effect: \`creditText\` now only claims *"inscribed on Bitcoin"* when the record is actually on-chain. We no longer overstate provenance in structured data.

## Production evidence

Verified these currently render the wrong "editorial review" copy on \`https://aibtc.news\`:
- \`/signals/63d3eaf9-8243-4fc6-b7a9-050ef374f084\` (aibtc-network beat)
- \`/signals/99061dcf-1741-4f12-9d8d-7b2986dfce0e\` (bitcoin-macro beat)
- \`/signals/e0598ecd-58c9-46ed-acd9-38ea0d59fb40\` (bitcoin-macro beat)

Inscribed-brief signals (e.g. \`7d995511-...\` from the 2026-03-29 inscription) already render correctly — that path isn't changed.

## Test plan

- [x] Typecheck clean (discriminated union forces all callers to narrow).
- [x] 2 new tests in \`signal-page.test.ts\` cover the brief-pending path:
  - Visible copy contains "Included in the 2026-04-19 daily brief" and "Awaiting Bitcoin inscription"; does NOT contain "currently in editorial review".
  - JSON-LD \`identifier[]\` contains \`BriefDate\` but NOT \`BitcoinInscriptionId\`.
  - No \`archivedAt\` / \`sameAs\` leaks for pending briefs.
- [x] Full suite: 309 pass / 4 fail (same pre-existing failures: \`identity-gate\`, \`scoring-math\`). Zero new failures.
- [ ] After merge: confirm \`63d3eaf9-...\` and \`99061dcf-...\` render the corrected copy in production.

### DIFF
--- a/src/__tests__/signal-page.test.ts
+++ b/src/__tests__/signal-page.test.ts
@@ -11,11 +11,13 @@ import { SELF } from "cloudflare:test";
 const APPROVED_ID = "sigpage-approved-001";
 const DRAFT_ID = "sigpage-submitted-002";
 const HOSTILE_SOURCE_ID = "sigpage-hostile-003";
+const BRIEF_PENDING_ID = "sigpage-brief-pending-004";
 const MISSING_ID = "00000000-0000-0000-0000-000000000000";
 
 const APPROVED_ADDR = "bc1qsigpageapproved0000000000000000000000000";
 const DRAFT_ADDR = "bc1qsigpagedraft00000000000000000000000000000";
 const HOSTILE_ADDR = "bc1qsigpagehostile00000000000000000000000000";
+const PENDING_ADDR = "bc1qsigpagebriefpending00000000000000000000";
 
 beforeAll(async () => {
   await SELF.fetch("http://example.com/api/test-seed", {
@@ -60,6 +62,27 @@ beforeAll(async () => {
           created_at: "2026-04-20T12:00:00Z",
           status: "approved",
           disclosure: "",
+        },
+        {
+          id: BRIEF_PENDING_ID,
+          beat_slug: "bitcoin-macro",
+          btc_address: PENDING_ADDR,
+          headline: "Signal in a compiled-but-not-yet-inscribed brief",
+          body: "brief-pending provenance copy coverage.",
+          sources: "[]",
+          created_at: "2026-04-19T08:00:00Z",
+          status: "brief_included",
+          disclosure: "",
+        },
+      ],
+      briefs: [
+        {
+          date: "2026-04-19",
+          text: "test brief body",
+          json_data: null,
+          compiled_at: "2026-04-19T23:50:00Z",
+          inscribed_txid: null,
+          inscription_id: null,
         },
       ],
     }),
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Fixes a correctness bug I've seen directly — `brief_included` signals (including ones filed to `bitcoin-macro` and `aibtc-network` beats) were rendering "currently in editorial review" copy that's wrong on both counts: those signals *have* cleared editorial review, and we do know the brief date before the inscription lands.

**What works well:**
- The discriminated union (`InscribedProvenance | BriefPendingProvenance`) is exactly the right modeling choice — TypeScript enforces exhaustive narrowing at all call sites, so future states won't silently fall through
- `getBriefByDate` is already wrapped in `.catch(() => null)` so network errors degrade gracefully to the status-based fallback
- Using `brief.date` (the canonical brief field) rather than the locally derived `getUTCDate(signal.created_at)` is correct — they should agree since the brief is fetched by that key, but `brief.date` is authoritative
- Test fixture cleanly seeds the new state with `inscription_id: null` and verifies both visible copy and JSON-LD shape — good coverage of the brief-pending path
- The JSON-LD structured data is now honest: `creditText` only claims "inscribed on Bitcoin" when the record is actually on-chain; `sameAs`/`archivedAt` don't leak for pending briefs

**[question] Edge case: brief fetch succeeds but `brief.date` is unexpected?**
`getBriefByDate(env, briefDate)` is called with `briefDate = getUTCDate(new Date(signal.created_at))`, so `brief.date` should always match. Not a bug — just confirming the fetch is keyed the same way the date is used. Nothing to change here.

**Code quality notes:**
The three-case structure in `renderProvenance` (inscribed → brief-pending → status fallback) is clear and correctly ordered. The inline comments (`// Case 1:`, `// Case 2:`, `// Case 3:`) help future readers understand the branching intent. No simplification needed — the added complexity is proportionate to the three real states.

**Operational note:** Arc files signals to `bitcoin-macro`, `quantum`, and `aibtc-network` beats. Signals filed today (2026-04-22) that are `brief_included` but not yet inscribed would have rendered the wrong copy on prod — this fix closes that gap. The specific signal IDs called out in the PR description (`63d3eaf9`, `99061dcf`) are exactly the beat-relevant cases.

biwasxyz:
Correct — confirmed with concrete evidence. Two facts back this up:

**1. \`briefs.date\` is the row key.** \`news-do.ts:2757\` uses \`ON CONFLICT(date) DO UPDATE\` on insert, so \`SELECT * FROM briefs WHERE date = ?\` can only return a row where \`date\` equals the queried value byte-for-byte.

**2. The server uses the same helper to compute the date.** When a signal moves into a brief, the DO derives the target date as \`getUTCDate(new Date(signalRow.created_at))\` (\`news-do.ts:1279\`) — the exact same call \`signal-provenance.ts\` makes before \`getBriefByDate\`. So if a signal is \`brief_included\`, it's tied to the brief keyed by that UTC calendar day, and that's the key we lookup with.

Net: \`brief.date === briefDate\` always holds when the fetch succeeds. Nothing to narrow or guard. Thanks for the spot-check.

PR #600, opened 2026-04-22
feat(seo): server-render homepage with HTMLRewriter + JSON-LD (phase 2b)

biwasxyz:
## Summary

Server-renders the homepage by intercepting \`GET /\` with a Worker that transforms the static shell in-flight — **without touching the client-side SPA code**. Google, Discover, and social crawlers now see today's lead story in the initial HTML response instead of an empty shell that only becomes content after \`/api/init\` loads and the JS paints.

Users see the exact same interactive homepage they always have — the \`/api/init\` bootstrap still runs, the topnav still renders client-side, the \`?signal=:id\` deep-link bootstrap still opens the modal.

## How it works

1. \`wrangler.jsonc\` adds \`assets.binding = "ASSETS"\` and \`assets.run_worker_first = ["/"]\`. The pattern is a **literal exact match per Cloudflare docs** — only \`/\` hits the Worker first, every other asset path (sections, CSS, JS, images) keeps serving directly with zero Worker overhead.
2. New \`homeRouter\` at \`src/routes/home-page.ts\` handles \`GET /\`:
   - Pulls the static shell via \`env.ASSETS.fetch(request)\` (canonical pattern; \`html_handling\` respected — \`/\` serves \`public/index.html\`).
   - Fetches latest brief + front-page signals in parallel via \`Promise.allSettled\`.
   - Streams the asset response through \`HTMLRewriter\`, rewriting \`<title>\` / \`description\` / \`og:*\` / \`twitter:*\` content attrs with dynamic values and appending 3 JSON-LD blocks before \`</head>\`.
3. Body DOM, scripts, and placeholder divs are untouched.

## Structured data injected

- \`NewsMediaOrganization\` — signals Google we're a news publisher (more specific than bare \`Organization\`; powers Top Stories eligibility).
- \`WebSite\` — with \`publisher\` \`@id\` reference for knowledge-graph linking.
- \`ItemList\` (up to 10) — today's front-page signals as \`ListItem[]\` with \`url\` + \`name\` so Discover can resolve top stories directly from homepage markup.

## Safety model (explicit — this is the "don't break anything" part)

1. \`env.ASSETS.fetch\` wrapped in try/catch → 503 fallback. Zero infinite-loop risk because the ASSETS binding bypasses the Worker entirely (per docs).
2. Only \`2xx text/html\` responses get transformed; 404s / redirects / binary assets pass through untouched.
3. Data fetch failures pass the shell through **untouched** — SEO loses freshness but UX is untouched.
4. HTMLRewriter only targets head tags + the closing \`</head>\`. Body DOM + all scripts unchanged.
5. Explicit \`Cache-Control: public, max-age=60, s-maxage=300\` — Cloudflare does **NOT** auto-cache Worker responses (confirmed via docs), so we set this header ourselves to match the signal-page cache strategy.

## Commits (in dependency order, each green)

- \`823d653\` — \`wrangler.jsonc\`: ASSETS binding + run_worker_first
- \`d463a5f\` — \`types.ts\` + \`x402-rpc\` test fixture: adds \`ASSETS: Fetcher\` to Env (atomic for typecheck)
- \`d2e0632\` — \`src/routes/home-page.ts\`: SSR handler
- \`f6c2cbe\` — \`src/index.ts\` + \`health.test.ts\`: wire router, retire dead JSON handler, update GET / test
- \`60f1144\` — \`home-page.test.ts\`: 9 integration tests

## Test plan

- [x] Typecheck clean on all 5 commits in sequence
- [x] 9 new homepage tests pass (title/description rewrites, JSON-LD injection, scope check for /robots.txt and /signals/:id)
- [x] Full suite: 318 pass / 4 fail (same 4 pre-existing: \`identity-gate\`, \`scoring-math\`). Zero new failures.
- [x] Classifieds x402 tests pass (no intersection; homepage handler touches neither \`POST /api/classifieds\` nor any payment middleware)
- [x] \`signal-page.test.ts\` passes (13 tests) — \`/signals/:id\` not intercepted by homeRouter (literal-exact run_worker_first match)
- [x] \`seo.test.ts\` passes — \`/robots.txt\`, \`/sitemap.xml\`, \`/news-sitemap.xml\` not intercepted
- [ ] Preview verification: hit \`GET /\` on the staging URL, confirm title reflects seeded lead, confirm 3 JSON-LD blocks, confirm existing shell JS still boots
- [ ] Production verification (post-merge): hit \`https://aibtc.news/\`, check title/description rendered from real brief, validate JSON-LD in Google Rich Results Test

## What doesn't change

- The entire existing homepage SPA (4946 lines of \`public/index.html\` + \`shared.js\`) is untouched.
- \`/api/init\` still fires on load and paints the interactive surface.
- Homepage modal on signal click — unchanged.
- \`?signal=:id\` deep-link bootstrap — unchanged and verified in tests.
- Every other route — untouched.
- The Cloudflare AI Crawl Control \`/robots.txt\` rules we discussed — completely separate, not affected.

## Next after this

Phase 3: \`/agents/:addr\` + \`/beats/:slug\` as real pages with \`ProfilePage\` / \`CollectionPage\` JSON-LD, closing the \`beatLink\` TODO from PR #597.

### DIFF
--- a/src/__tests__/health.test.ts
+++ b/src/__tests__/health.test.ts
@@ -6,12 +6,14 @@ import { SELF } from "cloudflare:test";
  * These use SELF to fetch through the real Workers runtime.
  */
 describe("GET /", () => {
-  it("returns 200 with service info", async () => {
+  // GET / now serves the server-rendered homepage (public/index.html
+  // transformed via HTMLRewriter). Detailed SSR assertions live in
+  // home-page.test.ts; here we just confirm the root still resolves
+  // successfully and returns HTML, not the old JSON service-info blob.
+  it("returns 200 HTML (homepage SSR, not service-info JSON)", async () => {
     const res = await SELF.fetch("http://example.com/");
     expect(res.status).toBe(200);
-    const body = await res.json<{ service: string; version: string }>();
-    expect(body.service).toBe("agent-news");
-    expect(typeof body.version).toBe("string");
+    expect(res.headers.get("content-type")).toMatch(/text\/html/);
   });
 });
 
--- a/src/__tests__/home-page.test.ts
+++ b/src/__tests__/home-page.test.ts
@@ -1,0 +1,188 @@
+import { describe, it, expect, beforeAll } from "vitest";
+import { SELF } from "cloudflare:test";
+
+/**
+ * Integration tests for the homepage SSR handler (GET /).
+ *
+ * With run_worker_first: ["/"] the worker intercepts the root, fetches the
+ * static shell via env.ASSETS, and streams it through HTMLRewriter to
+ * inject dynamic meta + JSON-LD. These tests verify:
+ *   - The shell loads via ASSETS and comes back as HTML.
+ *   - Head tags (title, description, og:*, twitter:*) get dynamic content
+ *     derived from the seeded brief + front-page signals.
+ *   - JSON-LD (NewsMediaOrganization, WebSite, ItemList) is injected.
+ *   - Original client bootstrap scripts (script tags, topnav placeholder,
+ *     /api/init fetch) are preserved — the client UX is untouched.
+ *   - Cache-Control is set (Worker responses aren't auto-cached at edge).
+ *   - The ?signal= deep-link bootstrap is still present.
+ *   - Other root-adjacent paths (e.g. /robots.txt) are NOT intercepted.
+ */
+
+const LEAD_HEADLINE =
+  "Homepage SSR lead — today's biggest story on AIBTC News";
+const SECOND_HEADLINE = "Second story that should appear in ItemList";
+const LEAD_ID = "home-ssr-lead-001";
+const SECOND_ID = "home-ssr-second-002";
+
+beforeAll(async () => {
+  // Seed two approved signals with fresh timestamps so listFrontPage
+  // picks them up (its window is -2 days).
+  const now = new Date();
+  const recent = new Date(now.getTime() - 60 * 60 * 1000).toISOString();
+  const earlier = new Date(now.getTime() - 90 * 60 * 1000).toISOString();
+
+  await SELF.fetch("http://example.com/api/test-seed", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({
+      signals: [
+        {
+          id: LEAD_ID,
+          beat_slug: "bitcoin-macro",
+          btc_address: "bc1qhomessrlead0000000000000000000000000000",
+          headline: LEAD_HEADLINE,
+          body: "Body content for the lead signal used in homepage SSR tests.",
+          sources: "[]",
+          created_at: recent,
+          status: "approved",
+          disclosure: "",
+        },
+        {
+          id: SECOND_ID,
+          beat_slug: "bitcoin-macro",
+          btc_address: "bc1qhomessrsecond000000000000000000000000000",
+          headline: SECOND_HEADLINE,
+          body: "Body content for the second signal.",
+          sources: "[]",
+          created_at: earlier,
+          status: "approved",
+          disclosure: "",
+        },
+      ],
+    }),
+  });
+});
+
+// ---------------------------------------------------------------------------
+// Baseline: still serves HTML with the correct cache story
+// ---------------------------------------------------------------------------
+
+describe("GET / — baseline", () => {
+  it("returns 200 HTML with explicit Cache-Control", async () => {
+    const res = await SELF.fetch("http://example.com/");
+    expect(res.status).toBe(200);
+    expect(res.headers.get("content-type")).toMatch(/text\/html/);
+    // Worker responses aren't auto-cached at the edge — we set this header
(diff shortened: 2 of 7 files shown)


arc0btc (approved):
Server-renders the homepage by intercepting `GET /` with HTMLRewriter + JSON-LD injection — **without touching the client-side SPA**. The Cloudflare-native approach is sound and the safety model is tight.

**What works well:**
- `Promise.allSettled` for parallel DO calls — one slow/failing call can't block the other, and the fallback (pass through untouched shell) is exactly right.
- `escJsonLd` correctly escapes `<` in JSON strings to prevent `</script>`-injection via headline content.
- `run_worker_first: ["/"]` being a literal exact match (not a prefix) is confirmed in tests — the scope check for `/robots.txt` and `/signals/:id` is exactly the right thing to verify.
- Test suite is thorough: 9 integration tests covering baseline, head rewrites, JSON-LD blocks, scope containment, and the `?signal=` deep-link bootstrap being preserved.
- Cache-Control on the Worker response is necessary (not auto-applied) — good that it's explicit and matches signal-page.
- Retiring the dead JSON service-info handler is the right call; it would have conflicted with `run_worker_first`.

**[suggestion] try/catch around `fetchHomepageData` won't trigger** (`src/routes/home-page.ts:210`)

`Promise.allSettled` never rejects — both DO call failures are already silently captured inside `fetchHomepageData` via the `status === "fulfilled"` checks. The outer try/catch only fires if `getLatestBrief` or `listFrontPage` throw *synchronously* before returning a promise (very unlikely). The warning message "data fetch failed" could mislead future debuggers. Consider keeping it but making the warning more accurate:

```suggestion
  } catch (err) {
    logger.warn("homepage: unexpected sync error in data fetch, passing through shell", {
      error: err instanceof Error ? err.message : String(err),
    });
    return withCacheHeaders(assetResponse);
  }
```

**[suggestion] Title length can reach 84 chars** (`src/routes/home-page.ts:93`)

`truncate(lead.headline, 70)` + ` — AIBTC News` = up to 84 chars. Google typically truncates display titles around 60-65 chars. Consider dropping the cap to 55 so the formatted title stays within display window:

```suggestion
    return `${truncate(lead.headline, 55)} — ${SITE_NAME}`;
```

Cosmetic SEO nit — JSON-LD is unaffected regardless.

**[nit]** `Jsonish` (`{ [k: string]: unknown }`) could be `Record<string, unknown>` — same semantics, standard TypeScript idiom.

**Code quality notes:**
- `setContent` helper is a clean DRY pattern for the repeated `setAttribute` calls across 4 meta properties.
- `buildTitle`, `buildDescription`, and the three JSON-LD builders are well-separated — easy to test and extend for phase 3 (`/agents/:addr` + `/beats/:slug`).
- Router mounting order (seoRouter before homeRouter) is correct — SEO routes are claimed first and won't fall through to the homepage handler.

**Operational note:** We process signal data daily through the same DO layer. The `listFrontPage` + `getLatestBrief` calls are proven stable at our 24/7 dispatch cadence — no concerns about DO reliability on this path.

biwasxyz:
Thanks both — all 5 points validated and applied in commit f629435.

**1. Edge caching via caches.default (Copilot) — applied.**
Confirmed the gap: \`withCacheHeaders\` only set Cache-Control, which doesn't populate Cloudflare's edge cache for Worker responses in this zone. Rewired to use \`edgeCacheMatch\` + \`edgeCachePut\` from \`src/lib/edge-cache.ts\` — same pattern /api/init and /api/beats already use. First request pays the full SSR cost; subsequent hits within s-maxage skip ASSETS.fetch + DO calls + HTMLRewriter entirely. This is probably the most user-visible of the five fixes.

**2. Stale validator headers (Copilot) — applied.**
Valid concern. \`new Headers(res.headers)\` copies \`ETag\` / \`Last-Modified\` / \`Content-Length\` from the original \`public/index.html\` — after HTMLRewriter modifies the body, a conditional \`If-None-Match\` could 304-with-stale-HTML and Content-Length would mismatch. Now explicitly \`headers.delete()\` all three before setting our own.

**3. try/catch around fetchHomepageData (arc0btc + Copilot) — partial.**
Both of you flagged this. I kept the try/catch but clarified the log message + added a comment explaining it's defensive against sync throws only — \`Promise.allSettled\` genuinely can't reject on the async path. Went with arc0btc's "keep it defensive, make the message honest" over Copilot's "remove entirely" because a missing DO binding at boot would otherwise bubble up as a 500 instead of a pass-through shell. Small cost, real defense.

**4. Title cap 70 → 55 (arc0btc) — applied.**
\`truncate(70) + " — AIBTC News"(13) = up to 83 chars\`. Google's desktop truncation is ~60-65. Cap at 55 gives \`~68 chars\` which fits. JSON-LD \`headline\` field is unaffected (not truncated).

**5. Jsonish → Record<string, unknown> (arc0btc nit) — applied.**
Dropped the alias, using \`Record<string, unknown>\` inline. More idiomatic, same semantics.

Test suite: 318/322 (same 4 pre-existing failures on main). Preview will redeploy on push.

PR #602, opened 2026-04-22
perf(homepage): bundle beatStats into /api/init + drop per-beat limit to 10

biwasxyz:
## Summary

Fixes the real "Today's Beats takes a minute" issue. The homepage fires 4 \`/api/signals/counts\` calls in Phase 1 that all hit the single Durable Object and serialize through it, and then 3 × \`/api/signals?beat=X&limit=200\` in Phase 2 that each return ~77 KB (230 KB total) just to extract hourly timestamps.

This PR collapses the Phase 1 fan-out to zero extra HTTP calls and shrinks Phase 2 by ~95%.

## What changes

**Backend (commit \`3dbf3cd\`):**
- New \`beatStats\` map (per-beat, per-status counts since UTC-midnight) and \`signalsCount1h\` scalar, computed inside the single existing \`/init\` DO handler via two extra SQL queries on the same SQLite context. No new DO calls.
- Mirrors the exact windowing rule \`/signals/counts\` uses (\`submitted\` → \`created_at\`; terminal statuses → \`COALESCE(reviewed_at, created_at)\`) so counts match byte-for-byte what the old per-beat fetches returned.
- \`InitBundle\` type updated; \`/api/init\` passes both fields through.

**Frontend (commit \`da8e0da\`):**
- \`beatStats\` / \`signalsCount1h\` globals populated from \`/api/init\`.
- \`renderBeatsRail\` Phase 1 drops 4 fetches — reads from globals.
- Phase 2 per-beat signals fetch: \`limit=200\` → \`limit=10\`. Sparkline (24 hourly buckets) + agentsOnline gauge both tolerate sampling.
- Fallback path (individual endpoints when \`/api/init\` fails) preserves empty-state rendering; no crash.

## Expected impact

| Path | Before | After |
|---|---|---|
| Phase 1 fan-out | 4 serialized DO calls (~6 s cold) | 0 extra calls (baked into /api/init) |
| Phase 2 per-beat | 3 × ~77 KB, ~1.8 s each | 3 × ~4 KB, <100 ms each |
| Total bandwidth savings on first load | — | **~230 KB** |
| Total wall-time savings on first load | — | **~6–7 s** |

Projected only — will measure on preview + prod after deploy.

## Test plan

- [x] Typecheck clean.
- [x] Lint: 5 pre-existing warnings in \`news-do.ts\` (not from this diff).
- [x] Full suite: 309 pass / 4 fail — same pre-existing failures (\`identity-gate\`, \`scoring-math\`). Zero new failures.
- [ ] Preview: confirm \`/api/init\` includes \`beatStats\` keyed by beat slug + \`signalsCount1h\` scalar.
- [ ] Preview: open homepage in browser; verify beats rail renders today-counts per beat and ticker shows 1h signal count.
- [ ] Post-merge prod: measure full homepage render time, compare to pre-merge baseline (~1 min).

## Related

- Builds on PR #601 (\`/api/init\` slim). Independent of PR #600 (homepage SSR).
- The \`/api/brief?format=json\` fetch in \`renderBeatsRail\` is left in place — could also be sourced from \`data.brief\` in a tiny follow-up, but kept separate here to limit blast radius to user-requested scope (#1 + #2).

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4434,15 +4434,26 @@
       // /api/init.brief in a follow-up; left separate to limit blast radius).
       const perBeatCounts = top.map(b => beatStats[b.slug] || {});
       const hourlyCounts = { total: signalsCount1h };
-      const latestBrief = await fetchJSON('/api/brief?format=json');
 
-      // limit=10 is plenty for the sparkline (24h hourly buckets show trend,
-      // not per-signal detail) and the agentsOnline gauge (distinct-address
-      // count tolerates sampling). 200 meant up to ~77KB × 3 beats = ~230KB
-      // per first load just to aggregate timestamps — pure waste.
+      // Fire brief + phase2 simultaneously so the per-beat signals fetch
+      // isn't gated on brief latency. We only block on brief before render
+      // (phase2 progressively fills the sparklines + agentsOnline gauge
+      // later via phase2.then below — same pattern as the old phase1/phase2
+      // split).
+      const briefPromise = fetchJSON('/api/brief?format=json');
+
+      // limit=50 balances two consumers of this fetch: the 24-hour
+      // sparkline (24 hourly buckets — needs enough samples to span the
+      // window without clustering all points in the last few hours) and
+      // the agentsOnline gauge (distinct-address count across top beats).
+      // A busy beat with 40+ unique filers/day would undercount at
+      // limit=10; 50 covers realistic daily volume while still being 4×
+      // lighter than the old limit=200 (~77KB × 3 = ~230KB per first load).
       const phase2 = Promise.all(top.map(b =>
-        fetchJSON('/api/signals?beat=' + encodeURIComponent(b.slug) + '&since=' + encodeURIComponent(since24h) + '&limit=10')
+        fetchJSON('/api/signals?beat=' + encodeURIComponent(b.slug) + '&since=' + encodeURIComponent(since24h) + '&limit=50')
       ));
+
+      const latestBrief = await briefPromise;
 
       // Sparkline geometry — used in both the placeholder render and the lazy
       // fill, so hoisted out of the per-tile loop.
--- a/public/index.html
+++ b/public/index.html
@@ -3077,6 +3077,13 @@
 
     let beats = [];
     let correspondentsList = [];
+    // beatStats: { [slug]: { approved, brief_included, submitted, ... } } —
+    // today's UTC-midnight counts, served by /api/init (replaces what used to
+    // be N × /api/signals/counts?beat=X fan-out on first paint).
+    // signalsCount1h: total signals in the last hour, for the Wire Status
+    // ticker (replaces the /api/signals/counts?since=1h call).
+    let beatStats = {};
+    let signalsCount1h = 0;
 
     // Top nav + LIVE ticker (shared.js provides renderTopNav / renderTicker).
     // Both are mounted before data loads so they appear instantly.
@@ -3130,6 +3137,11 @@
         const correspondentsData = data ? data.correspondents : null;
         correspondentsList = (correspondentsData && correspondentsData.correspondents) ? correspondentsData.correspondents : [];
         const signalsData = data ? data.signals : null;
+        // Per-beat counts + 1h total come from /api/init now (the handler
+        // bundles them from the same DO call). Fallback to empty/0 on the
+        // degraded path where data was stitched from individual endpoints.
+        beatStats = (data && data.beatStats && typeof data.beatStats === 'object') ? data.beatStats : {};
+        signalsCount1h = (data && typeof data.signalsCount1h === 'number') ? data.signalsCount1h : 0;
 
         const brief = data ? data.brief : null;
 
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Eliminates 4 serialized Durable Object calls from the homepage's Phase 1 load path and cuts per-beat payload from ~77 KB to ~4 KB — the kind of fix that has an outsized impact on cold-start latency when the DO is the bottleneck. Deployment is already green. Preview looks good.

**What works well:**
- The SQL windowing logic in `news-do.ts` correctly mirrors what `/signals/counts` uses — same three-part CASE (`submitted` by `created_at`, terminal statuses by `COALESCE(reviewed_at, created_at)`, others by `created_at`). Good that this is called out explicitly in the PR body; it's the kind of invariant that's easy to get wrong silently.
- Fallback to `beatStats = {}` / `signalsCount1h = 0` on the degraded path handles gracefully — no crash, just empty counts, which matches the old behavior when individual endpoints failed.
- Type updates to `InitBundle` are clean and well-documented.
- The `??` null-coalescing in `init.ts` is a nice safety net for DO responses that might omit the new fields (e.g., during a partial deploy).

**[suggestion] `latestBrief` is now awaited before `phase2` launches** (`public/index.html`, around line 4440)

Previously all phase1 fetches (beat counts + ticker + brief) ran in a single `Promise.all`, and phase2 launched immediately after phase1 resolved. The new code awaits `latestBrief` synchronously, then starts phase2 — so a slow `/api/brief` fetch gates the phase2 network fan-out unnecessarily.

```suggestion
      const briefPromise = fetchJSON('/api/brief?format=json');
      const phase2 = Promise.all(top.map(b =>
        fetchJSON('/api/signals?beat=' + encodeURIComponent(b.slug) + '&since=' + encodeURIComponent(since24h) + '&limit=10')
      ));
      const [latestBrief, beatSignals] = await Promise.all([briefPromise, phase2]);
```
(then use `beatSignals` in place of the `await phase2` result below)

This keeps brief and the per-beat signal fetches running in parallel, preserving the concurrency that existed in the old `phase1` / `phase2` split.

**[question] `limit=10` and the `agentsOnline` gauge** (`public/index.html`, sparkline + gauge computation)

The PR says "distinct-address count tolerates sampling" but if the gauge counts distinct agent addresses from the 10 most-recent signals in a 24h window, it could show `2–3 agents` on a beat that had 40+ unique filers that day. Is the gauge actually computing distinct addresses from the signals array, or pulling from a separate count? If it's from the signals array, 10 may be too sparse for beats with high signal volume.

**[suggestion] Duplicate SQL windowing WHERE clause**

The complex three-branch WHERE clause appears twice in `news-do.ts` — once for `beatStats` (window: `todayUtcMidnight`) and once for `signalsCount1h` (window: `oneHourAgo`). This is fine as-is, but the coupling isn't obvious: if the windowing rule in `/signals/counts` changes (e.g., how terminal statuses are bucketed), both queries need to change together.

At minimum, a comment linking them would help:
```
// NOTE: windowing logic must stay in sync with /signals/counts handler
// and the beatStats query above — any change to status bucketing applies to all three.
```

A CTE could unify the logic in a follow-up if the `signals` table grows and the query needs an index hint.

**[nit] Type inconsistency between declaration and production site**

`do-client.ts` declares `beatStats: Record<string, Partial<Record<string, number>>>` (inner map is Partial), but `news-do.ts` produces `Record<string, Record<string, number>>` (inner map is non-Partial — every status key is explicitly set). The Partial is defensive and not wrong, but the mismatch means consumers have to guard against undefined for status keys that the production code always sets. Could align to `Record<string, Record<string, number>>` if the DO always produces complete records.

---

**Operational note:** We file signals to `aibtc-network`, `bitcoin-macro`, and `quantum` daily and read beat counts from this homepage. The Phase 1 fan-out latency was visible on our end — beat rail rendering was consistently slow on cold sessions. This fix addresses the root cause rather than papering over it.

309 tests passing, 0 new failures, preview deploy successful — approving.

biwasxyz:
Thanks — 3 of 4 applied in commit 24eb452, one pushed back. Walk-through:

**Applied:**

1. **Parallelize brief + phase2 (valid — real regression).** You're right, I serialized phase2 behind brief by awaiting before declaring it. Restored the original concurrency: brief + phase2 fire simultaneously, only brief is awaited before render.

2. **limit=10 → 50 (valid concern).** Confirmed at \`index.html:4597-4599\`: \`realAgents = new Set(allSignals.filter(...).map(s => s.btcAddress))\` — 10 samples per beat × 3 beats = 30 max signals. A beat with 40+ unique filers/day would significantly undercount. Sparkline (24 hourly buckets) also sparse. Bumped to 50 — still 4× lighter than the old 200, but gives enough samples for accurate distinct-address count + a proper 24h trend. Cleaner long-term fix would be an aggregate \`agentsOnline24h\` field in /api/init — noted as a follow-up.

3. **SQL windowing link comment.** Added — the comment now explicitly flags that the two beatStats queries + the \`/signals/counts\` handler (~line 2220) must move together on any status-bucketing change. A CTE consolidation is mentioned as a future refactor.

**Pushed back:**

4. **Type \`Partial<Record<string, number>>\` is accurate, not defensive.** The DO's query is \`GROUP BY beat_slug, status\` which only returns rows where \`COUNT(*) > 0\`. A beat with only approved signals today produces \`beatStats[slug] = {approved: 5}\` — no \`submitted\`, \`rejected\`, etc. keys. The inner object is genuinely sparse. The frontend's existing \`c.submitted || 0\` pattern correctly handles the undefined case. Aligning to \`Record<string, number>\` would be misleading — TypeScript would claim every status key exists when in practice they often don't. Leaving \`Partial\` in place.

Full suite: 309/313 (same 4 pre-existing failures).

PR #603, opened 2026-04-22
perf(homepage): one-roundtrip init — kill 2 redundant fetches, show staleness, 30-min edge cache

biwasxyz:
## Summary

Collapses the homepage's first-paint critical path from **3 blocking fetches (\`/api/init\` → \`/api/beats\` → \`/api/brief?format=json\`)** down to **just \`/api/init\`**, bumps the edge cache from 5 min to 30 min, and surfaces honest staleness so the longer TTL is invisible to users.

Direct response to the "loading fast sometimes and very slow sometimes" report. Builds on PR #601 (init slim) and PR #602 (beatStats bundle) — this is the next logical step in the same direction.

## Measured current state

| Request | Warm | Cold |
|---|---|---|
| \`/api/init\` | 570 ms | **2.8 s** |
| \`/api/beats\` (via overrideBeatsWithCanonical) | 300 ms | **1.7 s** |
| \`/api/brief?format=json\` (in renderBeatsRail) | 450 ms | ~1 s |
| **Total warm first paint** | **~1.3 s** | |
| **Total cold first paint** | | **~5.5 s** |

## What changes

**Backend (commit \`e41ae62\`):**
1. **DO \`/init\` beats query**: add \`LEFT JOIN beat_editors\` (mirrors \`/beats\` handler), preserve DB \`retired\` status, include \`daily_approved_limit\` + \`editor_review_rate_sats\`. Beat shape now matches \`/api/beats\` so the separate call becomes redundant.
2. **\`/api/init\` response**: expose \`editor\`, \`dailyApprovedLimit\`, \`editorReviewRateSats\` on beats; add \`generatedAt: ISO\` at the top level.
3. **Cache-Control**: \`s-maxage=300\` → \`s-maxage=1800\`. Cold-miss rate drops ~6×. \`max-age\` unchanged.

**Frontend (commit \`e1718ae\`):**
4. **Delete** \`overrideBeatsWithCanonical()\` and its \`await\` call (~10 lines deleted).
5. **Delete** \`fetchJSON('/api/brief?format=json')\` in \`renderBeatsRail\` — brief now comes from \`/api/init.brief\` via a new \`globalBrief\` variable.
6. **Add** "Counts: Xm ago" row to Wire Status, computed from \`data.generatedAt\`. Makes the 30-min cache honest — users see exactly when numbers were computed.

## Why 30-min cache is safe

- \`beatStats\` / \`signalsCount1h\` are **glance info** on side rails, not primary content.
- The brief compiles once per day — the \`compiledAt\` field in \`data.brief\` is the fresh freshness signal.
- The Wire feed + signal detail pages use different endpoints (\`/api/front-page\`, \`/api/signals\`) with their own caches — signal *content* stays fresh.
- "Counts: Xm ago" surfaces exactly how stale the rail is, so the UX is honest.

## Expected impact

| Visitor type | Before | After |
|---|---|---|
| Warm cache (common case) | ~3 s full paint | **~1 s full paint** |
| Cold-cache first-time visitor | ~5.5 s full paint | **~2.5 s full paint** |
| Cold-miss frequency | every 5 min / PoP | every 30 min / PoP |

Also eliminates the "sessionStorage has stale shape" class of bugs — returning users with cached \`/api/init\` from a previous shape-change now only see stale once per 30 min of edge expiry, and the visible "Counts: Xm ago" label tells them what they're looking at.

## Backward compatibility

- Response shape is **additive** — no removed fields. \`editor\`, \`dailyApprovedLimit\`, \`editorReviewRateSats\`, \`generatedAt\` are new; everything else unchanged.
- Degraded-fallback path (when \`/api/init\` fails and we stitch from 5 individual endpoints) still works — \`globalBrief\` picks up whatever the fallback assembled.
- No API consumer greps flag \`overrideBeatsWithCanonical\` or \`/api/brief?format=json\` usage outside the homepage.

## Test plan

- [x] Typecheck clean on both commits.
- [x] Full suite: 309/313 (same 4 pre-existing failures).
- [ ] Preview: verify \`/api/init\` response includes \`editor\`, \`generatedAt\`, and \`s-maxage=1800\` in headers.
- [ ] Preview: open homepage, verify "Counts: Xm ago" renders in Wire Status, no second \`/api/beats\` fetch fires in DevTools Network tab.
- [ ] Post-merge prod: measure full homepage load — target ~1 s warm / ~2.5 s cold.

## Related

- Independent of PR #600 (homepage SSR). No file overlap.
- Builds on PR #601 (init slim) and PR #602 (beatStats bundle).

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -3084,23 +3084,19 @@
     // ticker (replaces the /api/signals/counts?since=1h call).
     let beatStats = {};
     let signalsCount1h = 0;
+    // ISO timestamp of when /api/init was generated server-side. The edge
+    // cache can serve a response up to s-maxage old; this lets the Wire
+    // Status block render "Counts: Xm ago" so the staleness is honest.
+    let initGeneratedAt = null;
+    // Latest brief (from /api/init.brief) — exposed as a global so
+    // renderBeatsRail can read compiledAt / archive without a second
+    // /api/brief?format=json round-trip.
+    let globalBrief = null;
 
     // Top nav + LIVE ticker (shared.js provides renderTopNav / renderTicker).
     // Both are mounted before data loads so they appear instantly.
     renderTopNav({ active: 'front', showUtility: false });
     renderTicker();
-
-    // /api/init returns a simplified beats shape that marks every beat as
-    // 'active' and omits the real editor field. /api/beats is the source of
-    // truth for status + editor. Override the `beats` global with it whenever
-    // available so every downstream renderer (Bureau Roster, Today's Beats
-    // rail, etc.) gets accurate statuses.
-    async function overrideBeatsWithCanonical() {
-      const canonical = await fetchJSON('/api/beats');
-      if (Array.isArray(canonical) && canonical.length) {
-        beats = canonical;
-      }
-    }
 
     async function init() {
       try {
@@ -3142,11 +3138,15 @@
         // degraded path where data was stitched from individual endpoints.
         beatStats = (data && data.beatStats && typeof data.beatStats === 'object') ? data.beatStats : {};
         signalsCount1h = (data && typeof data.signalsCount1h === 'number') ? data.signalsCount1h : 0;
+        initGeneratedAt = (data && typeof data.generatedAt === 'string') ? data.generatedAt : null;
 
         const brief = data ? data.brief : null;
+        globalBrief = brief;
 
-        // Override with canonical beats data (/api/init has stale status field)
-        await overrideBeatsWithCanonical();
+        // (Previously called overrideBeatsWithCanonical here to get editor +
+        // proper retired status. /api/init now returns the full canonical
+        // beat shape including `editor`, `dailyApprovedLimit`, and respects
+        // the retired flag — so the second /api/beats round-trip is gone.)
 
         // Seed agentCache from correspondents data (no extra network call needed)
         for (const c of correspondentsList) {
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Collapses homepage first-paint from 3 blocking fetches to 1 by bundling editor data + brief into `/api/init`, bumping the edge cache to 30 min, and surfacing honest staleness — solid incremental step in the #601/#602 series.

**What works well:**
- The DO-side change correctly mirrors the `/beats` handler's `LEFT JOIN beat_editors` pattern — no divergence risk on the beat shape between the two endpoints
- Preserving `retired` status (instead of always overriding from `last_signal_at`) fixes a subtle bug where retired beats could ghost back as inactive on the homepage
- `generatedAt` is frozen at generation time, not serve time — correct; a 20-min-old cached response will accurately show "20m ago"
- The degraded-fallback path analysis in the commit message is thorough; `globalBrief` picks up whatever the fallback assembled, so no regression there
- 309/313 test pass rate (same 4 pre-existing failures) — clean

**[suggestion] Negative clock skew in staleness label** (`public/index.html`, `countsAgeLabel` block)

If the visitor's clock is behind the server clock, `ageMs` will be negative and `ageMin` will be a negative integer — the label renders as `-1m ago` or worse. Low probability but easy to guard:

```suggestion
        const ageMin = Math.max(0, Math.floor(ageMs / 60000));
```

**[question] Editor map ordering under multiple active editors** (`src/objects/news-do.ts`, `editorByBeat` loop)

The query orders by `registered_at DESC` (newest first), but the loop uses `editorByBeat.set(...)` which overwrites on each iteration — so if two rows share the same `beat_slug`, the *oldest* entry wins (last write). The commit message says "one per beat enforced by registration logic," which makes this safe in practice. Just confirming that invariant is enforced at insert time, not just relied on at read time?

**[suggestion] `as Beat` cast may shadow type errors** (`src/objects/news-do.ts:4722`)

The returned object now includes `daily_approved_limit`, `editor_review_rate_sats`, and `editor` — if the `Beat` interface wasn't updated to include these, the `as Beat` cast suppresses the TypeScript error silently. Worth checking the interface definition to confirm these fields are declared (or use `satisfies Beat` which would catch shape mismatches without widening).

**Code quality notes:**
- The IIFE `(() => { ... })()` inside the ternary for active/inactive status is a bit unusual — could flatten to a single expression: `lastSignalAt && now - new Date(lastSignalAt).getTime() < expiryMs ? "active" : "inactive"`. Not blocking, just slightly simpler.
- The `beat_editors` query fetches all active editors regardless of which beats are being returned. Fine at current scale; worth a `WHERE beat_slug IN (...)` if the table grows.

**Operational note:** We run the aibtc.news signal pipeline 24/7 and the "loading fast sometimes, very slow sometimes" pattern matches exactly what we see with cold-miss spikes. The 6× reduction in cold-miss frequency should eliminate the worst-case delays our sensors observe when checking beat activity immediately after an edge eviction.

PR #614, opened 2026-04-23
feat(seo): server-render /agents/:addr + /beats/:slug (phase 3)

biwasxyz:
## Summary

Phase 3 SSR: gives each AI correspondent and each beat a clean, canonical, indexable URL with real content in the initial HTML response. Follows the Phase 2A (`/signals/:id`) pattern exactly — dedicated Hono routers, full HTML + JSON-LD + BreadcrumbList, no touch to the client SPA.

- **`GET /agents/:addr`** → `ProfilePage` + `Person` (with `BitcoinAddress` identifier) + `ItemList` of the agent's signals.
- **`GET /beats/:slug`** → `CollectionPage` + `ItemList` of up to 20 approved/brief_included signals on the beat.
- **`signal-page.ts`** updated to link authors + beats to the new path-param URLs. Old JSON-LD referenced `/agents/?addr=...#agent`; new one references `/agents/:addr#person` to match the `@id` exposed by `ProfilePage.mainEntity`.
- **Sitemap** gains `/sitemap/beats.xml` + `/sitemap/agents.xml`. Retired beats are excluded from the beats sitemap (the pages stay reachable with noindex). Zero-signal addresses are excluded from the agents sitemap.

## How it works

1. **No `run_worker_first` change.** `/agents/:addr` and `/beats/:slug` have no matching static file, so Cloudflare Assets bypasses both and the Worker owns them naturally. The existing `/agents/` and `/beats/` listing pages continue serving directly from Assets with zero Worker overhead — same pattern as `/signals/:id` today.
2. **Input validation gates the DO round-trip.** `normalizeAddr` and `isValidSlug` reject obviously-malformed paths (e.g. `/agents/foo.php`) with 404 + noindex before any DO call.
3. **Parallel fetch on the beat page.** `getBeat` + `listSignals` run via `Promise.all`, so a slow signal query doesn't block the beat lookup.
4. **Zero-signal addresses 404.** `getAgentStatus` returns a populated object even for unknown addresses — we explicitly treat `totalSignals === 0 && signals.length === 0` as "nothing indexable here" and return 404 + noindex.

## Commits

- `a9c7e17` — `src/routes/agent-page.ts` + mount.
- `4da2ac5` — `src/routes/beat-page.ts` + mount.
- `b1fdc15` — `src/routes/signal-page.ts` link + JSON-LD updates.
- `6f80299` — `src/routes/seo.ts` sitemap additions.
- `9cdcc40` — integration tests for both new routes (+ zero-signal 404 fix).

## Test plan

- [x] Typecheck clean across all 5 commits.
- [x] 17 new tests pass (agent-page × 9, beat-page × 8).
- [x] Full suite: 335 pass / 4 fail — same 4 pre-existing failures as `main` (`identity-gate` × 3, `scoring-math` × 1). Zero new failures.
- [x] signal-page.test.ts still passes after URL changes (no test asserted on the old `/agents/?addr=` pattern).
- [x] x402 classifieds + payment tests untouched and passing (no intersection with this PR).
- [ ] Preview verification: hit `/agents/<a-known-addr>` and `/beats/bitcoin-macro` on the staging URL; verify JSON-LD in Google Rich Results Test.
- [ ] Production verification (post-merge): same on `https://aibtc.news/`.

## What doesn't change

- `/agents/` and `/beats/` listing pages — untouched (static asset).
- `/signals/:id` HTML rendering — unchanged, only the author/beat hrefs and JSON-LD `@id`s point at the new URLs.
- `/api/*`, x402 classifieds, every other route — unchanged.
- No client JS changes. No wrangler / bindings changes.

## Next

Phase 3b: `/archive`, `/classifieds`, `/wire`, `/skills`, `/collection`, `/about` as a single follow-up PR with uniform `ItemList` / `WebPage` JSON-LD.

### DIFF
--- a/src/__tests__/agent-page.test.ts
+++ b/src/__tests__/agent-page.test.ts
@@ -1,0 +1,143 @@
+import { describe, it, expect, beforeAll } from "vitest";
+import { SELF } from "cloudflare:test";
+
+/**
+ * Integration tests for the agent profile page (GET /agents/:addr).
+ *
+ * Seeds two approved signals for a known test address so the handler has
+ * something to render in the ItemList / recent-signals HTML.
+ */
+
+const AGENT_ADDR = "bc1qagentpagetest000000000000000000000000000";
+const OTHER_ADDR = "bc1qunknownagent000000000000000000000000000x";
+const LEAD_HEADLINE = "Agent-page SSR lead — first seeded signal";
+const SECOND_HEADLINE = "Agent-page SSR second — also seeded";
+const LEAD_ID = "agent-page-lead-001";
+const SECOND_ID = "agent-page-second-002";
+
+beforeAll(async () => {
+  const now = Date.now();
+  const recent = new Date(now - 30 * 60 * 1000).toISOString();
+  const earlier = new Date(now - 90 * 60 * 1000).toISOString();
+
+  await SELF.fetch("http://example.com/api/test-seed", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({
+      signals: [
+        {
+          id: LEAD_ID,
+          beat_slug: "bitcoin-macro",
+          btc_address: AGENT_ADDR,
+          headline: LEAD_HEADLINE,
+          body: "Body content for the lead signal in agent-page tests.",
+          sources: "[]",
+          created_at: recent,
+          status: "approved",
+          disclosure: "",
+        },
+        {
+          id: SECOND_ID,
+          beat_slug: "bitcoin-macro",
+          btc_address: AGENT_ADDR,
+          headline: SECOND_HEADLINE,
+          body: "Body content for the second agent signal.",
+          sources: "[]",
+          created_at: earlier,
+          status: "approved",
+          disclosure: "",
+        },
+      ],
+    }),
+  });
+});
+
+// ---------------------------------------------------------------------------
+// Baseline
+// ---------------------------------------------------------------------------
+
+describe("GET /agents/:addr — baseline", () => {
+  it("returns 200 HTML with cache headers for a known agent", async () => {
+    const res = await SELF.fetch(`http://example.com/agents/${AGENT_ADDR}`);
+    expect(res.status).toBe(200);
+    expect(res.headers.get("content-type")).toMatch(/text\/html/);
+    expect(res.headers.get("cache-control")).toMatch(/s-maxage=/);
+  });
+
+  it("renders the short + full address in the header", async () => {
+    const res = await SELF.fetch(`http://example.com/agents/${AGENT_ADDR}`);
+    const body = await res.text();
+    // Full address should appear in the <code class="ap-addr-full"> block.
+    expect(body).toContain(AGENT_ADDR);
+    // Canonical URL should be the clean path-param form.
+    expect(body).toContain(
+      `<link rel="canonical" href="https://aibtc.news/agents/${AGENT_ADDR}">`
+    );
+  });
+
+  it("includes the seeded signals in the rendered recent list", async () => {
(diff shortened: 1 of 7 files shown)


arc0btc (approved):
Phase 3 SSR — gives `/agents/:addr` and `/beats/:slug` canonical, indexable URLs with full structured data. Follows the Phase 2A `/signals/:id` pattern cleanly, and the JSON-LD `@id` alignment between `agent-page.ts` (`#person`) and the updated `signal-page.ts` reference is exactly right for knowledge-graph consistency.

**What works well:**
- Input validation before the DO round-trip on both routes (`normalizeAddr`, `isValidSlug`) — good defense against scanner junk without any lock-in to specific extension patterns.
- The zero-signal 404 gate in `agent-page.ts` is the right call. The DO returning a populated object for unknown addresses was a subtle edge case worth explicitly testing.
- `Promise.all` for beat + signals in `beat-page.ts` — parallel fetch was the obvious win here and it's done.
- Retired beat handling: noindex at the page level, excluded from the sitemap entirely. Clean separation.
- `escJsonLd` escaping `<` to prevent `</script>` injection is the correct approach for JSON-LD in script tags.
- 17 tests with good coverage across the 404 paths — the zero-signal address test in particular exercises the fix that shipped in the last commit.

---

**[suggestion] Shared utilities across both route files** (`src/routes/agent-page.ts`, `src/routes/beat-page.ts`)

`esc()`, `escJsonLd()`, `Jsonish`, `buildOrganization()`, `buildSignalList()`, `jsonLdScript()`, and the module-level constants (`SITE_URL`, `SITE_NAME`, `ORG_ID`, `OG_IMAGE`) are duplicated identically between both new files — and likely partially overlap with `signal-page.ts` too. These are a maintenance footprint: a future change to `buildOrganization()` or the escaping logic needs to land in three places.

A `src/lib/seo-utils.ts` shared helper would collapse this. Not blocking for this PR since the duplication is mechanically correct, but Phase 3b is the natural moment to refactor before the pattern gets wider.

**[question] Agent signal visibility filtering** (`src/routes/agent-page.ts:858-865`)

`fetchBeatSignals` on the beat page explicitly filters by `isPubliclyVisible()` before rendering. The agent page gets signals via `getAgentStatus(env, addr)` and passes them directly to `renderSignalList` without an equivalent filter. Does `getAgentStatus` already return only approved/brief_included signals? If the DO includes submitted or rejected signals in the status response, they'd appear in the rendered HTML and JSON-LD `ItemList`. Worth verifying — or adding a `isPubliclyVisible` guard on the agent path to match the beat path's explicit safety.

**[nit] Double-slice in signal renderers** (`src/routes/beat-page.ts:1065`, `src/routes/agent-page.ts:537-538`)

Both `renderSignalList` functions call `.slice(0, SIGNAL_LIST_CAP)` internally, but the caller already capped the data upstream (`fetchBeatSignals` returns at most `SIGNAL_LIST_CAP * 2` filtered items, and `buildSignalList` also slices). The inner slice is harmless but redundant — one source of truth for the cap is cleaner.

**[suggestion] `CorrespondentRow` field name ambiguity** (`src/routes/seo.ts:1478`)

```ts
.filter((r) => (r.total_signals ?? r.signal_count ?? 0) > 0)
```

The `??` fallback chain suggests `CorrespondentRow` has two possible field names for the same value. If that's a live inconsistency in the type, it's worth normalizing in the type definition before the agents sitemap ships to production — a future rename of one field would silently make the sitemap include zero-signal addresses.

---

**Operational context:** We file signals from `bc1qlezz2cgktx0t680ymrytef92wxksywx0jaw933` — once this ships, our correspondent profile page will be indexable. The `currentStreak` / `longestStreak` stats surface nicely from the DO status response. Good for the ecosystem's agent-identity story.

The test suite baseline (335 pass / 4 pre-existing failures, 0 new) is clean. Approve.

PR #626, opened 2026-04-23
feat(news): derive paymentIdentifier from txHex for V2 RPC idempotency

### LINKED ISSUE #624: news-do: send paymentIdentifier with X402_RELAY.submitPayment for V2-parity idempotency
## Context

`src/objects/news-do.ts` calls `env.X402_RELAY.submitPayment(txHex, settle)` for beat payment submissions. Like landing-page, we don't pass an idempotency key, so legitimate retries (network timeout, alarm-triggered re-submission) collapse into `SENDER_NONCE_DUPLICATE` rejections at the relay's in-flight marker check.

The x402 V2 HTTP path already implements canonical idempotency via the `payment-identifier` extension governed by `tx-schemas` `CanonicalDomainBoundary.paymentIdentity`. The RPC path is being brought to parity (see depends-on).

## Depends on

- [aibtcdev/tx-schemas#28](https://github.com/aibtcdev/tx-schemas/issues/28) — `RpcSubmitPaymentArgsSchema.paymentIdentifier` field
- [aibtcdev/x402-sponsor-relay#351](https://github.com/aibtcdev/x402-sponsor-relay/issues/351) — relay routes RPC `submitPayment` through `PaymentIdService`

## Proposed change

For each beat payment submission in `news-do.ts`:

1. Derive a stable `paymentIdentifier` deterministic across retries within the relay's 5-minute idempotency window. For news, the natural key is `(beatId, signerHash, nonce)` or `(payerAddress, nonce, beatId)`:
   ```ts
   const idInput = `${beatId}|${senderAddress}|${nonce}`;
   const paymentIdentifier = `pay_${(await sha256(idInput)).slice(0, 28)}`;
   ```
2. Pass through to `env.X402_RELAY.submitPayment(txHex, settle, paymentIdentifier)`.
3. Bump `@aibtc/tx-schemas` to the version that ships #28.
4. Handle `PAYMENT_IDENTIFIER_CONFLICT` in the existing RPC error mapping — same handling as a non-retryable rejection.

The alarm-sweep path (`payment-stage-alarm-sweep.test.ts`) does not submit; it only polls `checkPayment(paymentId)`, which is unaffected. This change is submit-side only.

## Acceptance

- [ ] Deterministic `paymentIdentifier` derived per beat payment from a stable `(beatId, sender, nonce)` triple
- [ ] Identifier passed through to `env.X402_RELAY.submitPayment(...)`
- [ ] `PAYMENT_IDENTIFIER_CONFLICT` mapped in the RPC error handler
- [ ] Test: same beat re-submission produces same identifier; relay return value treated as idempotent success
- [ ] Test: different beats produce different identifiers

whoabuddy:
## Summary

- Derives a deterministic `paymentIdentifier` from the transaction hex using Web Crypto SHA-256, producing `pay_<28-hex-chars>` per `PaymentIdentifierSchema`
- Passes the identifier as the third arg to `X402_RELAY.submitPayment(txHex, settle, paymentIdentifier)` — same tx on retry → same identifier → relay returns cached `paymentId` idempotently
- Maps `PAYMENT_IDENTIFIER_CONFLICT` to a non-retryable 402 (client error, no Retry-After header)
- Bumps `@aibtc/tx-schemas` to `^1.1.0` (ships `RpcSubmitPaymentRequestSchema.paymentIdentifier` and `PAYMENT_IDENTIFIER_CONFLICT` error code)

## Context

Closes #624. Brings agent-news RPC path to canonical x402 V2 idempotency parity — the same contract already live in:
- [aibtcdev/x402-sponsor-relay#355](https://github.com/aibtcdev/x402-sponsor-relay/pull/355) — relay routes RPC `submitPayment` through `PaymentIdService`
- [aibtcdev/tx-schemas#29](https://github.com/aibtcdev/tx-schemas/pull/29) — `RpcSubmitPaymentRequestSchema.paymentIdentifier` field + `PAYMENT_IDENTIFIER_CONFLICT` error code (shipped as tx-schemas 1.1.0)

### Deterministic key composition

```ts
// In src/lib/helpers.ts — derivePaymentIdentifier(txHex)
const input = txHex;  // same tx on retry → same identifier
const hashHex = sha256hex(input);
return `pay_${hashHex.slice(0, 28)}`;  // 32 chars, satisfies [a-zA-Z0-9_-]{16,128}
```

The identifier is derived from `txHex` alone (extracted from the payment header inside `verifyPayment()`). The same pre-signed transaction resubmitted on network timeout produces an identical identifier, allowing the relay to return the cached `paymentId` without duplicate nonce consumption.

## Test plan

- [x] `derivePaymentIdentifier` is deterministic for the same inputs
- [x] `derivePaymentIdentifier` produces different outputs for different inputs
- [x] Identifier matches `^pay_[a-f0-9]{28}$` (satisfies `PaymentIdentifierSchema`)
- [x] `submitPayment` receives the derived identifier as third arg (mock-verified)
- [x] Retry with the same txHex produces the same identifier (determinism across calls)
- [x] `PAYMENT_IDENTIFIER_CONFLICT` maps to 402 non-retryable with no Retry-After
- [x] All existing x402-rpc, x402-error-mapping, and helpers tests still pass (60 total)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -358,6 +358,19 @@
         "node": ">=12"
       }
     },
+    "node_modules/@emnapi/core": {
+      "version": "1.9.2",
+      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.9.2.tgz",
+      "integrity": "sha512-UC+ZhH3XtczQYfOlu3lNEkdW/p4dsJ1r/bP7H8+rhao3TTTMO1ATq/4DdIi23XuGoFY+Cz0JmCbdVl0hz9jZcA==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "peer": true,
+      "dependencies": {
+        "@emnapi/wasi-threads": "1.2.1",
+        "tslib": "^2.4.0"
+      }
+    },
     "node_modules/@emnapi/runtime": {
       "version": "1.8.1",
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.8.1.tgz",
@@ -365,6 +378,18 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
+      }
+    },
+    "node_modules/@emnapi/wasi-threads": {
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.1.tgz",
+      "integrity": "sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "peer": true,
       "dependencies": {
         "tslib": "^2.4.0"
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "agent-news",
       "version": "1.24.0",
       "dependencies": {
-        "@aibtc/tx-schemas": "^0.5.2",
+        "@aibtc/tx-schemas": "^1.1.0",
         "@noble/curves": "^2.0.1",
         "@noble/hashes": "^2.0.1",
         "@noble/secp256k1": "^2.1.0",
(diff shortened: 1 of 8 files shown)


arc0btc (approved):
Brings the agent-news RPC path to V2 idempotency parity — same pattern already live in the relay and tx-schemas, and this is the correct client-side half.

**What works well:**
- `derivePaymentIdentifier` is clean: pure SHA-256 on `txHex`, variadic for future flexibility, and the `pay_<28-hex>` format satisfies `PaymentIdentifierSchema` exactly.
- Using `txHex` as the sole input at the call site is the right choice — same pre-signed tx on retry → same identifier → relay returns cached `paymentId` without consuming a new nonce.
- Test coverage is thorough: determinism, format, single-arg production use case, and the new error code path all covered. 60 existing tests still pass.
- Mapping `PAYMENT_IDENTIFIER_CONFLICT` as non-retryable (no `Retry-After` header) is correct — this is a client-side protocol error, not a transient relay failure.

**[question] Terminal reason label** (`src/services/x402.ts`, `RPC_ERROR_TO_TERMINAL_REASON` map)

The value `"unknown_payment_identity"` reads as "we don't know who this payment is from" rather than "idempotency key conflict." This affects internal tracking/logging, not the HTTP response, but an ambiguous terminal reason makes post-hoc debugging harder. Something like `"identifier_conflict"` would be clearer. Is `PaymentTerminalReason` in tx-schemas 1.1.0 constrained to specific values, or is it open? If constrained, TypeScript would have caught a mismatch, so I assume it compiled — but worth confirming the label is intentional.

**Code quality notes:**
- The `|" separator in `parts.join("|")" is safe for hex and address inputs, but worth a brief comment that `|` was chosen deliberately as a separator that won't appear in hex strings, in case the multi-arg API gets extended later.
- Minor: the variadic `...parts` API is flexible but the sole production call site uses a single arg. No YAGNI issue since the tests exercise multi-arg and the helper is general-purpose.

*Operational note:* We process x402 payments daily and the double-submission risk on network timeout is real — we've seen retry storms hit the relay when RPC calls don't get an ack. This fix eliminates that class of nonce double-spend at the source. The major version bump to `@aibtc/tx-schemas` (0.5.2 → 1.1.0) is expected given tx-schemas#29.

PR #641, opened 2026-04-24
fix: stop /beats+/agents rendering empty during cold-DO cache miss

biwasxyz:
## Summary

Three fixes to the data-fetch pain reported after #620 landed.

### 1. SWR on `/api/correspondents`

The `/api/correspondents` DO query was observed at **22s–146s** on a cold-PoP MISS (measured with cache-busted query strings: `146s`, `3.4s`, `25s` across three consecutive cold fetches). The DO hibernates during quiet windows and cold-boots on the next access.

With the prior 5-min edge TTL, every PoP paid that cost after every quiet window, and both `/beats` and `/agents` sat staring at a spinner — or rendered `No correspondents yet` if the browser gave up.

- `src/lib/edge-cache.ts` grows `edgeCacheMatchSWR()` + `triggerSWRRefresh()`.
- Cache layout: `s-maxage=1800` (30 min edge TTL), `freshSeconds=300`.
  - age ≤ 300s → serve HIT.
  - 300s < age ≤ 1800s → serve STALE immediately, fire a background rebuild guarded by a KV lock so concurrent stale hits don't stampede the DO.
  - age > 1800s → CF evicts; next request pays the full MISS cost.
- Only the very first visitor to a PoP after a full 30-min absence pays the cold-boot cost. Everyone else either gets fresh HITs or instant STALE hits while the rebuild runs in the background.

### 2. `/beats` no longer blocks on `/api/correspondents`

Previously `Promise.all([/api/beats, /api/correspondents])` blocked first paint on the 370 KB correspondents payload. If the browser gave up mid-flight, the page rendered `No beats found.`

- The "Correspondent Roster · Sort by score · 30d" section is removed (per request). `-358 / +44` lines on the page.
- Beat cards still show a human editor name. `/api/correspondents` fires non-blocking; editor labels on already-rendered cards are upgraded in place via `fillEditorName()` once it lands.
- The page now paints on `/api/beats` alone (≤50ms warm) and never shows `No beats found` during a slow correspondents fetch.

### 3. `/agents` listing shows a skeleton instead of "No correspondents yet"

The list view jumped straight to `No correspondents yet` when the fetch was still in flight. Now it paints a skeleton during the fetch and only shows an error (with a retry button) if `/api/correspondents` actually returns `null`.

### 4. Homepage `<title>` now reads brand-first

Was: `arxiv 2604.19481 Ships 'Walking Cat'… — AIBTC News`
Now: `AIBTC News — arxiv 2604.19481 Ships 'Walking Cat'…`

Browser tabs and SERPs display the leading characters, so the homepage should identify as AIBTC News before any individual signal headline. Applied to `<title>`, `og:title`, `twitter:title`. Test in `home-page.test.ts` updated to match.

## Test plan
- [x] `npm run typecheck` — clean
- [x] `npm run lint` — no new warnings in touched files
- [x] `npm test` — 335 pass / 4 fail (same pre-existing failures as `main`, per #620's commit note)
- [x] `home-page.test.ts` — all 9 tests pass with the new brand-first title format
- [ ] After deploy: confirm `/api/correspondents` shows `X-Edge-Cache: STALE` during a background refresh window
- [ ] After deploy: confirm `/beats` renders cards immediately even when correspondents is cold
- [ ] After deploy: confirm `/agents/` shows a loading skeleton instead of `No correspondents yet`
- [ ] After deploy: confirm homepage title reads `AIBTC News — …` (may take up to 5 min per PoP for old cached homepage HTML to expire, or purge via CF dashboard)

### DIFF
--- a/public/agents/index.html
+++ b/public/agents/index.html
@@ -612,6 +612,21 @@
       root.innerHTML = html;
     }
 
+    function paintListLoading(root) {
+      let html = '<div class="list-hero">'
+        + '<h1>Correspondents</h1>'
+        + '<span class="subtitle">Loading roster…</span>'
+        + '</div>'
+        + '<div class="list-table-wrap"><table class="list-table"><thead><tr>'
+        + '<th>#</th><th>Agent</th><th>Beat</th><th>Signals</th><th>Streak</th><th>Earned</th><th>Score</th>'
+        + '</tr></thead><tbody>';
+      for (let i = 0; i < 8; i++) {
+        html += '<tr><td colspan="7" style="padding:0"><div class="sk-stack"><div class="sk sk-row"></div></div></td></tr>';
+      }
+      html += '</tbody></table></div>';
+      root.innerHTML = html;
+    }
+
     async function renderList() {
       document.title = 'Correspondents — AIBTC News';
       const root = document.getElementById('root');
@@ -624,7 +639,21 @@
         paintCorrespondentList(root, cached);
         return;
       }
+      // No sync cache — paint a loading state while we wait on the network
+      // so the page never shows "No correspondents yet" during a cold-PoP
+      // fetch. The endpoint is edge-cached with SWR on the server side, so
+      // most real visits will land on a warm cache; this skeleton only
+      // shows for the first visitor to a given PoP after the 30-min edge
+      // TTL expires.
+      paintListLoading(root);
       const data = await fetchJSON('/api/correspondents');
+      if (data === null) {
+        root.innerHTML = '<div class="list-hero"><h1>Correspondents</h1></div>'
+          + '<div class="dash-err"><h2>Couldn\'t load correspondents</h2>'
+          + '<p>The roster service didn\'t respond in time. Try a reload in a moment.</p>'
+          + '<p><button class="btn" onclick="renderList()">Retry</button></p></div>';
+        return;
+      }
       paintCorrespondentList(root, data);
     }
 
--- a/public/beats/index.html
+++ b/public/beats/index.html
@@ -177,151 +177,6 @@
     }
     .beat-card-editor .who { color: var(--text); font-weight: 600; }
 
-    /* Roster */
-    .roster-toolbar {
-      display: flex;
-      align-items: baseline;
-      gap: 10px;
-      flex-wrap: wrap;
-      margin-bottom: 10px;
-    }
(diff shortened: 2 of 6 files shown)


arc0btc (approved):
Fixes the cold-DO MISS that left `/beats` and `/agents` rendering empty or blocking on a 22s–146s round-trip — a real problem, well-solved.

**What works well:**
- The SWR implementation in `edge-cache.ts` is architecturally correct: timestamp written before clone so the cached entry carries `X-Edge-Cached-At` while only the live response gets `X-Edge-Cache: MISS`. Subtle ordering, handled right.
- KV lock with 120s TTL + eager delete in `finally` is the right shape for stampede control without over-coordinating. Acknowledging the eventual-consistency caveat (2–3 concurrent rebuilds may fire) in the comment is honest and appropriate.
- Non-blocking `fetchJSON('/api/correspondents')` on `/beats` with `fillEditorName()` DOM upgrade is clean progressive enhancement. Editor labels degrade to truncated address instead of blocking first paint — the right tradeoff.
- `paintListLoading` skeleton on `/agents` + null-check error state with retry button is a genuine UX improvement over the silent "No correspondents yet" on an in-flight fetch.
- `buildCorrespondentsResponse` extraction makes the SWR dispatch in the route handler readable at a glance.

**[nit] Double URL parse in `triggerSWRRefresh`** (`src/lib/edge-cache.ts`)

```suggestion
  const { pathname, search } = new URL(c.req.url);
  const lockKey = `swr-lock:${bucket}:${pathname}${search}`;
```

Constructs `new URL(c.req.url)` twice. One parse, destructure what you need.

**[nit] `edgeCachePut` mutates the caller's response headers**

The live response returned to the route caller now carries `X-Edge-Cached-At` in its headers (set before the clone). Not harmful, but a future maintainer might not expect it. A brief comment near the header set would make the intent explicit: *"stamp is written to both live response and cache entry — harmless on the wire, used by edgeCacheMatchSWR"*.

**Operational context:**
We run sensors against agent-news 24/7. The correspondents cold-boot latency has been a visible pain point — `/beats` rendering empty after quiet windows is something our monitoring has caught. This fix directly addresses the symptom at the right layer (edge SWR, not client retry). The 30-min outer TTL with 5-min freshness window is a reasonable calibration given DO hibernation patterns.

CI green, Cloudflare preview deployed successfully. Nits are cosmetic — ready to merge.

PR #643, opened 2026-04-24
perf(homepage): fail-fast on cold DO + SWR on edge cache

biwasxyz:
## Symptom

After #641 merged into main, the Cloudflare deploy invalidated every PoP's edge cache. I probed production right after the deploy:

| URL | First MISS | Subsequent HIT |
|-----|-----------:|---------------:|
| `/` | **19.8 s** | 50 ms |
| `/api/correspondents` | 13.3 s | 50 ms |
| `/api/init` | 4.1 s | 50 ms |
| `/api/beats` | 2.1 s | 50 ms |
| `/api/front-page` | 1.0 s | 50 ms |

First visitor to each PoP paid a cold Durable Object boot. Homepage blocked on `Promise.allSettled([ getLatestBrief, listFrontPage ])` before returning any HTML, so users saw a ~20 s blank page.

SWR from #641 can't help on a *deploy-cleared* cache — there's nothing to serve stale.

## Fix

Fail-fast the DO on homepage MISS:

1. Fetch the static shell (~50 ms from CF asset cache).
2. Race the DO fetch against a **500 ms** timeout. Warm DO returns in <100 ms so the happy path still gets the full HTMLRewriter rewrite + cache.
3. On timeout, return the un-rewritten shell immediately and kick off a background rebuild via `waitUntil` that populates the cache for the next visitor. First visitor gets generic SEO metadata; every subsequent visitor gets the fully-rewritten cached copy.
4. KV-locked background refresh (same mechanism introduced in #641 for `/api/correspondents`) keeps a flurry of concurrent cold-miss visitors from each launching their own DO call.

Also adopt SWR on hits: `edgeCacheMatchSWR` with `freshSeconds: 300` means `300 < age ≤ 1800 s` serves STALE instantly and triggers a background refresh. Cold-DO pain only ever surfaces on the very first request after a full cache eviction; hibernation-induced slowness is invisible to users from that point on.

Edge TTL bumped `s-maxage=300 → s-maxage=1800` to match the SWR window.

## Tradeoff

The first visitor after cache eviction now sees the **default** SEO metadata (`AIBTC News — News for agents that use Bitcoin`) instead of today's lead headline. The page *content* is identical — the SPA boots and paints normally. Only the crawler-only `<title>` / `<meta>` / JSON-LD is generic on that very first response, and the cache fills within a second or two for everyone after.

## Test plan
- [x] `npm run typecheck` — clean for `home-page.ts`. (Unrelated pre-existing error in `x402-rpc.test.ts:454` from commit e132c08, not touched here.)
- [x] `npm test -- home-page` — all 9 tests pass. Tests skip the cache (`isTestEnv` guard) and the test DO is fast so the happy path is exercised; the fail-fast branch is guarded by timing and needs production to exercise.
- [ ] After deploy: `curl -s -o /dev/null -w "%{time_total}\n" https://aibtc.news/` first-MISS should be ≤600 ms (static shell + 500 ms timeout ceiling), not 20 s.
- [ ] After deploy: second MISS-on-fresh-PoP (different PoP) should also be ≤600 ms.
- [ ] After deploy: warmed PoP should serve `X-Edge-Cache: HIT` / `STALE` depending on age.

### DIFF
--- a/src/routes/home-page.ts
+++ b/src/routes/home-page.ts
@@ -13,25 +13,34 @@
  *   - Once the client JS boots, it overrides the DOM as usual — users see
  *     the same interactive homepage they always have.
  *
- * Safety model:
- *   1. Fetch shell via env.ASSETS.fetch(). If that fails → 503.
- *   2. Only transform 2xx + text/html responses. 404 / redirects pass through.
- *   3. Fetch brief + signals in parallel with Promise.allSettled so one
- *      slow or failing DO call does not block the other. On total failure,
- *      pass through the untouched asset — SEO takes a hit but UX is fine.
- *   4. HTMLRewriter only targets specific head tags + the closing </head>.
- *      Body DOM / scripts are not touched.
- *   5. Actually cache at the edge via the Workers Cache API (caches.default
- *      through src/lib/edge-cache.ts). `Cache-Control` alone does not
- *      populate the edge cache for Worker responses in this zone — we
- *      have to put-and-match explicitly. Matches the pattern used by
- *      /api/init and /api/beats.
+ * Cold-DO safety:
+ *
+ *   The Durable Object that backs getLatestBrief / listFrontPage hibernates
+ *   during quiet windows and can take 10–130s to cold-boot. Without
+ *   protection, a fresh deploy (which wipes the edge cache) or a period of
+ *   low traffic would force the next visitor per PoP to wait for that
+ *   cold boot before seeing any HTML.
+ *
+ *   Fail-fast pattern: on cache miss we race the DO fetch against a 500ms
+ *   timeout. If the DO answers in time, we do the full rewrite + cache.
+ *   If it doesn't, we serve the un-rewritten static shell immediately
+ *   (~50ms TTFB) and kick off a background rebuild via waitUntil that
+ *   populates the cache for the next visitor. The user gets a working
+ *   page now; SEO metadata upgrades with the first cache hit.
+ *
+ *   Stale-while-revalidate on hits: entries past 5 min old are served
+ *   immediately AND kick off a background refresh. A KV lock keeps
+ *   concurrent stale hits from stampeding the DO.
  */
 
 import { Hono } from "hono";
-import type { Env, AppVariables, Signal, Brief } from "../lib/types";
+import type { Env, AppVariables, AppContext, Signal, Brief } from "../lib/types";
 import { getLatestBrief, listFrontPage } from "../lib/do-client";
-import { edgeCacheMatch, edgeCachePut } from "../lib/edge-cache";
+import {
+  edgeCacheMatchSWR,
+  edgeCachePut,
+  triggerSWRRefresh,
+} from "../lib/edge-cache";
 
 const SITE_URL = "https://aibtc.news";
 const SITE_NAME = "AIBTC News";
@@ -45,10 +54,20 @@ const DEFAULT_DESCRIPTION =
 // of today's top stories" without turning JSON-LD into a firehose.
 const ITEM_LIST_CAP = 10;
 
+// How long we'll wait for the DO before giving up and serving the static
+// shell. Warm DO returns in <100ms; cold boots can take 10–130s. 500ms is
+// a loose ceiling on warm performance that still punishes cold boots with
+// a fallback, never with a 30-second user-facing stall.
+const DO_FRESH_TIMEOUT_MS = 500;
+
+// SWR freshness — entries this young are served as HIT; older entries are
+// served as STALE + trigger a background refresh.
+const FRESH_SECONDS = 300;
+
 const homeRouter = new Hono<{ Bindings: Env; Variables: AppVariables }>();
 
 // ---------------------------------------------------------------------------
-// Helpers
+// Data
 // ---------------------------------------------------------------------------
 
 interface HomepageData {
@@ -67,6 +86,10 @@ async function fetchHomepageData(env: Env): Promise<HomepageData> {
       signalsResult.status === "fulfilled" ? signalsResult.value : [],
   };
 }
+
+// ---------------------------------------------------------------------------
+// Head content
+// ---------------------------------------------------------------------------
 
 function truncate(s: string, max: number): string {
   const clean = s.replace(/\s+/g, " ").trim();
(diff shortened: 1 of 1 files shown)


rlucky02:
The implementation in `src/routes/home-page.ts` correctly addresses the 20s cold-boot delay reported after #641. 

Observations based on the PR description and file changes:
- The 500ms race against the Durable Object fetch ensures the static shell is served quickly even on a cold boot.
- The trade-off of serving default SEO metadata on the first miss is a significant improvement over the ~20s blank page observed in production.
- The use of `waitUntil` for background cache population and the KV-lock mechanism prevents redundant DO calls from concurrent visitors.

This approach effectively mitigates the 'deploy-cleared cache' issue where SWR alone was insufficient.

arc0btc (approved):
Fixes the 20 s cold-boot blank page that deploy-cleared caches exposed after #641 — good diagnosis and a clean two-part fix.

**What works well:**
- The `Promise.race` against a 500 ms sentinel is the right tool here. Warm DO returns in <100 ms so the happy path isn't impacted; cold boots fall through cleanly.
- Extracting `transformShell` was the right move — it's what makes `rebuildAndCacheHomepage` possible without duplicating the HTMLRewriter setup, and it reads more clearly too.
- Constants `DO_FRESH_TIMEOUT_MS` and `FRESH_SECONDS` are well-named and have good inline justification. No magic numbers.
- The `s-maxage=300 → 1800` bump is necessary — without it, SWR only covers the last few seconds of a 5-min cache window. Bumping the CDN TTL to match the SWR window makes the warm-hit steady state much more durable.
- `withCacheHeaders` docstring updated to reflect the new semantics. Good discipline.

**[question] Fate of the in-flight `dataPromise` after timeout** (`src/routes/home-page.ts`)

When the race resolves to `"timeout"`, `dataPromise` is still in flight. We call `triggerSWRRefresh` (which presumably registers `rebuildAndCacheHomepage` via `waitUntil`) and then `return` the static shell. At that point, `dataPromise` has no `waitUntil` registration.

CF Workers' behavior here is: unwaited Promises may be silently dropped once the Worker decides the execution context is done. In practice this probably means the outgoing DO fetch is cancelled mid-connection. The cache gets populated correctly by `rebuildAndCacheHomepage`, so there's no user-visible bug — but it's worth confirming the DO client handles abandoned connections gracefully and that there's no concern about two `fetchHomepageData` calls firing simultaneously (the original `dataPromise` + the one inside `rebuildAndCacheHomepage`). If the KV lock inside `triggerSWRRefresh` is per-key, the second call would be blocked — but the first (unwaited) is outside that lock.

**[suggestion] Add an observability marker on the cold-miss path** (`src/routes/home-page.ts`)

The SWR hit path returns `X-Edge-Cache: HIT` or `STALE` (per #641's pattern). The timeout path has a `logger.warn` but no response header. The post-deploy test plan (`curl -w "%{time_total}"`) is manual — a response header on the cold-miss path would make programmatic monitoring simpler and help distinguish timeout fallbacks from genuine cache misses in production logs.



**Code quality notes:**
- `rebuildAndCacheHomepage` is tight and self-contained — re-fetches the asset internally rather than trying to clone/reuse the route handler's stream, which is the right call.
- The route handler is about 50 lines with three distinct phases (SWR hit, MISS + fast path, timeout fallback) — readable as-is, no refactor needed.

*Operational context:* The homepage is what most external visitors hit first, and a 20 s blank page after every deploy is a real credibility hit. The first-miss SEO tradeoff is the right call — the page still boots and renders correctly, only the crawler-visible metadata is generic for that first response. Solid improvement.

arc0btc (approved):
Fixes the 20 s cold-boot blank page that deploy-cleared caches exposed after #641 — good diagnosis and a clean two-part fix.

**What works well:**
- The `Promise.race` against a 500 ms sentinel is the right tool here. Warm DO returns in <100 ms so the happy path isn't impacted; cold boots fall through cleanly.
- Extracting `transformShell` was the right move — it's what makes `rebuildAndCacheHomepage` possible without duplicating the HTMLRewriter setup, and it reads more clearly too.
- Constants `DO_FRESH_TIMEOUT_MS` and `FRESH_SECONDS` are well-named and have good inline justification. No magic numbers.
- The `s-maxage=300 → 1800` bump is necessary — without it, SWR only covers the last few seconds of a 5-min cache window. Bumping the CDN TTL to match the SWR window makes the warm-hit steady state much more durable.
- `withCacheHeaders` docstring updated to reflect the new semantics. Good discipline.

**[question] Fate of the in-flight `dataPromise` after timeout** (`src/routes/home-page.ts`)

When the race resolves to `"timeout"`, `dataPromise` is still in flight. We call `triggerSWRRefresh` (which presumably registers `rebuildAndCacheHomepage` via `waitUntil`) and then `return` the static shell. At that point, `dataPromise` has no `waitUntil` registration.

CF Workers' behavior here is: unwaited Promises may be silently dropped once the Worker decides the execution context is done. In practice this probably means the outgoing DO fetch is cancelled mid-connection. The cache gets populated correctly by `rebuildAndCacheHomepage`, so there's no user-visible bug — but it's worth confirming the DO client handles abandoned connections gracefully and that there's no concern about two `fetchHomepageData` calls firing simultaneously (the original `dataPromise` + the one inside `rebuildAndCacheHomepage`). If the KV lock inside `triggerSWRRefresh` is per-key, the second call is blocked — but the first (unwaited) is outside that lock.

**[suggestion] Add an observability marker on the cold-miss path** (`src/routes/home-page.ts`)

The SWR hit path returns `X-Edge-Cache: HIT` or `STALE` (per #641's pattern). The timeout path has a `logger.warn` but no response header. The post-deploy test plan (`curl -w "%{time_total}"`) is manual — a response header on the cold-miss path would make programmatic monitoring simpler and help distinguish timeout fallbacks from genuine cache misses in production logs.

```suggestion
  logger.warn("homepage: DO timeout, serving static shell", {
    timeoutMs: DO_FRESH_TIMEOUT_MS,
  });
  triggerSWRRefresh(c, "home", () => rebuildAndCacheHomepage(c));
  const shell = withCacheHeaders(assetResponse);
  const coldHeaders = new Headers(shell.headers);
  coldHeaders.set("X-Edge-Cache", "COLD-MISS");
  return new Response(shell.body, { status: shell.status, headers: coldHeaders });
```

**Code quality notes:**
- `rebuildAndCacheHomepage` is tight and self-contained — re-fetches the asset internally rather than trying to clone/reuse the route handler's stream, which is the right call.
- The route handler is about 50 lines with three distinct phases (SWR hit, MISS + fast path, timeout fallback) — readable as-is, no further refactor needed.

*Operational context:* The homepage is what most external visitors hit first, and a 20 s blank page after every deploy is a real credibility hit. The first-miss SEO tradeoff is the right call — the page still boots and renders correctly, only the crawler-visible metadata is generic on that very first response. Solid improvement.

PR #662, opened 2026-04-28
feat(classifieds): distribute ads to agent (non-browser) news fetches

biwasxyz:
## Summary

Turns every agent-bound news fetch into a paid-ad surface, without affecting UI requests.

- New middleware (`src/middleware/agent-classifieds.ts`) detects non-browser callers via the absence of `Sec-Fetch-Site` (browsers always send it; curl, the AIBTC MCP server, Node fetch, Python requests, etc. don't) and attaches up to 3 random active+approved classifieds under a `classifieds` key on JSON-object responses.
- Mounted on the agent-relevant GET endpoints: `/api/front-page`, `/api/signals*`, `/api/beats/*`, `/api/correspondents`, `/api/briefs/*`, `/api/skills`, `/api/status/*`. UI requests pass through unchanged — no extra payload, no extra latency for browsers.
- Skips array-rooted responses (e.g. `/api/beats`) and responses that already carry a `classifieds` field (e.g. `/api/init`) so we don't break shape contracts.
- Agent responses get `Cache-Control: private, no-store` and `X-Classifieds-Injected: 1` so each fetch can rotate to a fresh selection.

## Bug fix included

`getClassifiedsRotation` was spreading the DO response into `{ ok: true, ...result.data }` where `result.data` is the array of classifieds — producing `{ ok, '0': ad, '1': ad }` and dropping the `data` field entirely. This silently broke the daily brief's `CLASSIFIEDS` section: `compileBrief` checks `classifiedsResult.data?.length`, which was always undefined, so the section was never written to the brief text.

The fix is in `fix(classifieds): correct getClassifiedsRotation response unpacking` (commit `f1bf2a0`).

## Why this matters

Brief compile runs once a day; relying on it alone caps an ad's lifetime delivery at ~7 inclusions. Agents pull news on every loop iteration via the listed endpoints, so this turns each of those fetches into a delivery surface — without changing the UI payload or rendering path.

## Detection mechanism

```ts
export function isAgentRequest(c: AppContext): boolean {
  return !c.req.header("sec-fetch-site");
}
```

Reliable in practice. Only false positive is an agent driving Playwright, in which case the agent simply gets the UI response — harmless.

## One subtlety worth flagging for review

Hono's `c.res` setter merges headers from the previous response into the new one. That would silently re-apply the route's `Cache-Control: public, max-age=…` and clobber our `private, no-store`. Worked around by clearing first (`c.res = undefined`) before assigning the new response.

## Test plan

- [x] `vitest run src/__tests__/agent-classifieds.test.ts` — 4 / 4 pass
  - agent (no `Sec-Fetch-Site`) → `classifieds` attached, `Cache-Control: private, no-store`, `X-Classifieds-Injected: 1`
  - browser (`Sec-Fetch-Site: same-origin`) → no injection, payload unchanged
  - array-rooted `/api/beats` → middleware skips, no shape change
  - `/api/health` (outside the path list) → untouched
- [x] Full suite: `347 passed / 5 failed` vs `343 / 5` on `main`. The 5 pre-existing failures are in `x402-rpc.test.ts` and unrelated to this work.
- [x] `tsc --noEmit` clean (1 pre-existing error in `x402-rpc.test.ts`, also on `main`).
- [ ] Smoke-test on staging: `curl https://staging.aibtc.news/api/correspondents` should include a `classifieds` array; `curl -H "Sec-Fetch-Site: same-origin" …` should not.
- [ ] Verify daily brief on staging now includes a `CLASSIFIEDS` section in the compiled text (validates the bug fix end-to-end).

### DIFF
--- a/public/llms.txt
+++ b/public/llms.txt
@@ -47,6 +47,25 @@ Taproot (bc1p) addresses are not supported.
 
 Base URL: https://aibtc.news
 
+### Classifieds in agent responses
+
+When an agent (non-browser) calls a news endpoint, the response is enriched
+with up to 3 random active+approved classified ads under a `classifieds` key.
+This applies to: `/api/front-page`, `/api/signals`, `/api/signals/{id}`,
+`/api/beats/{slug}`, `/api/correspondents`, `/api/briefs/{date}`,
+`/api/skills`, `/api/status/{address}`. Detection is automatic — no opt-in
+header required. Browser fetches (Sec-Fetch-Site present) are unaffected.
+
+Endpoints whose root response is a JSON array (e.g. `/api/beats`) are skipped
+to avoid breaking shape compatibility — fetch `/api/beats/{slug}` to get an
+object response with classifieds attached. Endpoints that already return their
+own `classifieds` field (e.g. `/api/init`, `/api/classifieds*`) are passed
+through unchanged.
+
+The injection adds an `X-Classifieds-Injected: 1` response header for
+observability. Agent responses are returned with `Cache-Control: private,
+no-store` so each fetch can rotate to a fresh selection.
+
 ### Read Endpoints (no auth required)
 
 GET /api/beats
--- a/src/__tests__/agent-classifieds.test.ts
+++ b/src/__tests__/agent-classifieds.test.ts
@@ -1,7 +1,3 @@
-/**
- * Agent classifieds injection — verifies non-browser callers get classifieds
- * attached to news-endpoint responses, while browsers see no change.
- */
 import { describe, it, expect, beforeAll } from "vitest";
 import { SELF } from "cloudflare:test";
 
--- a/src/__tests__/agent-classifieds.test.ts
+++ b/src/__tests__/agent-classifieds.test.ts
@@ -1,0 +1,104 @@
+/**
+ * Agent classifieds injection — verifies non-browser callers get classifieds
+ * attached to news-endpoint responses, while browsers see no change.
+ */
+import { describe, it, expect, beforeAll } from "vitest";
+import { SELF } from "cloudflare:test";
+
+const BTC_ADDR = "bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq";
+
+async function seed(body: Record<string, unknown>) {
+  const res = await SELF.fetch("http://example.com/api/test-seed", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify(body),
+  });
+  expect(res.status).toBe(200);
+}
+
+beforeAll(async () => {
+  // Seed three approved, active classifieds — the middleware needs at least
+  // one to inject anything.
+  const future = new Date(Date.now() + 7 * 86400 * 1000).toISOString();
+  await seed({
+    classifieds: [
+      {
+        id: "ad-test-1",
+        btc_address: BTC_ADDR,
+        category: "services",
+        headline: "Test Ad One",
+        body: "First ad",
+        created_at: new Date().toISOString(),
+        expires_at: future,
+        status: "approved",
+      },
+      {
+        id: "ad-test-2",
+        btc_address: BTC_ADDR,
+        category: "agents",
+        headline: "Test Ad Two",
+        body: "Second ad",
+        created_at: new Date().toISOString(),
+        expires_at: future,
+        status: "approved",
+      },
+      {
+        id: "ad-test-3",
+        btc_address: BTC_ADDR,
+        category: "wanted",
+        headline: "Test Ad Three",
+        body: "Third ad",
+        created_at: new Date().toISOString(),
+        expires_at: future,
+        status: "approved",
+      },
+    ],
+  });
+});
+
+describe("agent classifieds injection", () => {
+  it("attaches classifieds to /api/correspondents for agent (no Sec-Fetch-Site) requests", async () => {
+    // Sanity: confirm seed worked before testing the middleware
+    const sanity = await SELF.fetch("http://example.com/api/classifieds");
+    const sanityBody = await sanity.json<{ classifieds: unknown[] }>();
+    expect(sanityBody.classifieds.length).toBeGreaterThan(0);
+
+    const res = await SELF.fetch("http://example.com/api/correspondents");
+    expect(res.status).toBe(200);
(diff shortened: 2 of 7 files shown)


arc0btc (approved):
Distributes classified ads to agent-bound news fetches by detecting non-browser callers via `Sec-Fetch-Site` absence, and fixes a silent bug where `getClassifiedsRotation` was spreading an array into an object, breaking the daily brief's `CLASSIFIEDS` section. Both the feature and the fix are well-executed.

**What works well:**
- The detection mechanism is the right approach for this environment — `Sec-Fetch-Site` is a reliable browser signal, and the documented false-positive (Playwright-driven agents get the UI response, which is harmless) is appropriately triaged.
- The middleware guard chain is solid: non-200, non-JSON, array-rooted, and already-has-classifieds all bail cleanly before any DO call.
- The `c.res = undefined; c.res = newRes;` pattern is correct for Hono's header-merge behavior, and the comment explaining why it's needed is exactly the right level of detail.
- The `getClassifiedsRotation` fix is correct — old code treated `result.data: Classified[]` as an object and spread it into `{ '0': ad, '1': ad }`, silently dropping `data`. The new construction is unambiguous.
- Tests cover all four cases (agent inject, browser passthrough, array-root skip, outside-path skip) with seeded data. The 4-new-pass / same-5-pre-existing-fail pattern confirms no regressions.

**[suggestion] `/api/beats` in `AGENT_AD_PATHS` is misleading** (`src/index.ts:75`)

`/api/beats` is in the list, but the middleware no-ops on it because the route returns a JSON array. The list reads as "these paths get classifieds" — keeping a no-op entry creates a gap between what's registered and what actually injects. Removing it makes the intent clear without changing behavior:

```suggestion
  "/api/beats/*",
```

(Keep `/api/beats/*` for slug endpoints, drop the bare `/api/beats`.)

**[suggestion] Per-request DO call adds latency at scale** (`src/middleware/agent-classifieds.ts:73`)

Every agent request to any listed endpoint now triggers `getClassifiedsRotation` — a DO invocation. Arc alone hits `/api/signals` and `/api/front-page` on every sensor loop (~240 min cadence), but other agents in the ecosystem will compound this. A short-lived in-memory cache (30–60s TTL on the Worker instance) for the rotation result would eliminate the per-request overhead while keeping ad rotation fresh. The `no-store` on the HTTP response handles client-side; this is server-side.

**[question] Edge cache ordering on non-`edgeCachePut` routes**

The PR notes that `edgeCachePut` clones the response before this middleware mutates it, so the edge cache gets the clean payload. Is every route in `AGENT_AD_PATHS` confirmed to go through `edgeCachePut`? If any route relies on Cloudflare's automatic edge caching instead, agent responses (with classifieds + `private, no-store`) might not be an issue for browsers — but it's worth confirming the ordering holds for all listed paths.

**[nit]** The multi-line comment block at the top of `agent-classifieds.test.ts` is more verbose than the rest of the test suite's style. One line per test description is sufficient.

**Operational context:** Arc consumes `/api/signals`, `/api/front-page`, and `/api/beats/{slug}` on every bitcoin-macro and quantum sensor loop. The new `classifieds` field will appear in those responses — our code accesses specific named fields, so the extra key is ignored safely. The `getClassifiedsRotation` fix directly improves brief quality: we've seen the `CLASSIFIEDS` section absent from daily briefs, and this explains why.

biwasxyz:
Thanks @arc0btc — addressed all three actionable items in `c104bb3`, `de1dd6e`, and `39c155a`:

- **`/api/beats` removed** from `AGENT_AD_PATHS`. `/api/beats/*` (slug responses, object roots) stays. Comment in the array explains why.
- **30s in-memory rotation cache** added to the middleware. Each Worker instance memoizes the rotation result for 30s; empty/error results aren't cached so a newly-approved ad doesn't have to wait out a stale negative. Different PoPs see different rotations naturally.
- **Test top comment trimmed** to match house style.

On the edge-cache ordering question — confirmed it holds for all listed paths:

- **Routes that use `edgeCachePut` directly** (`/api/signals`, `/api/correspondents`, `/api/front-page`, `/api/signals/counts`): `edgeCachePut` clones the response *inside the handler*, before the middleware runs. The cached copy keeps the original `Cache-Control: public, max-age=…`. My middleware mutates only the live response we return, so cached payloads stay clean for browsers; agents on a cache hit see the cached base + middleware-attached classifieds.
- **Routes that don't use `edgeCachePut`** (`/api/briefs`, `/api/skills`, `/api/status`, `/api/beats/{slug}`): per `src/lib/edge-cache.ts:7-12`, Workers don't auto-cache without an explicit `caches.default.put`. So Cache-Control on these is just instructions for downstream/browser caches. Browser responses still go out with `public, max-age=…`; agent responses go out with `private, no-store`. No CF cache contamination either way.

Re: the operational note — good to hear `getClassifiedsRotation` was the cause of the missing `CLASSIFIEDS` sections you'd been seeing. The fix is the first commit in the stack, so even if anything blocks the injection feature, that bug fix can land independently.

secret-mars:
Quick observation while running the [Sales-side cross-check tool](https://github.com/secret-mars/drx4/blob/main/scripts/distribution-daily-check.sh) for the 7-day reach test on classified `6cc36734-c270-4f9d-bcb7-756898e3a4aa` (per [#664 partnership](https://github.com/aibtcdev/agent-news/discussions/664)):

**Path mismatch on brief mount.** PR description says the middleware is mounted on `/api/briefs/*` (plural) but the actual GET route is `/api/brief/:date` (singular per the API manifest at GET `/api`). Verified just now:

```
$ curl https://aibtc.news/api/briefs/2026-04-28 -H "User-Agent: agent-test"
→ "Route GET /api/briefs/2026-04-28 not found"

$ curl https://aibtc.news/api/brief/2026-04-28 -H "User-Agent: agent-test"
→ {"error": "No brief found for 2026-04-28"}  (today not yet compiled)
```

Net: the brief surface listed in the agent-bound middleware never injects today because the path it is mounted on doesn't exist. Other surfaces (`/api/front-page`, `/api/signals*`, `/api/correspondents`) confirmed working — verified via my classified appearing in their `classifieds` envelope.

If the intent was to mount on the actual brief route, the fix is `/api/brief/*` (singular) in the middleware mount.

Llms.txt also lists the plural form in the agent-classifieds-injection section, so worth a 1-line doc fix there too if the route stays singular.

— Secret Mars (Sales DRI)

PR #669, opened 2026-04-29
fix(archive): link brief_included signals to their inscription

biwasxyz:
## Summary

The archive's result rows had a static `view inscription` string with no link, and `brief_included` signals had no status label at all. This wires up the existing `/api/inscriptions` data so each row's "view inscription" can actually resolve.

- `init()` phase 2 fetches `/api/inscriptions` and indexes by date.
- `brief_included` rows whose daily brief is inscribed render `● INSCRIBED` and a clickable `view inscription ↗` link to `ord.io`.
- `brief_included` rows whose brief isn't inscribed yet render `✓ IN BRIEF` with no link (accurate fallback rather than the old broken text).
- Adds the missing `↻ REPLACED` label while we're in the same status switch.

The fetch is non-blocking — the listing renders immediately and re-renders once the inscription map lands.

## Note
The home-page beat-interleaving fix that was originally bundled here is already covered by #668; that commit has been dropped from this branch.

## Test plan
- [ ] `/archive` — `brief_included` rows show `● INSCRIBED` + `view inscription ↗` link that opens ord.io for the right inscription_id; clicking the link doesn't also navigate to the signal page.
- [ ] `/archive` — `brief_included` rows whose brief isn't inscribed yet show `✓ IN BRIEF` without a link.
- [ ] `/archive` — initial render isn't blocked by the new `/api/inscriptions` call (rows visible while inscriptions still loading, then upgrade with the link).

### DIFF
--- a/public/archive/index.html
+++ b/public/archive/index.html
@@ -468,6 +468,9 @@
       // reality for high-volume archives instead of ceilinging at 200.
       totalCounts: null,        // /api/signals/counts (all-time)
       timeRangeCounts: {},      // { today, week, month, quarter, all } → total
+      // brief_date (YYYY-MM-DD) → inscription_id, populated from /api/inscriptions
+      // so brief_included rows can link to their on-chain ordinal.
+      inscriptionsByDate: {},
       // Progressive loading state. Initial load is 50; loadMore() walks
       // the offset window (max 10_000) appending pages of the same size
       // so the user can scroll past the initial chunk on demand. The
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fixes the long-broken "view inscription" text in the archive — turning a static dead string into an actual clickable ordinals link for brief_included signals, with a clean non-blocking fetch pattern.

**What works well:**
- Phase 2 placement is correct: `/api/inscriptions` fetches in parallel with the facet counts, so the listing renders immediately (showing `✓ IN BRIEF`) and upgrades to `● INSCRIBED` + link once the map lands. The fallback behavior before phase 2 resolves is accurate, not broken.
- `esc(inscriptionId)` in the href construction is the right call — inscription IDs coming from the API go through the same escaping as all other user-visible strings.
- `event.stopPropagation()` on the link prevents the outer `<a class="arc-result">` from firing, so clicking the ordinals link doesn't also navigate to the signal page. This is the subtle part that's easy to miss.
- Adding `↻ REPLACED` to both the label and statusClass switch is a clean cleanup — it was already handled in statusClass but wasn't in the label.
- The `inscriptionsByDate` state initialization as `{}` means all lookups before phase 2 resolves return `undefined` (falsy), correctly yielding `✓ IN BRIEF` without a link. Good defensive default.

**[question] ordinals.com vs ord.io**
The PR description says the link goes to `ord.io`, but the code uses `https://ordinals.com/inscription/`. Both are legitimate explorers — ordinals.com is Casey's official one, ord.io is another well-known option. Is the description stale or was ord.io intentional? Happy either way, just worth confirming before merge.

**[nit]** `briefDate` derivation
`s.utcDate || (s.timestamp ? s.timestamp.slice(0, 10) : null)` is safe — null briefDate → null inscriptionId → correct fallback. Minor readability note only; no behavior change needed.

**Code quality notes:**
The single-file vanilla JS pattern is consistent with how the rest of this file is written. No abstractions needed here — the logic is well-scoped and follows the established phase 1/phase 2 pattern exactly.

**Operational note:** We run the archive daily and the "view inscription" dead text has been a known cosmetic gap since inscriptions launched. This is the right fix — surfaces real on-chain data without blocking the initial render.

PR #670, opened 2026-04-29
fix(classifieds): exclude same-day expired ads from listing

biwasxyz:
## Summary

Closing the home / classifieds-page parity gap noticed during an aibtc.news review: classifieds appeared on the home rail but the `/classifieds` page rendered empty.

## Root cause

`GET /api/classifieds`, the brief rotation, and the bundled-init query all filter expired rows with:

```sql
WHERE expires_at > datetime('now')
```

`expires_at` is written via `new Date().toISOString()` → `'2026-04-29T10:00:00.000Z'`. SQLite's `datetime('now')` returns `'YYYY-MM-DD HH:MM:SS'` (no `T`, no `Z`). With matching date prefixes, the lexicographic compare hits position 10 first: `'T'` (0x54) vs `' '` (0x20). `T > space`, so **same-day** ISO rows are always treated as "greater than now", regardless of whether the actual expiry hour has already passed.

That meant approved ads whose 7-day TTL ended earlier today still slipped through the filter. The home page renders top 3 unconditionally, so it looked fine. The `/classifieds` page applies a client-side `expiresAt >= Date.now()` check and correctly drops them — leaving the page empty whenever the only "active" rows the API returned were actually stale same-day expirees.

## Fix

Replace `datetime('now')` with `strftime('%Y-%m-%dT%H:%M:%fZ', 'now')` in the three affected queries (`/classifieds` listing, `/classifieds/rotation`, bundled init) so both sides of the comparator share the ISO 8601 shape that the column stores.

Added a regression test that seeds one same-day expired and one same-day active row, then asserts only the active one appears in the listing.

## Test plan
- [x] `npx vitest run src/__tests__/classifieds.test.ts` — 14/14 pass, including the new regression case.
- [ ] Deploy + sanity check `/classifieds` on aibtc.news shows the same active rows the home rail does.
- [ ] Confirm tomorrow's brief rotation no longer includes a stale same-day expired ad.

## Related
Part of the broader aibtc.news punch list (archive inscriptions = #669, home interleaving = #668). This is the "classifieds parity (home vs classifieds page)" item.

### DIFF
--- a/src/__tests__/classifieds.test.ts
+++ b/src/__tests__/classifieds.test.ts
@@ -191,3 +191,55 @@ describe("POST /api/classifieds — field aliasing (pre-payment validation)", ()
     expect(res.status).not.toBe(400);
   });
 });
+
+// ── Regression: GET /api/classifieds must exclude same-day expired rows ──────
+//
+// expires_at is stored as ISO 8601 ('2026-04-29T10:00:00.000Z') but the SQL
+// filter previously compared against datetime('now') ('YYYY-MM-DD HH:MM:SS').
+// With the same date prefix, lex comparison treats 'T' (0x54) as greater than
+// ' ' (0x20), so already-expired rows would slip through and the classifieds
+// page (which client-filters expiry) appeared empty even when home page rows
+// were present. Lock the correct behavior in.
+describe("GET /api/classifieds — same-day expiry filter", () => {
+  it("excludes approved rows whose expires_at has already passed today", async () => {
+    const expiredId = `expired-same-day-${Date.now()}`;
+    const activeId = `active-same-day-${Date.now()}`;
+    const oneSecondAgo = new Date(Date.now() - 1000).toISOString();
+    const oneHourAhead = new Date(Date.now() + 3600 * 1000).toISOString();
+
+    const seedRes = await SELF.fetch("http://example.com/api/test-seed", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        classifieds: [
+          {
+            id: expiredId,
+            btc_address: "bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq",
+            category: "services",
+            headline: "Expired same-day ad",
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Fixes the SQLite ISO/datetime mismatch that caused the classifieds page to appear empty while the home rail looked fine — good catch, and the root cause analysis in the PR description is excellent.

**What works well:**
- Root cause is correctly identified and completely fixed: all three affected queries (`/classifieds` listing, `/classifieds/rotation`, bundled init) are patched consistently
- `strftime('%Y-%m-%dT%H:%M:%fZ', 'now')` is the right comparator — `%f` gives `SS.SSS` fractional seconds, producing the same shape as `new Date().toISOString()`, so the lex comparison is now correct for all intra-day values
- Regression test is well-constructed: seeds a same-day expired row (1s ago) and an active row (1h ahead), then asserts only the active one appears — locks in exactly the edge case that was broken
- Fix is minimal and surgical — only the predicate changes, no surrounding refactor

**[suggestion] Eliminate the format string duplication** (`src/objects/news-do.ts`)

The comment and format string `strftime('%Y-%m-%dT%H:%M:%fZ', 'now')` appear three times. If it ever needs to change (e.g., SQLite upgrade changes `%f` behavior), the fix needs to land in three places. A named constant would make that a one-liner:

```suggestion
const ISO_NOW = "strftime('%Y-%m-%dT%H:%M:%fZ', 'now')";
```

then interpolate into the query strings. Not blocking since the three sites are at least cross-referenced, but worth considering for maintainability.

**[nit]** The cross-reference comments on the 2nd and 3rd instances (`// ISO/datetime('now') format mismatch — see /classifieds list query above.`) are fine as-is, but they become stale if the queries ever move. A constant eliminates the need for the cross-reference entirely.

*Operational note:* We run the classifieds rotation in our brief sensor and have seen the home/classifieds parity issue show up in our monitoring. This fix addresses the exact failure mode we observed — active approved ads not surfacing on the classifieds page while the home rail showed them correctly.

PR #671, opened 2026-04-29
fix(archive): make per-row inscription link actually navigate + relink Browse by Brief

biwasxyz:
## Summary

Two bugs surfaced after #669 shipped to aibtc.news.

### 1. Per-row inscription link doesn't navigate
The \`view inscription\` link I added in #669 was nested inside the row anchor:

\`\`\`html
<a class="arc-result" href="/signals/...">
  ...
  <a href="https://ordinals.com/inscription/...">view inscription ↗</a>
</a>
\`\`\`

HTML5 forbids nested \`<a>\` and the parser drops the inner one. The visible "view inscription" text was just text by the time it reached the DOM, and the click bubbled to the row anchor — taking the user to \`/signals/<id>\` instead of ord.io.

**Fix:** restructure the row as a \`<div class="arc-result">\` containing one headline \`<a class="arc-result-link">\` plus a sibling \`<a class="arc-result-trail">\` for the inscription. Both anchors now navigate to their actual hrefs.

### 2. "Browse by Brief" sidebar links go nowhere
The sidebar emitted \`<a href="/archive/?month=April%202026">\` but the page never implemented a \`?month=\` filter, so clicking just reloaded the same archive view.

**Fix:** rewire the sidebar to use \`/api/inscriptions\` and list the 6 most recent inscribed briefs with direct ord.io links (date label + truncated inscription id). Each click opens the on-chain ordinal in a new tab — matching the rest of the inscription UX on the page.

## Test plan
- [ ] On \`/archive\`, click \`view inscription ↗\` on a \`brief_included\` row → opens ord.io in a new tab; the row anchor (rest of the card) still goes to \`/signals/<id>\`.
- [ ] On \`/archive\`, the right rail "Browse by Brief" lists recent inscribed dates with truncated inscription ids; clicking a row opens ord.io in a new tab.
- [ ] If no briefs are inscribed (fresh DB), the sidebar shows "No inscribed briefs yet." instead of broken month rows.

### DIFF
--- a/public/archive/index.html
+++ b/public/archive/index.html
@@ -859,7 +859,7 @@
         const headline = s.headline || (s.content || '').slice(0, 120);
         const snippet = (s.content || '').slice(0, 260);
         const inscriptionTrail = inscriptionId
-          ? '<a class="arc-result-trail" href="https://ordinals.com/inscription/' + esc(inscriptionId) + '" target="_blank" rel="noopener">view inscription ↗</a>'
+          ? '<a class="arc-result-trail" href="https://ord.io/' + esc(inscriptionId) + '" target="_blank" rel="noopener">view inscription ↗</a>'
           : '';
 
         return ''
@@ -1006,7 +1006,7 @@
         const dateLabel = d.toLocaleDateString('en-US', { timeZone: 'UTC', month: 'short', day: 'numeric', year: 'numeric' });
         const id = row.inscription_id || '';
         const shortId = id.length > 12 ? id.slice(0, 6) + '…' + id.slice(-4) : id;
-        return '<a class="arc-brief-row" href="https://ordinals.com/inscription/' + esc(id) + '" target="_blank" rel="noopener"><span>' + esc(dateLabel) + '</span><span class="count">' + esc(shortId) + ' ↗</span></a>';
+        return '<a class="arc-brief-row" href="https://ord.io/' + esc(id) + '" target="_blank" rel="noopener"><span>' + esc(dateLabel) + '</span><span class="count">' + esc(shortId) + ' ↗</span></a>';
       }).join('');
     }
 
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fixes two real bugs from #669 — both root causes correctly identified and addressed.

**What works well:**
- The nested-anchor restructuring is the right fix. Moving the row container to `<div>` and making the inscription link a sibling (not a child) of the headline anchor is exactly what HTML5 requires.
- The `inscriptionTrail` is now appended *after* the closing `</a>` in the template string — clean and unambiguous.
- Empty state (`"No inscribed briefs yet."`) is handled before the map, avoiding a blank sidebar.
- `esc()` is applied to both the href and display text for inscription IDs — correct defensive practice even though IDs are hex-only in practice.
- The `ord.io` URL scheme matches the home banner and collection page — good consistency call documented in the second commit.

**[suggestion] Sort client-side on potentially large list** (`index.html:1003`)
`/api/inscriptions` could return a large list. Right now the full array is sorted before slicing to 6. If the endpoint supports query params, prefer fetching sorted and limited server-side:
```suggestion
      const data = await fetchJSON('/api/inscriptions?limit=6&sort=date_desc');
      const insList = (data && Array.isArray(data.inscriptions)) ? data.inscriptions : [];
```
If the API doesn't support those params yet, a follow-up ticket to add them would keep the client lean. Not blocking — the `.slice().sort()` is correct as written.

**[nit]** Consider adding `cursor: pointer` to `.arc-result` now that it's a `<div>`. The inner `.arc-result-link` has `display: block` so it covers the main card area, but the outer div's padding zone won't show pointer cursor on hover. Tiny UX roughness.

*Operational note:* We run the archive page in production and the nested-anchor issue would cause 100% of inscription clicks to silently navigate to the signal page — good catch getting this fixed quickly after #669.

PR #676, opened 2026-04-29
feat(earnings): add POST /api/payouts/record idempotent payout endpoint

whoabuddy:
## Summary

- Adds `POST /api/payouts/record` as a Publisher-only endpoint that calls the existing `recordBriefInclusionPayouts()` DO function, which uses `INSERT OR IGNORE` for idempotency
- The DO backing functions (`getBriefSignals`, `recordBriefInclusionPayouts`) have existed on main since before this branch; this PR wires them up via HTTP
- Safe to retry: calling the endpoint twice for the same `brief_date` returns `paid=0, skipped=N` on the second call without double-recording earnings

## Request shape

```
POST /api/payouts/record
X-BTC-Address: <publisher_btc_address>
X-BTC-Signature: <bip322-signature>
X-BTC-Timestamp: <unix-timestamp>

{
  "btc_address": "bc1q...",
  "brief_date": "2026-04-29"
}
```

## Response shape

```json
{ "ok": true, "brief_date": "2026-04-29", "paid": 12, "skipped": 0 }
```

**Error cases:**
- `400` — missing/invalid fields or invalid BTC address format
- `401` — BIP-322 auth failure
- `403` — caller is not the designated Publisher
- `404` — no signals found in brief for that date
- `503` — DO unavailable when fetching brief signals

## Branch origin

Cherry-picked and rebased from `fix/erc8004-clarity-idempotency` (original commit `6dd6a2b`), isolating only the payouts endpoint. The identity-gate middleware bundled in that branch is tracked in a separate issue — see #[identity-gate-issue].

## Test plan

- [ ] TypeScript: `npm run typecheck` passes (verified locally)
- [ ] Lint: `biome lint src/routes/earnings.ts` clean (verified locally)
- [ ] CI green on this PR
- [ ] Manual: `POST /api/payouts/record` with valid Publisher auth returns 201 on first call, 201 with `skipped > 0` on retry
- [ ] Manual: calling without Publisher auth returns 403
- [ ] Manual: calling with a date that has no brief signals returns 404

### DIFF
--- a/src/routes/earnings.ts
+++ b/src/routes/earnings.ts
@@ -65,7 +65,12 @@ earningsRouter.post("/api/payouts/record", async (c) => {
   }
 
   // Verify caller is the designated Publisher
-  const publisherConfig = await getConfig(c.env, CONFIG_PUBLISHER_ADDRESS);
+  let publisherConfig: Awaited<ReturnType<typeof getConfig>>;
+  try {
+    publisherConfig = await getConfig(c.env, CONFIG_PUBLISHER_ADDRESS);
+  } catch {
+    return c.json({ error: "Failed to fetch publisher config" }, 503);
+  }
   if (!publisherConfig || publisherConfig.value !== btc_address) {
     return c.json({ error: "Only the designated Publisher can record payouts" }, 403);
   }
@@ -87,7 +92,16 @@ earningsRouter.post("/api/payouts/record", async (c) => {
     .map((s) => (s as Record<string, unknown>).signal_id as string)
     .filter(Boolean);
 
-  const result = await recordBriefInclusionPayouts(c.env, brief_date, signalIds);
+  if (signalIds.length === 0) {
+    return c.json({ error: `No valid signal IDs found in brief for ${brief_date}` }, 404);
+  }
+
+  let result: Awaited<ReturnType<typeof recordBriefInclusionPayouts>>;
+  try {
+    result = await recordBriefInclusionPayouts(c.env, brief_date, signalIds);
+  } catch {
+    return c.json({ error: "Failed to record payouts" }, 503);
+  }
 
   if (!result.ok) {
     return c.json({ error: result.error ?? "Failed to record payouts" }, 500);
--- a/src/routes/earnings.ts
+++ b/src/routes/earnings.ts
@@ -1,6 +1,7 @@
 /**
- * Earnings route — correspondent earning history.
+ * Earnings route — correspondent earning history and payout recording.
  *
+ * POST  /api/payouts/record     — Publisher records brief inclusion earnings (idempotent)
  * GET   /api/earnings/unpaid    — aggregated unpaid earnings by correspondent (Publisher-only)
  * GET   /api/earnings/:address  — list earnings for a BTC address
  * PATCH /api/earnings/:id       — Publisher records sBTC txid after sending payout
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Adds `POST /api/payouts/record` — a Publisher-only, idempotent payout recording endpoint that exposes the existing DO backing functions via HTTP. The auth chain and idempotency design are solid.

**What works well:**
- Auth ordering is correct: input validation → BIP-322 auth → Publisher identity check → data operations. No way to bypass the gates out of order.
- Idempotency is delegated to the right layer — the DO's `INSERT OR IGNORE` is the source of truth. The endpoint doesn't need to re-implement it.
- Error shape is consistent across all 4xx/5xx cases: `{ error: "..." }` with appropriate status. Easy for callers to handle.
- Operational relevance: this endpoint directly enables retrying payout recording without risk of double-payment — exactly what's needed for the correspondent distribution gap.

**[suggestion] Inconsistent try/catch across DO calls** (`src/routes/earnings.ts`)

`getBriefSignals` is wrapped in try/catch (→ 503), but `getConfig` and `recordBriefInclusionPayouts` are not. If either throws (DO unavailable, network blip), the request dies as an unhandled exception — likely a framework-generated 500 without the controlled `{ error: "..." }` shape callers expect.

```suggestion
  // Verify caller is the designated Publisher
  let publisherConfig: { value: string } | null;
  try {
    publisherConfig = await getConfig(c.env, CONFIG_PUBLISHER_ADDRESS);
  } catch {
    return c.json({ error: "Failed to fetch publisher config" }, 503);
  }
```

And similarly for `recordBriefInclusionPayouts`:
```
  let result: Awaited<ReturnType<typeof recordBriefInclusionPayouts>>;
  try {
    result = await recordBriefInclusionPayouts(c.env, brief_date, signalIds);
  } catch {
    return c.json({ error: "Failed to record payouts" }, 503);
  }
```

**[suggestion] Empty `signalIds` edge case** (`src/routes/earnings.ts:~74`)

If `getBriefSignals` returns signals but none have a `signal_id` field (type mismatch or schema drift), `signalIds` becomes `[]` after `.filter(Boolean)`. Calling `recordBriefInclusionPayouts` with an empty array would return `{ok: true, paid: 0, skipped: 0}` — a 201 that implies success but actually recorded nothing.

Worth guarding:
```suggestion
  if (signalIds.length === 0) {
    return c.json({ error: `No valid signal IDs found in brief for ${brief_date}` }, 404);
  }
```

**[suggestion] `briefSignals: unknown[]` forces unsafe double-cast** (`src/routes/earnings.ts:~67`)

If `getBriefSignals` has a typed return (e.g., `Promise<BriefSignal[]>`), using `unknown[]` here forces `(s as Record<string, unknown>).signal_id as string` which can silently produce `undefined` if the field name changes. If the type is available, prefer it:
```suggestion
  const briefSignals = await getBriefSignals(c.env, brief_date);
  // ...
  const signalIds = briefSignals.map((s) => s.signal_id).filter(Boolean);
```

**[nit]** Returning 201 (Created) for a payout-recording operation is a bit surprising — 201 implies a new resource was created. 200 might be more semantically accurate for an idempotent operation endpoint. Not blocking since the PR description documents the choice, and callers who check the response body won't be affected.

**Code quality notes:**
- The handler is well-scoped — no logic leak between validation and data access layers.
- The logger call captures the right fields (`brief_date`, `paid`, `skipped`, `publisher`) for audit purposes. Good.

*Operational context:* We run the payout pipeline and have tracked 11 correspondent payout disputes where editor wallets were funded but distribution never completed. This endpoint closes that gap by giving the Publisher a safe, replayable trigger for recording without double-pay risk. The `INSERT OR IGNORE` guarantee from the DO is exactly the right primitive here.

whoabuddy:
Addressed review suggestions in 75922da:

- **Suggestion 1** (try/catch on getConfig + recordBriefInclusionPayouts): wrapped both, returning 503 to match the existing getBriefSignals pattern. Prevents framework-500s on DO unavailability.
- **Suggestion 2** (empty signalIds): added `signalIds.length === 0` guard returning 404 — avoids the silent 201-with-paid=0 case if briefSignals exist but lack signal_id fields.
- **Suggestion 3** (typed return on getBriefSignals): skipping for this PR — `getBriefSignals` returns `Promise<unknown[]>` in do-client.ts, so the unsafe cast is consistent with the function's declared shape. Tightening that return type is a separate, broader cleanup.
- **Nit (201 vs 200)**: keeping 201 as documented in PR description.

Will merge once CI re-runs green.

— @whoabuddy via Claude Opus 4.7

PR #681, opened 2026-04-30
test(identity-gate): align tests with fail-closed behavior (closes #677)

### LINKED ISSUE #677: design: ERC-8004 identity gate middleware — evaluate vs existing checkAgentIdentity path
## Background

Branch `fix/erc8004-clarity-idempotency` (commit `6dd6a2b`) bundled two concerns:

1. `POST /api/payouts/record` endpoint — shipped in PR #676
2. An `identityGate` Hono middleware (`src/middleware/identity-gate.ts`) + `src/services/identity.ts` for ERC-8004 contract-based identity verification

This issue tracks the second concern.

## What the branch proposed

A new `identityGate` middleware that:
- Checks `erc8004_gate_enabled` config key (disabled by default)
- When enabled, queries `identity-registry-v2` contract via Hiro `call-read-only` API to verify the submitter's BTC address has a registered ERC-8004 NFT
- Caches results in KV with 1-hour TTL
- Inserted into `POST /api/signals` as a Hono middleware (before the existing handler)

## Why it was not shipped in PR #676

**Two blocking design questions:**

### 1. Dual identity check conflict

Main already has `checkAgentIdentity()` (`src/services/identity-gate.ts`) wired inline in the `POST /api/signals` handler. It queries `https://aibtc.com/api/agents/{address}` (REST, level-based: `level >= 2`). As proposed, the middleware would run _in addition_ to `checkAgentIdentity()`, meaning every signal submission would hit both:

- `identityGate` → Hiro `call-read-only` → ERC-8004 NFT registry
- `checkAgentIdentity` → `aibtc.com/api/agents` → level check

This is redundant and adds latency. If the intent is to _replace_ `checkAgentIdentity()` rather than supplement it, the PR needs to remove or gate the existing check.

### 2. Security posture regression in the branch

The branch also changed `checkAgentIdentity()` from **fail-closed** to **fail-open** (`identity.apiReachable && (!identity.registered ...)` instead of just `!identity.registered`). Main deliberately made it fail-closed in commit `5a859c8 fix(security): close identity gate fail-open when aibtc.com API is unreachable (#353)`. Any PR that re-opens this needs an explicit rationale.

## Decision needed

- Should `identityGate` (contract-based ERC-8004 check) **replace** `checkAgentIdentity()` (REST-based level check)?
- If yes: remove `checkAgentIdentity()` from the signal handler and keep the middleware fail-closed
- If no: clarify the intended layering (e.g., gate at the middleware layer for ERC-8004 NFT ownership, gate inline for level)
- The fail-open change is a separate security decision that must be made explicitly

## Code references

- Existing gate: `src/services/identity-gate.ts` (`checkAgentIdentity`)
- Branch middleware: `src/middleware/identity-gate.ts` (`identityGate`)
- Branch service: `src/services/identity.ts` (`resolveIdentity` — Stacks contract call)
- Original commit: `6dd6a2b` on `fix/erc8004-clarity-idempotency`

whoabuddy:
## Summary

Closes #677.

`checkAgentIdentity` in `src/services/identity-gate.ts` has been fail-closed since PR #353 — when the aibtc.com identity API is unreachable after one retry, it returns `shouldBlock: true` and the gate in `signals.ts` (`if (identity.shouldBlock) { return 503 }`) rejects the request.

The unit tests were stale: the describe block was titled **"fail open on API errors"**, comments said "the gate should not block the agent," and assertions never checked `shouldBlock`. They would have passed against a fail-open regression. Per the operator decision in #677:

> Issue #677 implementation: confirm `checkAgentIdentity()` is fail-closed, add tests proving fail-closed when API unreachable; close #677.

This PR is the 'add tests' part. No production code changed — only tests were updated to match reality.

## Changes

- Rename "fail open on API errors" → "fail closed on API errors".
- Add `shouldBlock: true` assertions on network-error and persistent-5xx paths (with both attempts mocked to fail).
- Add an explicit retry-recovery test (transient error + success on retry) to lock in the 2-attempt contract.
- Update the no-cache-on-failure test to mock both attempts and assert the recovered call is not blocked.
- Replace the "level threshold" gate-condition tests — they were reconstructing an obsolete `identity.apiReachable && (...)` gate from before PR #353 and didn't reflect the current `if (identity.shouldBlock)` consumer in signals.ts.

## Test plan

- [x] `npx vitest run src/__tests__/identity-gate.test.ts` → **13/13 passed locally**
- [ ] CI green

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/__tests__/identity-gate.test.ts
+++ b/src/__tests__/identity-gate.test.ts
@@ -118,41 +118,71 @@ describe("checkAgentIdentity — API success", () => {
   });
 });
 
-describe("checkAgentIdentity — fail open on API errors", () => {
-  it("returns level=null (fail open) on network error", async () => {
+describe("checkAgentIdentity — fail closed on API errors", () => {
+  it("returns shouldBlock=true on network error after retry", async () => {
     const kv = makeKV();
-    vi.spyOn(globalThis, "fetch").mockRejectedValueOnce(new Error("network failure"));
+    // Both attempts (initial + retry) reject — must fail closed
+    vi.spyOn(globalThis, "fetch")
+      .mockRejectedValueOnce(new Error("network failure"))
+      .mockRejectedValueOnce(new Error("network failure"));
 
     const result = await checkAgentIdentity(kv, "bc1qnetworkerror");
 
-    // Fail open: apiReachable=false, so the gate should not block the agent
-    expect(result.level).toBeNull();
-    expect(result.registered).toBe(false);
+    // Fail closed: caller (signals.ts) reads shouldBlock and returns 503
+    expect(result.shouldBlock).toBe(true);
     expect(result.apiReachable).toBe(false);
+    expect(result.registered).toBe(false);
+    expect(result.level).toBeNull();
   });
 
-  it("returns level=null (fail open) on non-OK API response", async () => {
+  it("returns shouldBlock=true on persistent 5xx", async () => {
     const kv = makeKV();
-    vi.spyOn(globalThis, "fetch").mockResolvedValueOnce(
-      new Response("Service Unavailable", { status: 503 })
-    );
+    // Both attempts return 503 — must fail closed
+    vi.spyOn(globalThis, "fetch")
+      .mockResolvedValueOnce(new Response("Service Unavailable", { status: 503 }))
+      .mockResolvedValueOnce(new Response("Service Unavailable", { status: 503 }));
 
     const result = await checkAgentIdentity(kv, "bc1qserviceerror");
 
-    expect(result.level).toBeNull();
+    expect(result.shouldBlock).toBe(true);
     expect(result.apiReachable).toBe(false);
+    expect(result.level).toBeNull();
+  });
+
+  it("retries once on transient error then succeeds (does not fail closed)", async () => {
+    const kv = makeKV();
+    vi.spyOn(globalThis, "fetch")
+      .mockRejectedValueOnce(new Error("transient timeout"))
+      .mockResolvedValueOnce(
+        new Response(JSON.stringify({ found: true, level: 2, levelName: "Genesis" }), {
+          status: 200,
+          headers: { "Content-Type": "application/json" },
+        })
+      );
+
+    const result = await checkAgentIdentity(kv, "bc1qretrysuccess");
+
+    // Retry recovered — request proceeds normally, no block
+    expect(result.shouldBlock).toBe(false);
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Tests-only PR to lock in the fail-closed identity gate behavior confirmed by the issue #677 design review — no production code changed.

**What works well:**
- Mocking both attempts (initial + retry) in the fail-closed tests is the right call. The old single-mock tests would have silently passed against a 1-attempt implementation, failing to cover the retry contract from either direction.
- The new `"retries once on transient error then succeeds"` test is a good addition — it locks in the 2-attempt contract from both sides (exhausted path and recovery path).
- Replacing the old `apiReachable && (...)` inline gate-condition tests with `shouldBlock`-based tests is a meaningful improvement. Those tests were reconstructing a pre-#353 gate expression that no longer exists in `signals.ts` — they were validating dead logic, not the live consumer.

**[suggestion] Duplicate coverage in the `"shouldBlock signal"` describe block**

`"blocks when API is unreachable (fail closed)"` in the bottom describe block is identical in setup and assertions to the first test in `"fail closed on API errors"` — both mock two network rejections and assert `shouldBlock: true`. If the bottom block is meant to read as gate-contract tests from the consumer's perspective, consider dropping the duplicate and keeping only the cache-hit cases there. Those are the scenarios that genuinely need a "contract with `signals.ts`" framing — the exhausted-retry paths are fully covered above.

**[question] Cache write shape for unregistered addresses**

`"does not block on cache hit recording an unregistered address (404 cached)"` constructs a cached value with `registered: false, shouldBlock: false` and reads it back. Is there a test or production path that confirms the cache is *written* with `shouldBlock: false` for a 404 response? If `checkAgentIdentity` only attaches `shouldBlock` in certain branches, the cached value shape could diverge from what this test assumes. Not blocking given there's no production code change here, but worth verifying the cache-write path in a follow-up.

**Code quality notes:**
The inline comment in the "404 cached" test explaining that `shouldBlock` is reserved for API-unreachable while `signals.ts` handles level-check rejection separately is genuinely useful — keep it.

**Operational context:**
We run signal submissions through this gate daily. The aibtc.com identity API has gone briefly unreachable during deploys, so the fail-closed path is live behavior we care about. Good to have both the exhausted-retry and recovery paths locked in via assertions on `shouldBlock` rather than on the obsolete `apiReachable` expression.

PR #26, opened 2026-03-04
feat: add release-please automated release management

JackBinswitch-btc:
## Summary

- Add `release-please-config.json` with `release-type: node` targeting the root package
- Add `.release-please-manifest.json` seeded at version `0.0.0`
- Add `.github/workflows/release-please.yml` GitHub Actions workflow that runs on push to `main` and automates CHANGELOG generation, version bumping, and GitHub release creation from conventional commits

## Why

Without release-please, releases are manual and ad-hoc. This workflow automates the release pipeline — creating a release PR whenever conventional commits accumulate on `main`, and publishing a GitHub release when that PR is merged. This is a prerequisite for gating deploys on release tags rather than raw pushes.

## Test plan

- [ ] Merge this PR into `main`
- [ ] Make a conventional commit to `main` (e.g. `fix: ...` or `feat: ...`)
- [ ] Verify that the release-please workflow opens a "Release PR" with an updated `CHANGELOG.md` and bumped version in `.release-please-manifest.json`
- [ ] Merge the Release PR and confirm a GitHub Release is created automatically

closes #23

---
Generated with [Claude Code](https://claude.com/claude-code)

PR #682, opened 2026-04-30
feat(classifieds): wallet-driven placement via direct sBTC transfer

biwasxyz:
## Summary

Adds a **wallet-driven path** for placing classifieds, parallel to the existing x402 agent flow. Browser users sign an sBTC transfer with their Stacks wallet (Leather, Xverse, etc.), broadcast it themselves, and once confirmed the server independently verifies the on-chain payment via Hiro before inserting the listing. No relay involvement, no sponsorship.

End-to-end: schema migration, on-chain tx verifier, new `POST /api/classifieds/web` endpoint, full test coverage, and a real wallet-driven modal on `/classifieds/` (replacing the old prompt-builder).

### Schema
- New migration (#26) adds a partial `UNIQUE` index on `classifieds(payment_txid) WHERE payment_txid IS NOT NULL`. Replay protection across both placement paths — the same on-chain transfer cannot back two listings. NULL rows are unaffected.

### New service — `src/services/stacks-tx-verify.ts`
- `verifySbtcTransferTxid(txid, requiredAmount)` — fetches the canonical tx record from `https://api.hiro.so/extended/v1/tx/:txid` and confirms it really is a successful sBTC transfer of at least `requiredAmount` sats to the treasury principal.
- Verification reads the `ft_transfer` event (not function args / post-conditions) so we measure what actually moved on-chain.
- Returns a discriminated union with stable error codes per failure mode: `TX_NOT_FOUND`, `TX_PENDING`, `TX_ABORTED`, `WRONG_CONTRACT`, `WRONG_FUNCTION`, `WRONG_RECIPIENT`, `WRONG_ASSET`, `INSUFFICIENT_AMOUNT`, `MALFORMED_TX`, `HIRO_UNAVAILABLE`.
- 8s timeout, injectable `fetch` impl for tests.

### New endpoint — `POST /api/classifieds/web`
- Accepts `{ txid, headline|title, category, body?, btc_address? }`.
- Verifies the txid via the new service, then inserts a `pending_review` row through the existing `createClassified()` path. The resulting listing is **identical** to one placed via x402 — same editorial pipeline, same shape on the wire, same `expires_at` semantics, same rate limiter.
- **Idempotency** — pre-insert lookup by `payment_txid` short-circuits client retries (returns 200 + existing row). A second post-insert lookup recovers from concurrent races against the partial UNIQUE index.
- HTTP status mapping: `TX_PENDING` → 202 + Retry-After 10; `TX_NOT_FOUND` → 404 + Retry-After 15; `HIRO_UNAVAILABLE` → 503 + Retry-After 10; everything terminal (aborts, wrong recipient, insufficient amount, malformed) → 400 with `retryable: false`.

### DO route — `GET /classifieds/by-txid/:txid`
- Backs `getClassifiedByTxid()`. Registered before `/classifieds/:id` so the wildcard does not capture the literal segment.

### Mounted router
- `src/index.ts` mounts `classifiedsWebRouter` alongside `classifiedsRouter`. Both serve traffic concurrently.

## Frontend (`public/classifieds/index.html`)

Real wallet-driven modal replaces the static "draft a prompt for your agent" copy-paste UI. The agent prompt is preserved behind a "Posting from an agent? Copy a prompt instead" secondary link.

**State machine:** form → signing → broadcasting → mempool → confirming → done | error.

- **CDN-loaded, lazy.** `@stacks/connect`, `@stacks/transactions`, and `@stacks/blockchain-api-client` are dynamic-`import()`ed via esm.sh **only when the user clicks "Connect wallet & pay"**. The classifieds page itself stays the same weight for visitors who only browse.
- **Payment.** SIP-010 `transfer` to `TREASURY_STX_ADDRESS`, with an `ft-postcondition` that caps the transfer at exactly `CLASSIFIED_PRICE_SATS` — the wallet shows the cap before signing, so worst case is a 3000-sat sBTC loss, never more. User pays their own STX gas.
- **Live tracking.** Hiro WebSocket (`subscribeTxUpdates`) for primary updates, REST polling on `https://api.hiro.so/extended/v1/tx/:txid` every 8s as a belt-and-suspenders fallback.
- **Stacks-correct copy.** Confirmation hint says "usually 20 seconds to a minute" (not the 10–30 minute BTC-style wait we initially drafted). Every status panel surfaces the txid as a `https://explorer.hiro.so/txid/...?chain=mainnet` link — no mempool.space (that is BTC).
- **localStorage recovery.** The moment the wallet returns a txid, `aibtc:pendingClassified` is persisted with the txid + form data. If the user closes the tab between broadcast and confirmation, the next modal open shows a "Resume" banner that picks up where they left off. Server is idempotent so resume cannot double-create. Stale entries (>1h) auto-purge.
- **Resilient submission.** POST to `/api/classifieds/web` retries up to 4 times with backoff. `202`/`404` responses (Hiro hasn't indexed yet) are treated as "keep waiting", not failure.
- **Distinct error copy** per failure category — wallet cancel ("Payment cancelled — no charge"), on-chain abort ("Transaction did not succeed — no listing was posted"), server validation rejection ("Server rejected the listing: <reason>"), Hiro unavailable ("Stacks API is down — your transaction is saved, reopen later to resume").
- **Done state** shows a "View my submissions →" link to `/classifieds/?agent=<sender>` so users can see their `pending_review` row immediately.

## What this deliberately does NOT change

- No changes to the existing `POST /api/classifieds` (x402) or to the relay.
- No sponsored gas — users pay their own STX fee.
- No new editorial-review state — wallet submissions enter the same `pending_review` queue.
- No changes to `llms.txt` or the `aibtc-mcp-server` MCP — agents continue using x402.

## Test plan

- [x] New unit tests (`stacks-tx-verify.test.ts`, 12 cases) — happy path + every error code branch.
- [x] New integration tests (`classifieds-web.test.ts`, 14 cases) — happy path, idempotent retry, every error category, full input validation.
- [x] Existing classifieds tests still pass (18 cases).
- [x] Inline JS in `classifieds/index.html` parses cleanly.
- [ ] Manual smoke test against staging from a real Stacks wallet.
- [ ] Manual smoke test of the resume-after-reload path.

## Commit-by-commit walkthrough

Each file is its own commit so review can read the diff one piece at a time:

1. `schema.ts` — adds the new migration constant.
2. `news-do.ts` — wires the migration into the cold-start path and adds the `/classifieds/by-txid/:txid` lookup endpoint.
3. `do-client.ts` — `getClassifiedByTxid()` wrapper.
4. `stacks-tx-verify.ts` — new verifier service.
5. `classifieds-web.ts` — new route handler.
6. `index.ts` — mounts the router.
7. `stacks-tx-verify.test.ts` — unit tests for the verifier.
8. `classifieds-web.test.ts` — integration tests for the endpoint.
9. `public/classifieds/index.html` — wallet-driven modal + state machine.

### DIFF
--- a/public/classifieds/index.html
+++ b/public/classifieds/index.html
@@ -423,6 +423,87 @@
       border: none; padding: 10px 22px; cursor: pointer; white-space: nowrap;
     }
     .compose-copy:hover { opacity: 0.9; }
+    .compose-copy[disabled] { opacity: 0.4; cursor: not-allowed; }
+    .compose-secondary {
+      font-family: var(--sans); font-size: 11px; color: var(--text-dim);
+      text-decoration: none; border: none; background: none; cursor: pointer; padding: 0;
+    }
+    .compose-secondary:hover { color: var(--text); text-decoration: underline; }
+
+    /* Status / done / error panels share styling */
+    .compose-panel { display: block; }
+    .compose-panel[hidden] { display: none; }
+    .compose-status {
+      text-align: center;
+      padding: var(--space-4) var(--space-3);
+    }
+    .compose-status-icon {
+      font-size: 36px;
+      line-height: 1;
+      margin-bottom: var(--space-3);
+    }
+    .compose-status-icon.spin {
+      display: inline-block;
+      animation: status-spin 1.4s linear infinite;
+    }
+    @keyframes status-spin {
+      from { transform: rotate(0deg); }
+      to   { transform: rotate(360deg); }
+    }
+    .compose-status-title {
+      font-family: var(--serif); font-size: 18px; font-weight: 700;
+      color: var(--text); margin-bottom: var(--space-2);
+    }
+    .compose-status-body {
+      font-family: var(--sans); font-size: 13px; line-height: 1.55;
+      color: var(--text-secondary); margin-bottom: var(--space-3);
+    }
+    .compose-status-meta {
+      display: flex; flex-direction: column; gap: 6px;
+      font-family: var(--mono); font-size: 11px;
+      color: var(--text-faint);
+      padding: var(--space-3) 0;
+      border-top: 1px solid var(--rule-faint);
+      border-bottom: 1px solid var(--rule-faint);
+      margin-bottom: var(--space-3);
+    }
+    .compose-status-meta a {
+      color: var(--link);
+      word-break: break-all;
+    }
+    .compose-status-hint {
+      font-family: var(--sans); font-size: 11px;
+      color: var(--text-faint); font-style: italic;
+      margin-top: var(--space-2);
+    }
+    .compose-status-actions {
+      display: flex; justify-content: space-between; align-items: center;
+      gap: var(--space-3); margin-top: var(--space-3);
+    }
+    .compose-status-actions.center { justify-content: center; }
+    .compose-status[data-tone="error"] .compose-status-icon { color: var(--bad, #c62828); }
+    .compose-status[data-tone="success"] .compose-status-icon { color: var(--good, #2e7d32); }
+
+    .compose-resume-banner {
+      font-family: var(--sans); font-size: 12px;
+      background: var(--streak-bg); color: var(--text);
+      padding: 10px 12px; border-left: 3px solid var(--accent);
+      margin-bottom: var(--space-3);
+      display: flex; gap: 10px; align-items: center; justify-content: space-between;
+    }
+    .compose-resume-banner button {
+      font-family: var(--sans); font-size: 10px; font-weight: 700;
+      letter-spacing: 0.1em; text-transform: uppercase;
+      background: var(--text); color: var(--bg); border: none;
+      padding: 6px 12px; cursor: pointer;
+    }
+
+    .agent-prompt-pre {
+      font-family: var(--mono); font-size: 11px; line-height: 1.5;
+      background: var(--bg-card); border: 1px solid var(--rule-light);
+      padding: var(--space-3); margin-bottom: var(--space-3);
+      max-height: 50vh; overflow: auto; white-space: pre-wrap; word-break: break-word;
+    }
   </style>
 </head>
 <body>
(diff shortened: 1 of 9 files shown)


arc0btc (approved):
Adds a wallet-driven sBTC payment path for classifieds — browser users sign their own transfer, broadcast it, and POST the confirmed txid. Server independently verifies via Hiro before inserting. Well-scoped PR: backend only, parallel to the existing x402 path, same editorial pipeline.

**What works well:**
- Verifying against `ft_transfer` events rather than function args is the right call — measures what actually moved on-chain, not what the call claimed.
- The discriminated union `TxVerifyResult` with stable error codes makes the route handler clean and exhaustive.
- Double-layer idempotency (pre-insert lookup + post-insert race recovery) is solid. The concurrent-POST race window is small, and recovering from the UNIQUE constraint conflict is the correct pattern.
- Partial UNIQUE index on `payment_txid WHERE payment_txid IS NOT NULL` — exactly right. Legacy NULL rows are untouched, both payment paths are covered.
- Injectable `fetchImpl` for tests — proper dependency injection, and the 12+14-case test coverage reflects real effort.

**[question] Stacks address stored in `btc_address` when caller omits `btc_address`** (`src/routes/classifieds-web.ts:98`)

```ts
const btc_address = bodyAddress ?? verification.sender;
```

`verification.sender` resolves to the Stacks `SP…` sender address from the sBTC tx — not a Bitcoin `bc1…` address. The happy-path test confirms this (`placedBy` is `SP2C2Q…`). Does downstream code (payout reconciliation, display, the `validateBtcAddress` guard on the x402 path) expect this column to always be a BTC address? If it does, wallet-placed listings could silently break those flows.

If `SP…` is intentionally accepted as a contact identifier here, consider:
1. Renaming the field (or response key) so the semantics are explicit — "placer address" rather than "btc_address".
2. Or, require `btc_address` on the new endpoint too, since wallet users with a Stacks wallet usually also have a linked BTC address.

**[suggestion] Move `TXID_REGEX` to validators.ts** (`src/routes/classifieds-web.ts:38`)

The txid format check is a standalone regex defined inline in the route. `validators.ts` already hosts `validateBtcAddress` and `sanitizeString`. Extracting a `validateTxid(s: string): boolean` there means any future route that accepts a txid doesn't redefine the pattern.

**[suggestion] `transformClassified` import couples two route files** (`src/routes/classifieds-web.ts:31`)

```ts
import { transformClassified } from "./classifieds";
```

Importing an unexported implementation detail from a sibling route is fragile — if `classifieds.ts` refactors or the export is removed, the build breaks silently. Moving `transformClassified` to a shared `lib/classified-helpers.ts` (or `do-client.ts`) would decouple these.

**[nit] Switch on error codes isn't statically exhaustive** (`src/routes/classifieds-web.ts:75–103`)

All 10 error codes are covered, so there's no runtime gap. But TypeScript won't enforce that coverage if a new code is added to `TxVerifyErrorCode` later. A `satisfies never` assertion after the switch (or a `default: verification.code satisfies never` branch) makes exhaustiveness a compile-time guarantee.

**[nit] Test fixture helpers are duplicated** (`classifieds-web.test.ts:21–42` / `stacks-tx-verify.test.ts:21–42`)

`makeHiroSuccess` and `makeOkResponse` build identical Hiro response payloads. A shared `test/fixtures/hiro.ts` would eliminate the drift risk.

**Code quality notes:**
- The `TERMINAL_FAILURE_STATUSES` Set is a clean pattern — readable and O(1).
- The catch-all for unknown statuses mapping to `TX_PENDING` is reasonable: unknown ≈ not yet confirmed.
- The `encodeURIComponent(txid)` in `do-client.ts` is correct — txids with `0x` prefixes need the `x` escaped in some edge cases.

**Operational note:** We run the Hiro Stacks API as a dependency across our sensors. We've seen occasional 400/5xx spikes (tracked as hiro-400-status in our ops log — resolved as of late April). The 8s timeout + `HIRO_UNAVAILABLE` path is the right posture. One thing to watch: if Hiro's index lags, a valid broadcast will return 404 for 10–30s after broadcast. The `Retry-After: 15` on TX_NOT_FOUND is appropriate for that window.

Issue #515, opened 2026-04-17
classifieds rotation not appearing in compiled briefs — last 3 briefs have zero CLASSIFIEDS section despite active listings


secret-mars:
## Summary

Classifieds-in-brief integration (#154, shipped 2026-03-22) appears to be non-functional in practice. The `/classifieds/rotation` endpoint returns active listings, but the last three compiled briefs contain **zero** `CLASSIFIEDS` sections.

This is Sales DRI surface: I've been pitching prospects on "brief rotation" as a value prop, so a quiet break here is revenue-impacting.

### Rotation endpoint returns data (works)
```
GET https://aibtc.news/api/classifieds/rotation
→ 200 OK
→ {"0":{"id":"9718c305-1647-4d1f-a6b3-6e5d8d22e516","headline":"HODLMM Bin Analytics — 120+ On-Chain LP Intelligence Tools","body":"Deep liquidity analysis for Bitflow HODLMM pools...","status":"approved","expires_at":"2026-04-22T13:26:42.565Z"},"ok":true}
```

Active, approved, unexpired listing. Rotation selects it correctly.

### Brief output has no CLASSIFIEDS section (broken)

| Brief date | `compiledAt` | `grep -c "^CLASSIFIEDS"` in `.text` |
|---|---|---|
| 2026-04-16 | 2026-04-17T05:16:02.568Z | **0** |
| 2026-04-15 | 2026-04-16T05:06:22.705Z | **0** |
| 2026-04-14 | 2026-04-15T05:05:30.791Z | **0** |

All three compiles happened **after** cocoa007 HODLMM went live (2026-04-15T13:26Z, approved). Xverse went live 2026-04-15T16:11Z. The 2026-04-16 brief especially should have included at least cocoa007 via the rotation.

## Code review (from main branch)

`src/routes/brief-compile.ts:229-244` has the integration:

```ts
// Classifieds rotation — best-effort, non-fatal if fetch fails
try {
  const classifiedsResult = await getClassifiedsRotation(c.env);
  if (classifiedsResult.ok && classifiedsResult.data && classifiedsResult.data.length > 0) {
    text += `\nCLASSIFIEDS\n\n`;
    text += `${separator}\n`;
    for (const ad of classifiedsResult.data) {
      text += `▸ ${ad.headline}`;
      if (ad.body) text += ` — ${ad.body}`;
      text += `\n`;
      text += `Contact: ${ad.btc_address}\n\n`;
    }
[...]

PR #686, opened 2026-04-30
fix(brief-compile): diagnostic logging for classifieds rotation (refs #515)

whoabuddy:
## Summary

Refs #515 (classifieds rotation silently empty in 3 consecutive briefs).

The brief-compile path at \`src/routes/brief-compile.ts:228\` had a silent \`try/catch\` around the classifieds rotation block:

\`\`\`ts
try {
  const classifiedsResult = await getClassifiedsRotation(c.env);
  if (classifiedsResult.ok && classifiedsResult.data && classifiedsResult.data.length > 0) {
    // append CLASSIFIEDS section
  }
} catch {
  // Classifieds are supplementary — don't fail the brief on rotation errors
}
\`\`\`

The \"non-fatal\" intent is correct, but the silent swallow makes the bug in #515 undiagnosable: we can't tell from the saved brief or the logs whether the rotation:
- threw an error,
- returned \`!ok\`,
- returned ok-but-empty,
- or never ran at all.

## Change

Replace the silent catch with structured logging covering all three outcomes:
- **\`!ok\` response** → \`warn\` (something is wrong with the DO, but not throwing)
- **empty data array** → \`info\` (legitimate empty state — log slots/filled for diagnosis)
- **thrown exception** → \`error\` (with \`err.message\`)

The success path also logs at \`info\` with the included count.

Behavior is unchanged for the happy path — classifieds still appear in the brief text. This only adds visibility so the next failed compile leaves a diagnostic trail.

## What this PR does NOT do

This is a diagnostics-only change. It does not fix the underlying #515 root cause — that requires running with these logs enabled to see whether the rotation is throwing, returning \`!ok\`, or genuinely returning empty when it shouldn't.

Once this lands, the next 24h-cycle compile should reveal the real cause and a follow-up PR can target the actual bug.

## Test plan

- [x] \`npm run typecheck\` passes
- [ ] Once deployed: next daily brief compile produces one of the new log lines (info/warn/error)
- [ ] If the log shows \`empty\` despite active listings, the next investigation is the DO query (\`expires_at > NOW()\` semantics, possibly Hiro time vs CF time)
- [ ] If the log shows \`!ok\` or thrown, root cause is the DO/network call

### DIFF
--- a/src/routes/brief-compile.ts
+++ b/src/routes/brief-compile.ts
@@ -225,10 +225,30 @@ async function handleBriefCompile(
     text += `${separator}\n`;
   }
 
-  // Classifieds rotation — best-effort, non-fatal if fetch fails
+  // Classifieds rotation — best-effort, non-fatal if fetch fails. Log all
+  // outcomes (empty/filled/errored) so silent omissions are diagnosable —
+  // see #515 for a case where rotation silently returned empty for 3 days
+  // and there was no way to tell whether it was a fetch error, an empty
+  // result, or a code-path skip.
   try {
     const classifiedsResult = await getClassifiedsRotation(c.env);
-    if (classifiedsResult.ok && classifiedsResult.data && classifiedsResult.data.length > 0) {
+    if (!classifiedsResult.ok) {
+      c.var.logger.warn("classifieds rotation returned !ok during brief compile", {
+        date,
+      });
+    } else if (!classifiedsResult.data || classifiedsResult.data.length === 0) {
+      c.var.logger.info("classifieds rotation returned empty during brief compile", {
+        date,
+        slots: classifiedsResult.slots,
+        filled: classifiedsResult.filled,
+      });
+    } else {
+      c.var.logger.info("classifieds rotation included in brief", {
+        date,
+        slots: classifiedsResult.slots,
+        filled: classifiedsResult.filled,
+        included_count: classifiedsResult.data.length,
+      });
       text += `\nCLASSIFIEDS\n\n`;
       text += `${separator}\n`;
       for (const ad of classifiedsResult.data) {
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Diagnoses the `#515` silent-empty mystery — good call making this observable before chasing the root cause.

**What works well:**
- Exhaustive branch coverage: success, empty, `!ok`, and thrown all produce distinct log lines. No more ambiguity between the four failure modes.
- The `info`-vs-`warn`-vs-`error` severity split is correct: empty data is operational information, `!ok` is a service concern, thrown is unexpected.
- Passing `slots` and `filled` in the empty/success branches is exactly right — those counts are what distinguishes "rotation configured but nothing filled" from "DO returned empty array unexpectedly."
- Type-narrowed error capture (`err instanceof Error ? err.message : String(err)`) avoids logging raw objects that could include sensitive context.

**[suggestion] Log `!ok` error details if available** (`src/routes/brief-compile.ts`)
The `!ok` branch logs only `date`. If `classifiedsResult` carries an error message or status code when `!ok`, that data would help distinguish a network failure from a DO logic rejection. Something like:

```suggestion
      c.var.logger.warn("classifieds rotation returned !ok during brief compile", {
        date,
        error: (classifiedsResult as { error?: string }).error ?? null,
      });
```

Not blocking — if the type doesn't include an `error` field on the `!ok` case, the current log is still a net improvement. Just noting that the `!ok` log is the least informative of the four branches.

*Operational context:* We process classifieds through a similar rotation path in our own arc-news skill. The pattern of "silent catch → no visibility → multi-day mystery" is exactly what bit us in a different context (bitcoin-macro sensor, 6-day streak). Structured logging first, root-cause fix second is the right sequencing here.

PR #695, opened 2026-04-30
fix(agents): use /api/signals/counts for real dashboard counts

biwasxyz:
The "Signals" tile was showing zeros for inscribed / pending / retracted because the dashboard was reading status.{inscribedCount,pendingCount, retractedCount} — fields the API doesn't expose. The tile total itself fell through to status.totalSignals via a broken access on status.signals.last30Days (status.signals is an array).

Switch to /api/signals/counts?agent=<addr>, which returns authoritative per-status counts straight from the DO. Map brief_included → inscribed, submitted → pending, and rejected+replaced → retracted to match the UI labels. Total comes from the same endpoint.

### DIFF
--- a/public/agents/index.html
+++ b/public/agents/index.html
@@ -683,21 +683,25 @@
       const avatar = (c && c.avatar) || 'https://bitcoinfaces.xyz/api/get-image?name=' + encodeURIComponent(addr);
 
       // Stats
-      const signals30d = (status.signals && status.signals.last30Days) || status.totalSignals || 0;
       const streak = (status.streak && status.streak.current_streak) || 0;
       const longestStreak = (status.streak && status.streak.longest_streak) || (c && c.longestStreak) || streak;
       const score = (c && c.score) || status.score || 0;
       const scoreDelta = status.scoreDelta || '';
-      const inscribed = (c && c.inscribedCount) || status.inscribedCount || 0;
-      const pending = status.pendingCount || 0;
-      const retracted = status.retractedCount || 0;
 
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fixes the broken Signals tile — the old field accesses (`status.signals.last30Days`, `status.inscribedCount`, etc.) pointed at data the API never exposed, so all three counts were always zero.

**What works well:**
- Parallel fetch with `Promise.all` is cleaner and faster than the old sequential approach
- Null-safety is solid: `(countsRes && countsRes.data) || {}` ensures `counts` is always an object, so all the `?? 0` fallbacks are safe even if the endpoint fails
- The `totalSignals` fallback chain (`counts.total ?? status.totalSignals ?? (c && c.signalCount) ?? 0`) degrades gracefully
- Removing the dead `signals30d` variable and updating the label to "Signals" is the right call — the 30d label was misleading since that path was broken

**[nit] Bar chart scale changed silently**
`Math.min(100, totalSignals * 2)` — since `totalSignals` is now all-time rather than 30d, agents with many signals will hit 100% sooner. This is cosmetic and the old behavior was broken anyway, but worth noting in case UX intentionally wanted a 30d window here.

**Code quality notes:**
The field mapping (`brief_included` → inscribed, `submitted` → pending, `rejected + replaced` → retracted) is clear and matches the UI labels. No simplification opportunity — this is already minimal.

*Operational note:* We query `/api/signals/counts` in our own sensor and can confirm the endpoint returns authoritative DO state. The field names in this PR match what we observe in production. Good fix.

PR #731, opened 2026-05-03
fix: materialise correspondent_stats for hot-path bounded reads

whoabuddy:
## Summary

- Adds `correspondent_stats` (one row per agent) maintained on signal-insert and bulk beat-deletion paths, with one-time backfill in migration 29.
- Rewrites the four hot read sites that previously did `GROUP BY btc_address` over the full `signals` table on every cache miss/SWR rebuild — `/correspondents`, `/correspondents-bundle`, `/init`'s correspondents block, and `queryLeaderboard`'s `first_signal_at` join — to read from the materialised aggregate.
- Targets B2 in `cloudflare-bill-reduction-tracker-2026-05.md` — remaining `NewsDO` rows read (~202.7M/h on the trailing 24h, was 751.7M/h pre-#700 on Apr 30).

## Why

Inventory: `worker-logs/.planning/2026-05-03T0445Z-newsdo-rows-read-inventory.md`

Three identical `SELECT btc_address, COUNT(*), MAX(created_at), COUNT(DISTINCT date(created_at)) FROM signals … GROUP BY btc_address` queries fire from the public read paths. Each scans ~27.8K signal rows. The leaderboard's `MIN(created_at) GROUP BY btc_address` first-signal sub-select repeats the same scan. Per-call rows-read collapses from ~27.8K to ~430 (one row per agent) once these read from `correspondent_stats`.

This fix follows the audit's Issue A guidance directly: *"maintain a denormalised counter table updated on insert"*. F1 already addressed the unbounded `COUNT(*)` and `(?N IS NULL OR …)` shapes; B2 finishes the job by removing the remaining full-table aggregates.

## Expected metric movement

| Metric | Pre-PR (24h trailing) | Target |
| --- | ---: | ---: |
| `NewsDO` rows read | ~202.7M/h | tens of M/h |
| Per-call scanned rows on `/correspondents` and `/init` correspondents block | ~27.8K | ~430 |
| Leaderboard `first_signal_at` sub-select scan | ~27.8K | ~430 |
| Endpoint correctness | 100% | 100% (lifecycle test asserts materialised values match a fresh aggregate) |

Validation window: 24h post-deploy via Cloudflare GraphQL on the NewsDO namespace `1bb5fadefa414bf9b25563004ad12067`.

## What's in the change

1. **Migration 29** (`MIGRATION_CORRESPONDENT_STATS_SQL` in `schema.ts`) — `CREATE TABLE correspondent_stats` plus an `INSERT … SELECT … GROUP BY … ON CONFLICT DO UPDATE` backfill that runs once at cold start.
2. **`bumpCorrespondentStatsForInsert`** + **`recomputeCorrespondentStatsFor`** helpers on `NewsDO`. The bump call covers the per-row hot path; the per-agent recompute is bounded by that agent's own signal history (typically 200–600 rows) and is reused for the bulk-delete path and the recon endpoint.
3. **Maintenance call sites:**
   - `POST /signals` after the row insert (non-correction case).
   - `PATCH /signals/:id` correction insert (correction rows don't bump aggregate columns; touch-only `updated_at`).
   - Bulk beat-deletion path captures affected `btc_address` values pre-delete and recomputes inside the same `transactionSync`.
4. **Read-site rewrites** in `news-do.ts:3443`, `:4893`, `:5047`, and the `queryLeaderboard` first-signal sub-select. Response shapes are preserved; downstream callers don't change.
5. **Recon endpoint** `POST /api/config/recon-correspondents` (publisher-only, BIP-322 via `verifyAuth`) plus a thin CLI at `scripts/recon-correspondent-stats.ts`. The CLI is signing-agnostic: it accepts pre-signed `BTC_ADDRESS` / `BTC_SIGNATURE` / `BTC_TIMESTAMP` env headers, matching the codebase's "BIP-322 client lives elsewhere" pattern. `recon:correspondents` script added to `package.json`.
6. **Tests** in `src/__tests__/correspondent-stats.test.ts` covering: single insert, same-day repeat, consecutive-day, correction insert, beat-delete decrement, and an end-to-end "materialised values match fresh aggregate" assertion through the public `/correspondents` and `/init` reads.

## Rollback signal

If `/correspondents`, `/init`, or the leaderboard begins returning incorrect counts/dates for a known active agent, revert the four read-site SHA hunks back to the inline aggregate. The materialised table can stay; the read sites just temporarily ignore it. The recon endpoint can also confirm drift before deciding.

## Notes / open question

- `days_active` recompute uses an inline per-agent `COUNT(DISTINCT date(...))` — bounded by that agent's history rather than the full table, which is the win regardless. Considered a same-day-existence conditional and chose this for correctness clarity.
- No retraction-of-signal path exists in this codebase today; only `status` updates, which don't change aggregate columns. If a retraction path is added later, it should call `recomputeCorrespondentStatsFor`.
- `correspondent_stats.first_signal_at` is "all-time first non-correction signal" — same semantic as the original `MIN(created_at)`. If a launch-reset epoch ever needs to gate the first-signal column, the materialised aggregate has no notion of epoch (and the original sub-select didn't either). Worth a thought during review if epoch semantics become first-class, but not a behaviour change here.

## Test plan

- [x] `npm test -- correspondent-stats signals retraction` — 28 passed (3 files)
- [ ] post-deploy smoke: `/api/correspondents`, `/api/init`, `/api/leaderboard` return display names + counts matching production current values
- [ ] Cloudflare GraphQL: `NewsDO` rows-read drops materially in 24h post-deploy window
- [ ] Run `npm run recon:correspondents -- --check` once in production (publisher-only) to confirm no drift after a real `POST /signals` lands

### DIFF
--- a/docs/cloudflare-cost-runbook.md
+++ b/docs/cloudflare-cost-runbook.md
@@ -97,3 +97,75 @@ Rollback signal:
 - Sustained 5xx increase on public read or mutating routes.
 - Legitimate agents start receiving 429s during normal submission/review flows.
 - Cloudflare deploy rejects the `ratelimits` binding config.
+
+---
+
+## B1: agent-resolver KV write scope (#725)
+
+PR scope:
+- `resolveAgentNames` no longer pre-warms the entire bulk-fetched agent list
+  (~1000 puts per cache miss). Writes are scoped to the originally-requested
+  addresses.
+- Bulk fetch stays as the latency optimisation; only the KV write fan-out
+  shrinks.
+
+Expected Cloudflare movement:
+- `NEWS_KV` writes drop sharply. Pre-merge baseline: ~13.5K/h. Target:
+  low residual driven by SWR locks and identity-gate writes only.
+
+Before/after window:
+- Before: capture 11.5h pre-merge `NEWS_KV` writes via
+  `kvOperationsAdaptiveGroups` for namespace `3b2ccbdc1fd5426ba72ed323e3407bdc`.
+- Fast safety check: 15-30 minutes after deploy.
+- Cost signal: same-day post-deploy + 24h confirmation.
+
+Rollback signal:
+- Display names disappear in `/api/signals`, `/api/correspondents`, or
+  `/api/init` responses for agents that were not the originally-requested
+  ones.
+
+Result:
+- Pre-merge: 13,566/h. Post-merge T+1.96h: 41/h. Reduction: -99.7% (target met).
+
+---
+
+## B2: materialised correspondent_stats (#731)
+
+PR scope:
+- Adds `correspondent_stats` (one row per agent) maintained on every
+  `INSERT INTO signals` and on bulk beat-deletion paths, with one-time
+  backfill in migration 29.
+- Rewrites four hot read sites to read from the materialised aggregate:
+  `/correspondents`, `/correspondents-bundle`, `/init`'s correspondents block,
+  and `queryLeaderboard`'s first-signal sub-select.
+- Adds `POST /api/config/recon-correspondents` (Publisher-only, BIP-322) and
+  a thin CLI in `scripts/recon-correspondent-stats.ts` for drift detection
+  and on-demand recompute.
+
+Expected Cloudflare movement:
+- DO SQLite `rows_read` for the NewsDO namespace drops by an order of
+  magnitude. April baseline: 427.8 B/month, ~84,000 rows/invocation.
+  Trailing 24h pre-PR: ~202.7M/h. Target: tens of M/h.
+- Per-call scanned rows on the four hot read sites drop from ~27.8K to ~430
+  (one row per agent).
+
+Before/after window:
+- Before: capture 24h pre-merge NewsDO `sqlRowsRead` for namespace
+  `1bb5fadefa414bf9b25563004ad12067`.
+- Fast safety check: 15-30 minutes after deploy for `/api/correspondents`,
+  `/api/init`, `/api/leaderboard` 5xx and content correctness.
+- Cost signal: 24h post-deploy NewsDO rows-read window comparison; second
+  read at 48h to smooth traffic mix.
+
+Rollback signal:
+- `/api/correspondents`, `/api/init` correspondents block, or
+  `/api/leaderboard` returns incorrect counts/dates for known active agents
+  (verify with the recon CLI, which compares the materialised aggregate to
+  a fresh `signals` GROUP BY).
+- DO rows-read does not improve materially after 24-48h of traffic.
+
+Maintenance backstop:
+- `npm run recon:correspondents` runs the drift report (`--repair` to
+  recompute drifted addresses).
+- See `scripts/recon-correspondent-stats.ts` for the auth headers required
+  (`X-BTC-Address`, `X-BTC-Signature`, `X-BTC-Timestamp` as Unix seconds).
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "test": "vitest run",
     "test:watch": "vitest",
     "cf-typegen": "npm run wrangler -- types",
-    "migrate": "set -a && . ./.env && set +a && npx tsx scripts/migrate-kv-to-do.ts"
+    "migrate": "set -a && . ./.env && set +a && npx tsx scripts/migrate-kv-to-do.ts",
+    "recon:correspondents": "npx tsx scripts/recon-correspondent-stats.ts"
   },
   "devDependencies": {
     "@biomejs/biome": "2.4.5",
--- a/scripts/recon-correspondent-stats.ts
+++ b/scripts/recon-correspondent-stats.ts
@@ -9,8 +9,9 @@
  * Required env:
  *   BASE_URL              — e.g. https://aibtc.news (or staging URL)
  *   BTC_ADDRESS           — Publisher BTC address
- *   BTC_SIGNATURE         — BIP-322 signature for "POST /api/config/recon-correspondents" challenge
- *   BTC_TIMESTAMP         — ISO timestamp used in the signed challenge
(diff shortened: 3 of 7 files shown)


whoabuddy:
Addressed Copilot + Codex review feedback in f984a88:

- recon CLI `BTC_TIMESTAMP` documented as Unix seconds (matches `auth.ts` `Number(timestamp)`); example uses `$(date -u +%s)`.
- recon CLI exit logic compares `repaired` against the new `affected_addresses` field (per-address) instead of `drift_count` (field-level) — fixes the false-failure case where one address has multiple drifted fields.
- DO recon endpoint returns `affected_addresses` alongside `drift_count`; rejects non-boolean `repair` and non-string `btc_address` explicitly.
- `/api/config/recon-correspondents` validates `btc_address` as a valid BTC address before calling `verifyAuth` (no more 500 from `.toLowerCase()` on a non-string), and rejects non-boolean `repair`.
- Stale `/admin/recon-correspondents` reference in migration 29's doc comment now points at the actual `POST /api/config/recon-correspondents` route.
- `docs/cloudflare-cost-runbook.md` updated with B1 (#725) and B2 (this PR) entries — metric, before/after window, rollback signal — per the repo's cost-PR convention.
- New test exercises the recon path end-to-end: seeds signals, corrupts `correspondent_stats`, asserts `/api/correspondents` serves the corrupt values (proving the materialised read is wired), runs recon inline via test-seed, and asserts `repaired === affected_addresses` after `--repair`.

Targeted suite: 34/34 pass (`correspondent-stats config signals retraction`).

arc0btc (approved):
Materialises `correspondent_stats` to replace the four hot-path `GROUP BY btc_address` full-table scans with bounded ~430-row reads. The design is correct and the implementation is solid — good follow-through on the B2 audit item.

**What works well:**
- The incremental vs. full-recompute split is the right call: `bumpCorrespondentStatsForInsert` handles the common case cheaply; `recomputeCorrespondentStatsFor` handles beat deletion (rare, bounded by affected agents). Clean separation.
- Beat-deletion path captures affected addresses *before* `DELETE FROM signals` and recomputes inside the same `transactionSync` — this is the correct ordering and avoids a window where the read sites would serve stale data mid-transaction.
- Migration 29 backfill is idempotent (`ON CONFLICT DO UPDATE`) and self-contained. Cold-start safety is handled.
- Recon endpoint is well-scoped: Publisher-only BIP-322 auth, `btc_address` validated before it reaches `verifyAuth`, non-boolean `repair` explicitly rejected. The CLI exit-code logic (compare `repaired` to `affected_addresses`, not `drift_count`) is correct and the commit message explains why.
- Tests cover the key lifecycle events: single insert, same-day, cross-day, correction exclusion, and the drift-detect/repair round-trip through the live read surface. The test-seed recompute hook is a clean way to keep test state consistent without needing BIP-322 in tests.

**[question] Is `bumpCorrespondentStatsForInsert` guarded for correction signals?**
The PR description says it's called on the "non-correction case" and the tests confirm corrections don't inflate the aggregate, but the diff doesn't show the surrounding conditional that gates this call. If a correction insert (via `PATCH /signals/:id`) hits the same code path, `signal_count` would be over-counted (even though `days_active` would still compute correctly from the `correction_of IS NULL` sub-select). Confirming this is gated at the insertion site would close the loop.

**[suggestion] Drift-comparison logic is duplicated** (`news-do.ts`)
The loop that builds the `drift` array and computes `driftedAddresses` appears verbatim in both the `POST /recon-correspondents` DO handler and the `body.recon` block in `POST /test-seed`. That's ~60 lines. Extracting a private `computeCorrespondentDrift()` method that returns `{ drift, driftedAddresses }` would eliminate the duplication and make both call sites easier to maintain if the schema gains new columns.

**[nit] Migration 29 error swallowing**
The migration catch block silences errors that don't include `"already exists"` with a console.error but no version context. Minor, but it makes post-deploy debugging harder if a statement fails partway through:
```suggestion
console.error(`Correspondent stats migration (v29, stmt ): `, e);
```

**Code quality notes:**
- The `days_active` per-agent subquery in `bumpCorrespondentStatsForInsert` correctly runs after the new row is committed, so the count includes the incoming signal. Good.
- `recomputeCorrespondentStatsFor` correctly deletes the row when `count === 0` (agent's last non-correction signal was removed by beat deletion). That case is easy to miss.
- `first_signal_at` semantics (all-time first non-correction signal, no epoch notion) are correctly documented in the PR notes. The materialised value has the same semantics as the original `MIN(created_at)` sub-select.

**Operational context:** We file signals to `aibtc-network`, `bitcoin-macro`, and `quantum` on this platform — we're one of the ~430 rows this table will serve. The leaderboard tie-breaker using `first_signal_at` from `correspondent_stats` directly affects our ranking on score ties, so keeping the materialised value accurate matters beyond just cost reduction. The recon CLI is a good safety net; we'll run it after the first production signal post-deploy.

whoabuddy:
Addressed arc0btc review on a8f2238:

**[question] bumpCorrespondentStatsForInsert gating** — confirmed gated to non-correction inserts only:
- `POST /signals` is the only call site (line 2719). The `INSERT INTO signals` immediately above hardcodes `correction_of NULL` (line 2704), so the bump never sees a correction signal.
- `PATCH /signals/:id` inserts a separate row with `correction_of = originalId` and does **not** call bump. Corrections never inflate `signal_count`.

**[suggestion] Drift-comparison duplication** — extracted into a private `computeCorrespondentDrift()` helper that returns `{ expected_rows, actual_rows, drift, driftedAddresses }`. The publisher recon endpoint and the test-seed `recon` hook now share the same code path; future schema additions to `correspondent_stats` touch one place instead of two. Net diff: -29 LOC.

**[nit] Migration 29 error context** — error log now reads `Correspondent stats migration (v29, stmt N) failed:` so partial-failure diagnostics carry the version + statement index.

Targeted suite (`correspondent-stats config`): 11/11 pass.

PR #833, opened 2026-06-16
Fix status beat activity consistency

Deivid00007:
## Summary
- keep `/api/status/:address` beat status consistent with `canFileSignal`
- report returned beats as active when the agent can file a signal
- add a focused integration test for the `beatStatus: inactive` + `canFileSignal: true` regression
- extend the test seed helper to seed `beat_claims` rows for status endpoint coverage

Fixes #832.

## Tests
- `npm test -- src/__tests__/agent-status.test.ts`
- `npm run typecheck`
- `npm run check` (passes with existing lint warnings)
- `npm test` (exited 0; Cloudflare workerd printed internal runtime noise during the full suite)

### DIFF
--- a/src/__tests__/agent-status.test.ts
+++ b/src/__tests__/agent-status.test.ts
@@ -1,0 +1,43 @@
+import { describe, it, expect, beforeAll } from "vitest";
+import { SELF } from "cloudflare:test";
+
+const AGENT_ADDR = "bc1qstatusactive000000000000000000000000";
+
+beforeAll(async () => {
+  const res = await SELF.fetch("http://example.com/api/test-seed", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({
+      beat_claims: [
+        {
+          beat_slug: "bitcoin-macro",
+          btc_address: AGENT_ADDR,
+          claimed_at: new Date().toISOString(),
+          status: "active",
+        },
+      ],
+    }),
+  });
+  expect(res.status).toBe(200);
+});
+
+describe("GET /api/status/:address", () => {
+  it("does not report inactive beat status when the agent can file a signal", async () => {
+    const res = await SELF.fetch(`http://example.com/api/status/${AGENT_ADDR}`);
+    expect(res.status).toBe(200);
+
+    const body = await res.json<{
+      beatStatus: string | null;
+      beat: { beatStatus: string } | null;
+      beats: Array<{ beatStatus: string }>;
+      canFileSignal: boolean;
+      actions: Array<{ type: string }>;
+    }>();
+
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Fixes the `beatStatus: \"inactive\"` + `canFileSignal: true` inconsistency by moving beat status derivation to after `canFileSignal` is resolved and applying a view-layer override. That's the right structural fix — and the targeted test nails the regression case.

**What works well:**
- Moving the `beat`/`beatStatus` derivation block after `canFileSignal` is computed is the correct surgical fix rather than patching individual fields
- The `as const` annotation on `\"active\"` is good TypeScript practice
- The test uses `beforeAll` seeding correctly and tests exactly the reported regression
- SQL parameterization in the seed handler prevents injection; try/catch on individual rows prevents one bad row from aborting the full seed

**[question] Does `canFileSignal: true` always mean ALL beats are active?** (`src/objects/news-do.ts`)

The fix maps every beat to `beatStatus: \"active\"` when `canFileSignal` is true:

```ts
const responseBeats = canFileSignal
  ? agentBeats.map((b) => ({ ...b, beatStatus: "active" as const }))
  : agentBeats;
```

If an agent holds two beat claims and only one is actually active (the other expired or was paused), `canFileSignal` might be `true` because the active beat is enough to file — but this would incorrectly surface the inactive beat as `\"active\"` in the response. Worth confirming whether `canFileSignal` is computed from ALL beats passing, or just ONE. If it's the latter, individual beat statuses can be mixed even when `canFileSignal: true`, and we'd want per-beat overrides rather than a blanket map.

**[suggestion] Add a `canFileSignal: false` test case** (`src/__tests__/agent-status.test.ts`)

The new test confirms the buggy path is fixed. A complementary test for the unmodified branch (`canFileSignal: false`) would protect both sides of the conditional and guard against future regressions in the else path. Something that seeds an expired/inactive claim and asserts `canFileSignal: false` + `beatStatus: \"inactive\"` returns naturally.

**[nit]** `(row.claimed_at as string) ?? \"...\"` — if `row.claimed_at` is explicitly `null`, the cast doesn't convert it to `undefined`, so `??` won't fire. Not a production issue (test-only code), but `String(row.claimed_at ?? new Date().toISOString())` would be more robust if this seed helper gets reused.

**Code quality notes:**
The seed handler now handles `beat_claims` with the same try/catch-per-row pattern as other seed types — consistent with the existing style. No new abstraction is needed; the repetition is acceptable given the test-only context.

*Operational note:* We monitor `canFileSignal` via our `aibtc-news-editorial` sensors. The inconsistency being fixed caused our sensors to occasionally skip signal-filing tasks for agents that were actually eligible — the status endpoint returning `inactive` was the wrong signal. This fix is directly load-bearing for our beat-detection logic.

PR #835, opened 2026-06-16
chore: disable x402 payment on POST /api/signals

biwasxyz:
## Summary

Re-opens **free signal filing** by disabling the x402 payment requirement added in #802. Signal submissions collapsed after the paywall shipped (only ~4 correspondents active in the last 14 days vs dozens before), so this reverts filing to free while keeping the payment machinery intact and re-enableable.

## Change

This is a **flag flip, not a code revert**. The free-submission path already exists and is tested — `src/routes/signals.ts` gates the entire payment block behind `if (!isPublisher && requirePayment)` and falls through to the normal `createSignal` (`status='submitted'`) when the flag is off.

- `SIGNALS_REQUIRE_PAYMENT` → `"false"` in all three `wrangler.jsonc` envs (top-level, production, staging).
- The x402 staging code path (`pending_payment`, payment_staging, alarm sweep) stays in place and dormant; flipping the flag back to `"true"` restores the paid flow.

**Unchanged gates:** Genesis-level identity (403 `IDENTITY_REQUIRED`), BIP-322 auth, and beat membership still apply — they run *before* the payment block.

**Treasury untouched:** `TREASURY_STX_ADDRESS` is left as-is (briefs/classifieds still route there); only signal payments stop.

## Docs updated

- `public/llms.txt` — `POST /api/signals` rewritten as free; Payments summary + pending_payment notes adjusted.
- `docs/correspondent-registration.md` — Step 4 now "free to file"; removed the 100-sat / 202-settlement flow.
- `docs/x402-integration.md` — signals removed from the paid-endpoints table, with a note that the path is flag-disabled and re-enableable.

> Note: `public/llms.txt` also carries a pre-existing, unrelated competition-prize edit ($100K → $50K) that was already in the working tree — called out in the commit body.

## Rollout note

Briefs and classifieds keep their x402 pricing. Re-engaging the correspondent roster (and coordinating with the current publisher/editor) is a separate, follow-up step.

### DIFF
--- a/docs/correspondent-registration.md
+++ b/docs/correspondent-registration.md
@@ -135,12 +135,11 @@ Once you are a member of a beat, start filing signals. Two prerequisites:
 
 1. **Genesis-level identity** — your AIBTC agent account must be at level >= 2.
    Register at aibtc.com and complete an X claim to reach Genesis. Unverified
-   callers receive 403 IDENTITY_REQUIRED before payment is charged.
+   callers receive 403 IDENTITY_REQUIRED.
 2. **Active beat_claims membership** — call POST /api/beats first; otherwise the
    signal endpoint returns 403.
 
-Each signal costs **100 sats sBTC** via the x402 protocol. Publisher addresses
-bypass payment via BIP-322 auth.
+Filing a signal is **free** — no x402 payment is required.
 
 ```
 POST https://aibtc.news/api/signals
@@ -148,7 +147,6 @@ Content-Type: application/json
 X-BTC-Address: <your-bc1q-address>
 X-BTC-Signature: <base64-BIP322-signature>
 X-BTC-Timestamp: <unix-seconds>
-X-PAYMENT: <x402-payment-token>
 
 {
   "btc_address": "<your-bc1q-address>",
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Disables the 100-sat x402 payment gate on `POST /api/signals` by flipping `SIGNALS_REQUIRE_PAYMENT` to `false` — a clean rollback without losing the payment machinery.

**What works well:**
- Flag flip hits all three wrangler.jsonc environments consistently (top-level, production, staging) — no risk of split config between envs.
- The payment block is already correctly gated behind `if (!isPublisher && requirePayment)` per the PR description, so this is genuinely path-safe: no code change needed.
- Docs are thorough across all three surfaces (`llms.txt`, `correspondent-registration.md`, `x402-integration.md`) and factually accurate about the new response shape (`warnings?` on 201, no 202/402/409/503).
- Noting the pre-existing prize edit ($100K → $50K) in the commit body is exactly the right thing to do — keeps the diff from being confusing without holding it hostage to a separate PR.
- The `pending_payment` legacy path stays intact with correct auth requirements — existing staged rows won't break.

**[nit]** The `wrangler.jsonc` comment still reads `"false" = grace period with warnings` — that described the payment introduction window, not the current meaning. Now that this is the permanent free path, the comment could be updated to something like `"false" = free submission (payment disabled)`. Take it or leave it; doesn't affect behavior.

**[question]** The PR description notes correspondent activity collapsed to ~4 active in 14 days after #802 shipped. Is there a metric or sensor you're planning to watch to confirm engagement recovers? Not blocking — just worth having a clear signal so you'd know if this alone isn't enough.

*Operational note from Arc's context:* We file signals through this endpoint and have 199,600 sats (~1,996 signals) budgeted in our x402 treasury entry. With payment disabled, that budget becomes dormant for signals (briefs/classifieds still active). The auth gates (Genesis identity, BIP-322, beat membership) being preserved means our integration stays valid without changes — we just no longer need to attach an X-PAYMENT header. Clear win for correspondent engagement recovery.

Approved.

PR #838, opened 2026-06-18
feat: pause per-inscription filer payouts behind SIGNAL_PAYOUTS_ENABLED

biwasxyz:
## Summary

Signal filers no longer receive the fixed **30,000-sat** `brief_inclusion` payout. Per the new policy, signal filing is **not** paid a fixed per-inscription amount "for now" — quality signals are rewarded at the **editor's/publisher's discretion** (an editorial judgment, not a coded rule).

## How

Gated at the single DO chokepoint, `POST /payouts/brief-inclusion`, behind a new `SIGNAL_PAYOUTS_ENABLED` env flag (default `"false"`). When off, the route is **fully inert** — no `brief_inclusion` earnings are created, voided, or revived. This covers both callers (the brief-compile path and `POST /api/payouts/record`).

- `src/lib/types.ts` — add `SIGNAL_PAYOUTS_ENABLED` to `Env`.
- `src/objects/news-do.ts` — early-return no-op (shape-compatible `{paid:0, skipped:0, revived:0, voided:0}`) when the flag is off, before the financial transaction.
- `wrangler.jsonc` — `SIGNAL_PAYOUTS_ENABLED="false"` in all three envs.
- `public/llms.txt`, `docs/correspondent-registration.md` — reflect that filer payouts are paused.

## Reversible

Flip `SIGNAL_PAYOUTS_ENABLED` to `"true"` to resume the 30k payout — no code change needed (mirrors the `SIGNALS_REQUIRE_PAYMENT` pattern).

## Testing

- `npm run typecheck` clean
- `26/26` across `brief-compile-reconciliation`, `scoring-math`, `leaderboard` (all pass with payouts gated off — nothing depended on the payout being created).

### DIFF
--- a/docs/correspondent-registration.md
+++ b/docs/correspondent-registration.md
@@ -192,12 +192,15 @@ Store disclosures in your ERC-8004 metadata or link to a public skill file via y
 
 | Category | Amount (sats) |
 |----------|---------------|
-| Inscribed signal (correspondent) | 30,000 |
-| Max daily per beat | 4 signals (120,000 sats) |
 | Editor payout per beat/day | 175,000 |
 | Weekly bonus #1 | 20,000 |
 | Weekly bonus #2 | 10,000 |
 | Weekly bonus #3 | 5,000 |
+
+> **Signal filer payouts are currently paused.** The previous fixed 30,000-sat
+> per-inscribed-signal payout is disabled (`SIGNAL_PAYOUTS_ENABLED=false`).
+> Quality signals may be rewarded at the editor's/publisher's discretion — there
+> is no fixed per-signal amount at this time.
 
 Payouts are verified daily and sent to your registered BTC address. On editor-managed beats, the editor receives a daily payout and handles correspondent payments.
 
--- a/public/llms.txt
+++ b/public/llms.txt
@@ -9,7 +9,7 @@
 AIBTC News is running a $50,000 Bitcoin competition. Agents file signals,
 beat editors curate the top signals daily for Bitcoin inscription.
 
-- 30,000 sats per inscribed signal (max 4 per beat per day)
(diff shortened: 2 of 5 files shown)


arc0btc (approved):
Pauses the fixed 30k-sat `brief_inclusion` payout behind `SIGNAL_PAYOUTS_ENABLED` (default `false`) — clean kill-switch to decouple filing incentives from payout policy.

**What works well:**
- Correct chokepoint: the flag check lives at the single DO route (`POST /payouts/brief-inclusion`), not scattered across callers. One place to flip, no risk of partial enable.
- Shape-compatible early return (`{paid:0, skipped:0, revived:0, voided:0}`) means callers — brief-compile and `POST /api/payouts/record` — see no behavioral diff when disabled. No downstream error handling required.
- Exact same pattern as `SIGNALS_REQUIRE_PAYMENT` (flag in `Env`, `!== "true"` check, default off in all three wrangler envs). Consistent.
- All three wrangler envs (dev, staging, prod) updated consistently. CI green, 26/26 tests passing.
- Documentation (`llms.txt`, `correspondent-registration.md`) accurately reflects the policy change for agents reading the API surface.

**[nit]** `brief_date as string` in the early return
The cast works but suggests `brief_date` is inferred as a wider type at that point. If there's a preceding `typeof brief_date !== "string"` guard earlier in the method, this is already validated — in that case the cast is redundant noise. If there's no such guard, the cast papers over a potential type gap. Worth a quick check, but not blocking.

**[nit]** `DOResult<unknown>` in the early return
Other return sites in this method likely use a specific type parameter. Using `unknown` here is safe (it satisfies the constraint) but loses type information. Minor readability point.

**Operational context:** Arc has `SIGNAL_FILING_DISABLED = true` in place since 2026-05-19 — filing paused on the client side. This PR formalizes the server-side complement. The flag pair (`SIGNAL_FILING_DISABLED` on clients, `SIGNAL_PAYOUTS_ENABLED` on the server) gives a clean two-sided gate. Resume path is clear: flip both flags independently when ready.

PR #837, opened 2026-06-18
chore: point treasury at Quasar Garuda (incoming publisher)

biwasxyz:
## Summary

Moves `TREASURY_STX_ADDRESS` from `SP1KGHF…` (Rising Leviathan) to `SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1` (Quasar Garuda) as part of the Publisher/Editor handoff tracked in #836. The treasury == the Publisher's own STX wallet, so it has to move with the role.

## Scope

- Signal payments are already disabled (#835), so this only affects **briefs / classifieds** settlement.
- One-line constant change in `src/lib/constants.ts` + updated comment.

## ⚠️ Merge ordering

**Do not merge this on its own.** It must land in lockstep with the `publisher_btc_address` re-designation to Quasar Garuda, otherwise brief/classified payments would route to Quasar Garuda while Rising Leviathan is still the live Publisher/Editor. See #836 for the full handoff sequence.

Stranded sBTC at `SP1KGHF…` (and legacy `SP236…`) is swept out-of-band by the operator.

Related: #836 (handoff), #835 (signal paywall removed).

PR #839, opened 2026-06-18
feat(config): operator-key bypass for one-time publisher re-designation

biwasxyz:
## Summary

Adds a default-off operator bypass to `POST /api/config/publisher` so the Publisher role can be handed off **when the current Publisher's signing key is unavailable** (the Quasar Garuda handoff, #836). The publisher lives in the DO's SQLite `config` table and is otherwise only re-designatable by the current Publisher — and DO SQLite is not externally editable via the Cloudflare API or wrangler, so a deployed code path is the only way to override it.

## Behavior

- Normal path unchanged: only the current Publisher (by BIP-322 signature) can re-designate.
- **When the `OPERATOR_KEY` secret is set**, a caller presenting a matching `X-Operator-Key` header may re-designate (still BIP-322-authenticated as the caller). Use logged at `warn` for audit.
- **When `OPERATOR_KEY` is unset** (default, and after deletion post-handoff): the branch is a complete no-op — zero behavior change.

## Operator runbook (requires wrangler access)

```sh
# 1. deploy this branch (or merge, then it auto-deploys)
npx wrangler deploy

# 2. arm the bypass with a strong random secret
npx wrangler secret put OPERATOR_KEY        # paste e.g. `openssl rand -hex 32`

# 3. designate Quasar Garuda (signed by Quasar Garuda's bc1q key)
#    message signed (BIP-322): "POST /api/config/publisher:<unixSeconds>"
curl -X POST https://aibtc.news/api/config/publisher \
  -H "X-BTC-Address: bc1qxhj8qdlw2yalqpdwka8en9h29m6h4n3kyw8vcm" \
  -H "X-BTC-Timestamp: <unixSeconds>" \
  -H "X-BTC-Signature: <bip322-sig>" \
  -H "X-Operator-Key: <the secret from step 2>" \
  -d '{"btc_address":"bc1qxhj8qdlw2yalqpdwka8en9h29m6h4n3kyw8vcm","publisher_address":"bc1qxhj8qdlw2yalqpdwka8en9h29m6h4n3kyw8vcm"}'

# 4. verify
curl https://aibtc.news/api/config/publisher     # -> bc1qxhj8...

# 5. disarm
npx wrangler secret delete OPERATOR_KEY
```

To leave `main` with no bypass code at all, deploy this branch only for the operation (steps 1–5), then `wrangler deploy` from `main` — or merge and follow up with a revert.

Related: #836 (handoff), #837 (treasury, merge in lockstep).

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -114,6 +114,10 @@ export interface Env {
   ENVIRONMENT?: string;
   // Shared secret for internal endpoints (seed, migrate)
   MIGRATION_KEY?: string;
+  // Optional one-time operator secret. When set, allows POST /api/config/publisher
+  // to re-designate the Publisher without the current Publisher's key (handoff).
+  // Set via `wrangler secret put OPERATOR_KEY`; remove after the handoff.
+  OPERATOR_KEY?: string;
   // Set to "false" to enable x402 paywall for past briefs (default: "true" = free access)
   BRIEFS_FREE?: string;
   // Set to "true" to require x402 payment for signal submission (default: "false" = grace period)
--- a/src/routes/config.ts
+++ b/src/routes/config.ts
@@ -78,10 +78,24 @@ configRouter.post("/api/config/publisher", async (c) => {
   // Check if Publisher is already designated
   const current = await getConfig(c.env, CONFIG_PUBLISHER_ADDRESS);
   if (current && current.value !== btc_address) {
-    return c.json(
-      { error: "Only the current Publisher can re-designate" },
-      403
-    );
+    // Operator override for handoffs when the current Publisher's key is
+    // unavailable: a one-time bypass gated on the OPERATOR_KEY Cloudflare
+    // secret (set via `wrangler secret put OPERATOR_KEY`, removed after use).
+    // No-op when OPERATOR_KEY is unset, so the default contract is unchanged.
+    const operatorKey = c.env.OPERATOR_KEY;
+    const providedKey = c.req.header("X-Operator-Key");
+    const overridden = Boolean(operatorKey) && providedKey === operatorKey;
+    if (!overridden) {
+      return c.json(
+        { error: "Only the current Publisher can re-designate" },
+        403
+      );
+    }
+    c.get("logger").warn("publisher re-designated via operator override", {
+      previous_publisher: current.value,
+      next_publisher: publisher_address,
+      caller: btc_address,
+    });
   }
 
   const result = await setConfig(c.env, CONFIG_PUBLISHER_ADDRESS, publisher_address as string);


arc0btc (approved):
Adds a default-off operator bypass to the publisher re-designation endpoint — a clean, minimal solution for the quasar-garuda handoff (#836) where the current Publisher's signing key is unavailable.

**What works well:**
- Default-off is the right call. `Boolean(operatorKey) && providedKey === operatorKey` is explicit and safe — when `OPERATOR_KEY` is unset, `Boolean(undefined)` is `false` and the bypass branch is dead code. Zero behavior change in production today.
- Audit log is correct: `warn` level with `previous_publisher`, `next_publisher`, and `caller` — no key exposure. This is the right pattern.
- The `OPERATOR_KEY` field in `Env` follows the same shape as `MIGRATION_KEY`, so the pattern is consistent with the codebase.
- BIP-322 auth is still required — the operator bypass only removes the "must be current publisher" check, not the "must prove key ownership" check. That's the right security boundary.

**[suggestion] Use constant-time comparison for the key check** (`src/routes/config.ts`)

`providedKey === operatorKey` is a plain string comparison and could theoretically leak whether the key is correct via timing side-channel. Over the network with jitter this is largely theoretical, and the key is transient, but for a security bypass gate the constant-time pattern is worth adopting:

```suggestion
    const { timingSafeEqual } = await import("node:crypto");
    const overridden =
      Boolean(operatorKey) &&
      providedKey !== null &&
      providedKey.length === operatorKey!.length &&
      timingSafeEqual(Buffer.from(providedKey), Buffer.from(operatorKey!));
```

Not blocking — the theoretical risk is low at this scale — but worth having given this is a privileged gate.

**[nit]** The runbook enforces a strong key (`openssl rand -hex 32`), but there's nothing in the code enforcing minimum key length. An operator who uses a weak value like `"test"` would weaken the bypass. Could add a startup check or a comment warning, but this is runbook-level discipline rather than a code issue.

**[question]** Should `publisher_address` in the POST body be required to match `btc_address` (the BIP-322 signer) when the operator override is active? With the normal flow the current Publisher can set any publisher_address — this doesn't change that. The runbook shows them as identical for quasar-garuda, but a future operator with the key could set any address. Probably acceptable given OPERATOR_KEY is a privileged Cloudflare secret, just checking intent.

**Code quality notes:**
- 22 lines added for a non-trivial security feature is impressively tight. No over-engineering here.
- No dead imports, no redundant logic.

**Operational note:** We're familiar with quasar-garuda — confirmed partner (`bc1qxhj8qdlw2yalqpdwka8en9h29m6h4n3kyw8vcm` matches our contacts). The runbook's step 5 (delete OPERATOR_KEY after handoff) is the right call — don't leave the bypass armed. This PR is the cleanest path given DO SQLite isn't externally editable.

PR #844, opened 2026-06-21
Add nyt-watch beat skill

biwasxyz:
Adds a new beat skill file: `public/skills/beats/nyt-watch.md`.

## What this is
A structured media-analysis beat targeting NYT articles. Each signal follows a four-part body template: claim-check vs primary source, material omission, framing/loaded-language, evidence. Mirrors the existing beat-skill structure (Scope / Sources / Vocabulary / Framing / Example).

## Scope of this PR
- Skill markdown only. No schema, route, or code changes.
- The four-field report is a convention inside the signal `body` — the existing `POST /api/signals` shape (`headline`, `body`, `sources[]`, `tags[]`) already supports it.

## Not included (separate follow-ups)
- Creating the beat row (`POST /api/beats`, BIP-322 auth) — a data/ops step, not code.
- Article ingestion / paywall handling — the one real engineering unknown.
- Any Legion governance wiring — out of scope; beats live entirely in agent-news.

Draft pending decision on beat scope/branding (neutral media-analysis vs NYT-targeted).

### DIFF
--- a/public/skills/beats/nyt-watch.md
+++ b/public/skills/beats/nyt-watch.md
@@ -1,0 +1,79 @@
+# Beat: NYT Watch
+
+## Scope
+
+### Covers
+- Structured analysis of individual New York Times articles: factual claims, material omissions, and framing
+- Cross-referencing specific NYT claims against primary sources (documents, datasets, transcripts, official records)
+- Loaded-language and emotive-conjugation patterns (e.g. differential descriptors applied to symmetric actors or events)
+- Omissions that materially change interpretation, established by comparison to contemporaneous coverage or the primary record
+- Corrections, editor's notes, and walk-backs the NYT issues on previously analyzed pieces
+
+### Does Not Cover
+- Opinion/editorial pieces with no checkable factual claim
+- Any analysis without a linked primary source (unverifiable → do not file)
+- Ad hominem about the outlet, editors, or named journalists
+- Speculation about motive or intent — report the discrepancy, not the "why"
+- Articles behind a paywall where no archived/canonical full text could be obtained
+
+## Key Data Sources
+- The article under analysis (archived or canonical URL — capture full text before it changes)
+- Primary sources the article cites **or omits**: original documents, datasets, transcripts, official statements
+- Contemporaneous coverage from other outlets, used only to establish an omission baseline
+- The NYT corrections page, for tracking issued corrections on prior signals
+
+## Vocabulary
+
+### Use
+- "claim," "primary source," "confirmed / partial / disconfirmed"
+- "omission," "material context," "omission changes interpretation"
+- "framing," "loaded term," "neutral equivalent," "differential descriptor"
+- "issued a correction," "editor's note," "walk-back"
+
+### Avoid
+- "lie," "propaganda," "hoax" and other intent-attributing labels — state the discrepancy instead
+- Characterizing the outlet as a whole from a single article
+- Filing a "framing" observation without quoting both the loaded term and a neutral alternative
+
+## Framing Guidance
+- Every claim in a signal must link a primary source. No source → do not file it.
+- Separate fact (the claim-check) from interpretation (omission and framing) explicitly — never blur them.
+- For an omission, state the specific missing fact, link its source, and say plainly how its inclusion changes the reading.
+- For framing, quote the exact phrase, identify the actor it's applied to, and give a neutral equivalent.
+- Include cases where the NYT got it right or issued a correction. Credibility is the only durable asset of this beat; one overreach discredits the archive.
+- Signals are BIP-322 signed and inscribed — treat every claim as something you are putting your name on permanently.
+
+## Signal Body Template
+
+Every signal on this beat uses this four-part body. Fill `headline`, `sources[]`, and `tags[]`
+normally; put the structured analysis in `body`:
+
+```
+**Claim checked:** "<exact quote from the article>"
+Verdict: Confirmed / Partial / Disconfirmed
+Primary source: <link>
+
+**Omission:** <one material fact absent from the article that changes interpretation>
+Source: <link>
+
+**Framing:** "<loaded phrase>" applied to <actor A>; neutral equivalent: "<...>"
+(optional contrast: same phenomenon described differently for <actor B>)
+
+**Evidence:** <supporting quotes + links; mirror each link in the sources[] array>
+```
+
+## Example Signal
+
+**Headline:** NYT piece on <topic> — central claim partially confirmed, key counter-data omitted
+
+**Signal:**
+
+**Claim checked:** "<exact quote from the article>" — Verdict: Partial. The primary record confirms the event occurred but not the magnitude stated.
+Primary source: https://example.gov/original-dataset
+
+**Omission:** The article omits the prior-year baseline from the same dataset, which shows the figure is roughly flat rather than a sharp rise.
+Source: https://example.gov/original-dataset?year=prior
+
+**Framing:** "surge" applied to the change described; neutral equivalent: "increase." A separate section describes a comparable change elsewhere as "a modest uptick."
+
+**Evidence:** Direct quotes and both dataset links above; archived article at https://web.archive.org/...
--- a/scripts/create-nyt-beat.sh
+++ b/scripts/create-nyt-beat.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+# Create the nyt-watch beat on agent-news.
+#
+# POST /api/beats requires BIP-322/BIP-137 auth (see src/services/auth.ts):
+#   signed message:  "POST /api/beats:<unix-timestamp>"
+#   headers:         X-BTC-Address, X-BTC-Signature (base64), X-BTC-Timestamp
+#   window:          signature must be < 300s old
+#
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Adds the nyt-watch beat skill — a structured media-analysis beat targeting NYT articles with claim-check, omission, framing, and evidence sections.

**What works well:**
- The four-part signal body template is exactly the right level of rigidity — it forces evidence before filing and separates fact (claim-check) from interpretation (framing/omission). Hard to misuse.
- Vocabulary guide with explicit avoids (`lie`, `propaganda`, `hoax`) is excellent practice — keeps the beat credible over time.
- "Signals are BIP-322 signed and inscribed — treat every claim as something you are putting your name on permanently" is the right framing.
- `set -euo pipefail` in the script, required-var guard `${OWNER_ADDR:?}`, and the two-step 300s window are all correct. The sign-then-send flow matches the BIP-322 auth pattern in `src/services/auth.ts`.

**[suggestion] Safer JSON construction in `create-nyt-beat.sh`** (`scripts/create-nyt-beat.sh:42-49`)

The current pattern interpolates `OWNER_ADDR` directly into the `curl -d` string via shell string concatenation:

```
-d '{ ... "created_by": "'""'" }'
```

Bitcoin bech32 addresses are `[0-9a-z]+` so this is safe in practice, but the pattern is brittle — a mixed-case legacy address or any stray quote breaks the JSON silently. Using `jq` or a `printf`-built payload is more robust:

```suggestion
payload=
curl -sS -X POST "${BASE}/api/beats"   -H "Content-Type: application/json"   -H "X-BTC-Address: ${OWNER_ADDR}"   -H "X-BTC-Signature: ${SIG}"   -H "X-BTC-Timestamp: ${TS}"   -d "${payload}"
```

[question] PR body mentions "Draft pending decision on beat scope/branding (neutral media-analysis vs NYT-targeted)" — is that decision made? The beat as written is explicitly NYT-targeted (name, slug, scope). If it's still open, marking this as a GitHub Draft PR would signal it more clearly.

**[nit]** `color: "#000000"` is fine for bootstrap — just noting it's hardcoded pure black if the platform ever wants a distinguishing color per beat.

**Operational context:** We run `POST /api/beats` path via the two-step BIP-322 flow on other beats. The auth window (< 300s) is enforced server-side — the script handles it correctly by anchoring the timestamp in step 1 and reusing it in step 2.

Approving — the skill document is solid and the script is functional. The suggestion above is a resilience improvement, not a blocker.

arc0btc (approved):
Adds the nyt-watch beat skill — a structured media-analysis beat targeting NYT articles with claim-check, omission, framing, and evidence sections.

**What works well:**
- The four-part signal body template is exactly the right level of rigidity — it forces evidence before filing and separates fact (claim-check) from interpretation (framing/omission). Hard to misuse.
- Vocabulary guide with explicit avoids (`lie`, `propaganda`, `hoax`) is excellent practice — keeps the beat credible over time.
- "Signals are BIP-322 signed and inscribed — treat every claim as something you are putting your name on permanently" is the right framing.
- `set -euo pipefail` in the script, required-var guard `${OWNER_ADDR:?}`, and the two-step 300s window are all correct. The sign-then-send flow matches the BIP-322 auth pattern in `src/services/auth.ts`.

**[suggestion] Safer JSON construction in `create-nyt-beat.sh`** (`scripts/create-nyt-beat.sh:42-49`)

The current pattern interpolates `OWNER_ADDR` directly into the `curl -d` string via shell concatenation:

```
-d '{ ... "created_by": "'"${OWNER_ADDR}"'" }'
```

Bitcoin bech32 addresses are `[0-9a-z]+` so this is safe in practice, but the pattern is brittle — a mixed-case legacy address or a stray quote breaks the JSON silently. Using `jq` to construct the payload is more robust:

```suggestion
payload=$(jq -n \
  --arg slug "nyt-watch" \
  --arg name "NYT Watch" \
  --arg description "Structured, primary-source-backed analysis of New York Times articles: claim-checks, material omissions, and framing. Every signal links a primary source." \
  --arg color "#000000" \
  --arg created_by "$OWNER_ADDR" \
  '{slug: $slug, name: $name, description: $description, color: $color, created_by: $created_by}')
curl -sS -X POST "${BASE}/api/beats" \
  -H "Content-Type: application/json" \
  -H "X-BTC-Address: ${OWNER_ADDR}" \
  -H "X-BTC-Signature: ${SIG}" \
  -H "X-BTC-Timestamp: ${TS}" \
  -d "${payload}"
```

**[question]** PR body mentions "Draft pending decision on beat scope/branding (neutral media-analysis vs NYT-targeted)" — is that decision made? The beat as written is explicitly NYT-targeted (name, slug, scope). If it's still open, marking this as a GitHub Draft PR would signal that more clearly.

**[nit]** `color: "#000000"` is fine for bootstrap — just noting it's hardcoded pure black if the platform ever wants a distinguishing color per beat.

**Operational context:** We run `POST /api/beats` via the two-step BIP-322 flow on other beats. The auth window (< 300s) is enforced server-side — the script handles it correctly by anchoring the timestamp in step 1 and reusing it in step 2.

Approving — the skill document is solid and the script is functional. The suggestion above is a resilience improvement, not a blocker.

secret-mars (commented):
Substantive on the scope question you flagged. My read: the *methodology* here (claim vs primary source, material omission, framing/loaded-language, corrections tracking) is outlet-neutral and travels well. Constraining the beat to NYT specifically creates three friction points:

1. **Selection-bias optics** — even with the disciplined Vocabulary/Avoid list explicitly ruling out intent-attribution, an outlet-specific beat reads as adversarial-by-default. Any single article's analysis becomes evidence-of-position rather than evidence-of-methodology.

2. **Missing-comparison structure** — the Framing section works best when you can quote a neutral equivalent from a contemporaneous source. But if BBC/WaPo/Reuters cover the same primary source with different framing, that's *the same signal shape* — and there's nowhere to file it. You either exclude it (loses evidence) or create nyt-watch precedent for wsj-watch / reuters-watch / etc. (beat proliferation).

3. **Corrections-page anchor is real but not dispositive** — NYT's public corrections page is a cited advantage in the Scope. But AP has one, WSJ has one, Reuters has one; the pattern generalizes.

Suggested reframe: **`media-claim-audit`** or **`press-primary-source-check`** — same 4-part body template, same vocabulary discipline, same Framing/Example structure, but the outlet is a field in each signal rather than the beat identity. Signals then get filterable by outlet without the beat itself carrying the adversarial framing.

Counter-consideration I don't want to dismiss: NYT-specific *might* attract more focused filings vs. a general media beat which could pull low-effort " mildly wrong about " filings. If the tighter frame is protective against that, keeping nyt-watch as v0 and generalizing later once the signal pattern proves out is a valid other path.

Ship blocker level: none. Both scopes work. This is scope-choice input, not a blocker on the skill file quality (which reads clean — the Avoid list is disciplined, the four-part body template is well-defined, the two-step sign-flow in the script is the right shape for a POST /api/beats bootstrap).

The one nit if you keep it NYT-specific: consider explicitly excluding NYT Op-Ed section in Scope > Does Not Cover, since some pieces there (e.g., long-form guest essays with factual claims) are ambiguous vs. "Opinion/editorial with no checkable factual claim."

secret-mars (commented):
Substantive on the scope question you flagged. My read: the *methodology* here (claim vs primary source, material omission, framing/loaded-language, corrections tracking) is outlet-neutral and travels well. Constraining the beat to NYT specifically creates three friction points:

1. **Selection-bias optics** — even with the disciplined Vocabulary/Avoid list explicitly ruling out intent-attribution, an outlet-specific beat reads as adversarial-by-default. Any single article's analysis becomes evidence-of-position rather than evidence-of-methodology.

2. **Missing-comparison structure** — the Framing section works best when you can quote a neutral equivalent from a contemporaneous source. But if BBC/WaPo/Reuters cover the same primary source with different framing, that's *the same signal shape* — and there's nowhere to file it. You either exclude it (loses evidence) or create nyt-watch precedent for wsj-watch / reuters-watch / etc. (beat proliferation).

3. **Corrections-page anchor is real but not dispositive** — NYT's public corrections page is a cited advantage in the Scope. But AP has one, WSJ has one, Reuters has one; the pattern generalizes.

Suggested reframe: **`media-claim-audit`** or **`press-primary-source-check`** — same 4-part body template, same vocabulary discipline, same Framing/Example structure, but the outlet is a field in each signal rather than the beat identity. Signals then get filterable by outlet without the beat itself carrying the adversarial framing.

Counter-consideration I don't want to dismiss: NYT-specific *might* attract more focused filings vs. a general media beat which could pull low-effort "outlet-X mildly wrong about topic-Y" filings. If the tighter frame is protective against that, keeping nyt-watch as v0 and generalizing later once the signal pattern proves out is a valid other path.

Ship blocker level: none. Both scopes work. This is scope-choice input, not a blocker on the skill file quality (which reads clean — the Avoid list is disciplined, the four-part body template is well-defined, the two-step sign-flow in the script is the right shape for a POST /api/beats bootstrap).

The one nit if you keep it NYT-specific: consider explicitly excluding NYT Op-Ed section in Scope > Does Not Cover, since some pieces there (e.g., long-form guest essays with factual claims) are ambiguous vs. "Opinion/editorial with no checkable factual claim."

PR #853, opened 2026-07-06
perf(init): global KV cache in front of /init DO call

biwasxyz:
## Problem

The `news-singleton` Durable Object reads **~18.7B SQLite rows/month** — ~75% of the shared 25B/month included tier and ~97% of the account's total DO row reads. Traced to a single hot path:

```
GET /  ·  GET /api/init  ·  client on load
        └─► edgeCacheMatch (Workers Cache API — PER-COLO) ──miss──► getInitBundle() ──► DO /init ──► queryLeaderboard(200)
```

The edge cache is **per-colo**: each of Cloudflare's ~300 data centers misses independently, and `getInitBundle()` had **no global cache** in front of it — so every per-colo miss recomputed the whole `/init` bundle on the singleton DO. That bundle's `queryLeaderboard()` scans the ~31k-row `signals` table several times per call (a `DISTINCT btc_address` since-epoch scan plus two 30-day `GROUP BY btc_address` re-scans), which indexes can't eliminate. Result: hundreds of thousands of full-table-scan `/init` calls per month.

This is Cloudflare's [#1 named Durable Objects anti-pattern](https://developers.cloudflare.com/durable-objects/best-practices/rules-of-durable-objects/) — a global singleton recomputing shared read state per request — and the docs' recommended remedy is exactly "optimize/cache queries."

## Fix

Wrap `getInitBundle()` in a **global KV cache** (`init-bundle:v1`, 60s TTL — the KV minimum). KV is global, so all per-colo edge misses collapse into **~1 DO recompute per 60s window worldwide**. The edge cache still serves the bulk of traffic; KV only catches the misses that previously fell straight through to the DO.

- **Best-effort** — any KV read/write failure is logged and falls through to a direct DO fetch, so the endpoint can never break on a cache blip.
- **Zero added latency** — the miss-path cache write is handed to `waitUntil()`.
- **Deploy-safe** — the key carries a `v1` version token; bump it on any `InitBundle` shape change so a stale payload can't be served across a deploy (60s TTL bounds this regardless).

## Expected impact

| Metric | Before | After |
|---|---|---|
| `/init` DO calls / mo | ~470k | ~44k (**~10×**) |
| DO rows read / mo | ~18.7B (75% of tier) | <~2B (comfortably in tier) |
| Added cost | — | 1 KV read per edge-miss (negligible) |

Staleness ceiling rises by ≤60s on a bundle already served with `s-maxage` in the minutes range — no UX change.

## Verification

- `npm run typecheck` ✅
- `biome lint` on changed files — clean ✅
- `npm test` — **427/427 pass** ✅
- `wrangler deploy --dry-run` — bundles, `NEWS_KV` bound ✅

## Follow-up (not in this PR)

Fix #2 — materialize the leaderboard like `correspondent_stats` already is (recompute on write / periodic alarm, read ~200 precomputed rows) — would cut the per-call cost too, driving rows read down further. Happy to open it next.

### DIFF
--- a/src/lib/do-client.ts
+++ b/src/lib/do-client.ts
@@ -241,13 +241,72 @@ export interface InitBundle {
   signalsCount1h: number;
 }
 
-/** Fetch all initial page load data in a single DO round-trip. */
-export async function getInitBundle(env: Env): Promise<InitBundle> {
+/**
+ * Global KV cache in front of the `/init` DO call.
+ *
+ * `/api/init` is edge-cached (Workers Cache API), but that cache is *per-colo*
+ * — each of Cloudflare's ~300 data centers misses independently, and every
+ * miss recomputes the `/init` bundle on the singleton DO. The bundle's
+ * `queryLeaderboard()` scans the full `signals` table several times per call,
+ * so those per-colo misses dominated the DO's rows-read usage.
+ *
+ * KV is a *global* store, so a short-TTL entry collapses all per-colo edge
+ * misses into ~1 DO recompute per TTL window worldwide. The edge cache still
+ * serves the vast majority of requests; KV only catches the edge misses that
+ * previously fell straight through to the DO.
+ *
+ * TTL is 60s — the KV minimum, and well within the tolerance of a homepage
+ * bundle that is already served with `s-maxage` in the minutes range. The key
+ * carries a version token so a shape change to `InitBundle` cannot serve a
+ * stale, incompatible payload across a deploy (bump on any shape change).
+ */
+const INIT_BUNDLE_KV_KEY = "init-bundle:v1";
+const INIT_BUNDLE_KV_TTL_SECONDS = 60;
+
+/** Fetch the `/init` bundle straight from the DO, bypassing the KV cache. */
+async function fetchInitBundleFromDO(env: Env): Promise<InitBundle> {
   const stub = getStub(env);
   const result = await doFetch<InitBundle>(stub, "/init");
   if (!result.ok) throw new Error(result.error ?? "Failed to get init bundle");
   if (result.data === undefined) throw new Error("Missing data in response");
   return result.data;
+}
+
+/**
+ * Fetch all initial page load data in a single DO round-trip, served through a
+ * global KV cache (see above). Pass `waitUntil` (e.g.
+ * `(p) => c.executionCtx.waitUntil(p)`) so the cache write on a miss does not
+ * add latency to the response; without it the write is awaited inline.
+ *
+ * All KV access is best-effort: any cache read/write failure falls through to
+ * a direct DO fetch so the endpoint never breaks on a cache blip.
+ */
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Adds a global KV cache in front of the `/init` DO call to collapse per-colo edge-cache misses into a single DO recompute per TTL window. This maps directly onto a cost driver we track operationally — DO row-reads from `/init`'s `queryLeaderboard()` scan are one of our biggest usage lines, so this is a well-targeted fix.

**What works well:**
- The doc comment on `getInitBundle` (`src/lib/do-client.ts`) clearly explains *why* KV sits in front of an already-edge-cached endpoint (per-colo vs. global cache scope) — this is exactly the kind of context that saves the next reader from re-deriving it.
- KV read/write failures are fully non-blocking: a read error falls through to the DO, a write error is logged and swallowed. The endpoint can't break on a KV blip.
- The `waitUntil` param is optional and defaults to inline await, so the function signature change doesn't silently alter behavior for any caller that doesn't pass it (there's only the one caller in `src/routes/init.ts`, and it's updated correctly).
- 60s TTL matches the KV platform minimum and is explicitly justified against the endpoint's existing `s-maxage` tolerance.

**[suggestion] Possible thundering herd across colos within a TTL window** (`src/lib/do-client.ts:295-317`)
Multiple colos can each miss KV within the same ~60s window before the first successful write lands, each independently calling `fetchInitBundleFromDO`. `src/lib/edge-cache.ts` already has a stale-while-revalidate pattern with a KV lock for exactly this kind of stampede risk on expensive DO-backed payloads. Given `/init` is a homepage hot path, it might be worth reusing that lock pattern here instead of a bare get/put — though since every miss currently hits the DO today, this PR is already a strict improvement even without it, so I wouldn't block on it.

**[question] Manual version-token bump** (`src/lib/do-client.ts:266`)
`INIT_BUNDLE_KV_KEY = "init-bundle:v1"` is a nice safeguard against serving a stale/incompatible shape across a deploy, but it relies on someone remembering to bump `v1` whenever `InitBundle`'s shape changes. Given the 60s TTL, worst case is a brief window of a shape mismatch if someone forgets — probably fine, but flagging in case there's an appetite for deriving the version from a schema hash instead of a manually-maintained literal.

**Code quality notes:**
- Splitting `fetchInitBundleFromDO` out from `getInitBundle` is a clean way to keep the original DO-fetch logic testable/callable directly, and reads well alongside the new caching wrapper.
- No test coverage added for the new cache-hit/cache-miss/cache-failure paths, though the repo doesn't appear to have existing tests for `do-client.ts` either, so this isn't a new gap introduced by this PR.

**Operational context:** we've flagged DO row-reads as a dominant cost driver in our own monitoring (SQLite-backed DOs, cursor discipline on read-heavy sensors) — this change is squarely in line with that concern and should meaningfully cut `/init`'s contribution to it.

Nice fix — approving.

PR #854, opened 2026-07-06
perf(leaderboard): materialise top-N into a read-through cache

biwasxyz:
Follow-up to #853 (fix #2 of the DO rows-read reduction). Independent of #853 — that PR caches the worker→DO `/init` call; this one cuts the cost *inside* the DO.

## Problem

`queryLeaderboard()` scans the ~30k-row `signals` table **several times per call** — a `DISTINCT btc_address` scan since the scoring epoch, plus two 30-day `GROUP BY btc_address` re-scans (signal count + distinct active days), joined against brief_signals/corrections/referrals/earnings. Indexes can't eliminate these aggregate scans.

It ran on **every** `/init`, `/leaderboard`, and `/correspondents-bundle` request — the single largest contributor to the `news-singleton` DO's ~18.7B rows read/month (~75% of the shared 25B/month included tier).

## Fix

A read-through cache (`getLeaderboardCached`) backed by a single-row `leaderboard_cache` table:

- The three hot **top-200** read paths serve from the materialised snapshot; the expensive scan runs **at most once per 60s** instead of per request.
- Callers needing exact, unbounded results — weekly payouts, `/leaderboard/verify/:address`, reset snapshots (all `limit = 10_000`) — keep calling `queryLeaderboard()` directly (`limit > CACHE_SIZE` bypasses the cache).
- **Staleness bounded by the 60s TTL** — acceptable for a leaderboard already served to clients with minute-scale `s-maxage`.
- **Best-effort** — any cache read/parse/write failure falls back to a live query, so the leaderboard can never break on a cache blip.
- `leaderboard_cache` lives in the always-re-applied base schema (pure cache, no durable data) so it **self-heals on cold start** — no migration bump.

## Invalidation

- `/leaderboard/reset` clears the cache so a post-reset read never serves pre-reset scores.
- `/test-seed` clears it too — bulk seeding bypasses the normal write path, mirroring the existing `recomputeCorrespondentStatsFor` refresh right beside it. Everyday mutations rely on the TTL (bounded 60s staleness by design).

## Impact

| | Before | After |
|---|---|---|
| Rows read per hot leaderboard read | ~30–50k (3× signals scans) | ~1 cached row (fresh hit) |
| Leaderboard recompute frequency | every `/init` · `/leaderboard` · `/correspondents-bundle` | ≤ 1 / 60s |

Combined with #853 (fewer `/init` recomputes) this takes the worker's DO rows read from ~75% of the 25B/mo included tier to comfortably inside it, with years of headroom.

## Verification

- `npm run typecheck` ✅
- `biome lint` on changed files — clean (no new warnings) ✅
- `npm test` — **427/427 pass** ✅, including `scoring-math` and `home-page`, which exercise `/leaderboard` and `/init` through the real DO (miniflare) via the new cached path

Note: scoring-math tests initially failed on stale cross-test cache reads; resolved correctly by the `/test-seed` invalidation (the seed path is responsible for refreshing materialised views, as it already does for `correspondent_stats`).

### DIFF
--- a/src/objects/news-do.ts
+++ b/src/objects/news-do.ts
@@ -314,6 +314,13 @@ function rowToSignal(row: Record<string, unknown>): Signal {
 const FRONT_PAGE_WINDOW_SQL = "-2 days";
 const FRONT_PAGE_MAX_ROWS = 200;
 
+// Leaderboard materialisation — see getLeaderboardCached(). CACHE_SIZE is the
+// top-N stored; all hot read paths request ≤ this. TTL bounds staleness (and
+// therefore how often the expensive full-table scan runs) to one recompute per
+// window per read burst.
+const LEADERBOARD_CACHE_SIZE = 200;
+const LEADERBOARD_CACHE_TTL_MS = 60_000;
+
 // ISO 8601 'now' for SQLite comparators against ISO-stored timestamp columns
 // (e.g. classifieds.expires_at). Default datetime('now') returns
 // 'YYYY-MM-DD HH:MM:SS' (no T, no Z), and a same-day lex compare puts 'T'
@@ -5369,7 +5376,7 @@ export class NewsDO extends DurableObject<Env> {
       // doesn't break the correspondents endpoint (preserves old .catch(() => []) behavior)
       let leaderboard: Array<Record<string, unknown>> = [];
       try {
-        leaderboard = this.queryLeaderboard(200);
+        leaderboard = this.getLeaderboardCached(200);
       } catch (e) {
         console.error("Leaderboard query failed in correspondents bundle:", e);
       }
@@ -5500,7 +5507,7 @@ export class NewsDO extends DurableObject<Env> {
       // Leaderboard — wrapped in try/catch to preserve partial results on failure
       let leaderboard: Array<Record<string, unknown>> = [];
       try {
-        leaderboard = this.queryLeaderboard(200);
+        leaderboard = this.getLeaderboardCached(200);
       } catch (e) {
         console.error("Leaderboard query failed in init bundle:", e);
       }
@@ -5552,7 +5559,7 @@ export class NewsDO extends DurableObject<Env> {
 
     // GET /leaderboard — weighted scores across all roles
     this.router.get("/leaderboard", (c) => {
-      const rows = this.queryLeaderboard(200);
+      const rows = this.getLeaderboardCached(200);
       return c.json({ ok: true, data: rows } satisfies DOResult<unknown[]>);
     });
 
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Materialises the top-N leaderboard into a single cached row so the three hot read paths (`/init`, `/leaderboard`, `/correspondents-bundle`) stop re-running the full signals-table scan on every request.

**What works well:**
- Clean read-through cache design: `getLeaderboardCached()` bypasses to `queryLeaderboard()` directly whenever `limit > CACHE_SIZE`, so the exact-results callers (weekly payouts, `/leaderboard/verify/:address`, reset snapshots) are untouched and keep serving unbounded, non-cached data — no accidental truncation risk there.
- Best-effort fallback is done right: cache read, JSON parse, and cache write are each wrapped so a corrupted row or a storage hiccup degrades to a live query instead of a broken leaderboard. That's the correct failure mode for a pure-cache table.
- `computed_at` is stamped with `new Date(now).toISOString()` (JS `Date`, not SQLite `datetime('now')`) — that sidesteps a UTC-vs-local parsing skew bug we've hit in our own code before (naive `datetime('now')` has no `Z` suffix and gets misparsed as local time). Good instinct keeping the timestamp source consistent between write and read.
- Invalidation on `/leaderboard/reset` and `/test-seed` is targeted correctly — both are the paths that bypass the normal write flow, mirroring the existing `recomputeCorrespondentStatsFor` refresh pattern right next to it.
- Schema change lives in the always-re-applied base schema (no migration bump) since it's pure cache — self-heals on cold start, consistent with how this repo already treats similar tables.
- Tests (427/427), typecheck, lint, and `wrangler deploy --dry-run` all clean; the PR description is honest about the one hiccup encountered (stale cross-test cache reads) and how it was resolved.

**[question] Everyday mutation paths rely on TTL only**
New signals, corrections, referrals, and earnings recorded through normal flows don't invalidate the cache — they rely on the 60s TTL to eventually reflect. That's called out explicitly in the PR description as an intentional tradeoff, and it's a reasonable one for a leaderboard already served with minute-scale `s-maxage`. Just confirming this is the intended contract going forward (rather than something to tighten later) — worth a one-line comment on `LEADERBOARD_CACHE_TTL_MS` itself if not already documented elsewhere, since it's easy for a future editor to add a new scoring-mutation endpoint without realizing it needs a cache-clear too, the same way `/leaderboard/reset` and `/test-seed` needed it here.

**[nit]** `age >= 0` guards against clock skew (computed_at appearing in the future) by treating it as stale rather than throwing — good defensive touch, no change needed, just noting it as intentional in case a future reader assumes it's dead code.

**Code quality notes:** No reuse or simplification issues found — the cache is scoped tightly to the one function that needs it, and the two invalidation call sites are minimal, targeted additions next to the code they mirror.

*Operational note:* We've hit the same DO-rows-read ceiling class of problem in our own sensors — cursoring against SQLite-backed Durable Objects to avoid full-table rescans on a 1-minute cadence. This PR's shape (bounded TTL cache with defensive fallback) is the same pattern we'd reach for; nothing here raises a flag from that experience.

PR #857, opened 2026-07-06
perf(status): drop member fan-out in /status beat-activity query

biwasxyz:
The real DO rows-read hog, found by ranking **live production** DO endpoints by CPU (via `wrangler tail`): **`/status/:address`** — ~25 calls/min, uncached, dominating total CPU. (The earlier leaderboard PRs #853/#854/#856 were verified working — `/init` no longer hits the DO, `/leaderboard` is ~1 CPU/call — but the leaderboard was not the live hot path.)

## Root cause

`/status/:address` computed each of the agent's beats' last-activity timestamp with a member fan-out:

```sql
SELECT b.*, MAX(s.created_at) as last_signal_at
FROM beat_claims bc
JOIN beats b ON bc.beat_slug = b.slug
LEFT JOIN beat_claims bc_all ON b.slug = bc_all.beat_slug AND bc_all.status = 'active'
LEFT JOIN signals s ON bc_all.btc_address = s.btc_address AND s.beat_slug = b.slug ...
WHERE bc.btc_address = ? ... GROUP BY b.slug
```

It joined **every active member of every beat** (~250/beat across 13 beats) to their signals just to derive one MAX per beat — O(members × signals) rows per call. Uncached at ~25/min, this was the dominant read.

## Fix

Per-beat correlated MAX backed by the existing `idx_signals_beat_created (beat_slug, created_at DESC)` — ~1 row per beat:

```sql
SELECT b.*, (
  SELECT MAX(s.created_at) FROM signals s
  WHERE s.beat_slug = b.slug AND s.correction_of IS NULL
) AS last_signal_at
FROM beat_claims bc JOIN beats b ON bc.beat_slug = b.slug
WHERE bc.btc_address = ? AND bc.status = 'active' ORDER BY bc.claimed_at
```

## Measured reduction (not modelled)

New regression test seeds a 150-member / 2,250-signal beat and reads actual `cursor.rowsRead` for both forms, with `ANALYZE` so the planner matches production:

| | rows read |
|---|---|
| old (member fan-out) | **4,803** |
| new (per-beat MAX) | **5** |

~960× fewer rows for this query. The test asserts the new form's cost does **not** scale with signal count (guards against regression).

## Semantic note

`last_signal_at` is now "last signal in the beat" (any author) rather than "last signal by a still-active member". Signals carry `beat_slug`, so for beat activity/expiry these agree except when the single most-recent signal in a beat is from a since-departed member — where counting it as recent activity is at least as correct. Same-result assertion is in the test.

## Follow-ups (same pattern, not in this PR)

`/beats` and the `/init` beats block use the identical member-fan-out MAX and should get the same rewrite for consistency and further savings. `/status` has no worker-layer cache, which is why it dominated — worth considering a short KV cache there too.

## Verification
- `npm test` — **428/428** (new test included) ✅
- `biome lint` on changed files — clean ✅
- Post-deploy: I'll re-rank DO endpoints by CPU via tail to confirm `/status` drops.

### DIFF
--- a/src/__tests__/status-query-efficiency.test.ts
+++ b/src/__tests__/status-query-efficiency.test.ts
@@ -1,0 +1,131 @@
+/**
+ * Regression guard for the /status/:address beat-activity query.
+ *
+ * The beats-membership query computes each beat's last-activity timestamp. The
+ * original form joined every active member of every beat to their signals purely
+ * to derive one MAX per beat, so it read O(members × signals) rows per call — at
+ * ~25 uncached /status calls/min this was the single largest driver of the DO's
+ * SQLite rows-read bill. The current form uses a per-beat correlated MAX backed
+ * by idx_signals_beat_created, reading ~O(beats) rows.
+ *
+ * This test seeds a realistically large beat and asserts the live query reads a
+ * bounded number of rows regardless of how many members/signals the beat has.
+ */
+import { env, runInDurableObject } from "cloudflare:test";
+import { describe, it, expect } from "vitest";
+import type { NewsDO } from "../objects/news-do";
+
+describe("/status beat-activity query efficiency", () => {
+  it("reads O(beats) rows, not O(members × signals)", async () => {
+    const id = env.NEWS_DO.idFromName("news-singleton");
+    const stub = env.NEWS_DO.get(id);
+
+    const result = await runInDurableObject(stub, (_instance: NewsDO, state) => {
+      const sql = state.storage.sql;
+      const now = new Date().toISOString();
+      const beat = "measure-beat";
+      const me = "bc1qme000000000000000000000000000000000000";
+      const MEMBERS = 150;
+      const PER = 15; // signals per member
+
+      sql.exec(
+        "INSERT OR IGNORE INTO beats (slug, name, created_by, created_at, updated_at) VALUES (?, ?, ?, ?, ?)",
+        beat,
+        "Measure Beat",
+        "creator",
+        now,
+        now
+      );
+
+      for (let m = 0; m < MEMBERS; m++) {
+        const addr = m === 0 ? me : `bc1qmember${String(m).padStart(31, "0")}`;
+        sql.exec(
+          "INSERT OR IGNORE INTO beat_claims (beat_slug, btc_address, claimed_at, status) VALUES (?, ?, ?, 'active')",
+          beat,
+          addr,
+          now
+        );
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Nice catch — we've felt this DO's rows-read bill in our own sensor cadence, so a 960x reduction on the highest-traffic endpoint is a real win.

**What works well:**
- The correlated per-beat `MAX(s.created_at)` subquery is exactly the right shape for this: it drops the `bc_all`/`signals` fan-out join entirely and rides `idx_signals_beat_created (beat_slug, created_at DESC)`, which already exists in `schema.ts:132` — confirmed the index is live, so this isn't waiting on a migration.
- The regression test in `status-query-efficiency.test.ts` is a good addition, not just a correctness check but a cost-shape guard (`rowsRead` assertions) that will catch a future fan-out regression before it hits production traffic again.
- The commit message and inline comment both call out the semantic shift precisely (last-signal-in-beat vs last-signal-by-active-member) rather than glossing over it.

**[question] Semantic shift affects a user-visible field**
`beatStatus` (active/inactive) in the `/status` response is derived directly from `last_signal_at` (`news-do.ts:3967-3975`), so this isn't just an internal accounting change — a beat can now flip to "active" based on a signal from a departed member, and vice versa (a beat with an active member but no member-authored signals could show stale sooner under the old semantics, or fresher now). Given the "at least as correct" framing in the commit, this reads like an acceptable tradeoff, but worth a one-line confirmation that no downstream consumer (Whop bot, digest, etc.) branches on `beatStatus` in a way that's sensitive to this edge case.

**[suggestion] Follow-up scope note**
The PR correctly flags that `/beats` and the `/init` beats block still use the old member-fan-out join (confirmed both still present in `news-do.ts` at the other `LEFT JOIN beat_claims bc_all`-style sites). Worth a tracking issue so this doesn't get lost — `/beats` in particular sounds like it could see similar per-call cost if it's called at comparable frequency to `/status`.

**Code quality notes:** No reuse or over-engineering concerns — the rewrite is a direct, minimal diff (28 lines) that removes complexity rather than adding it. The test's OLD-vs-NEW dual-query approach adds a bit of duplication but it's justified: it's the only way to assert the *relative* improvement, not just an absolute bound.

*Operational context:* We've seen similar member-fan-out patterns cause SQLite DO rows-read spikes at 1-minute sensor cadence on our own side — this is the right fix pattern (correlated subquery over an existing composite index) for that class of problem.

PR #858, opened 2026-07-06
perf(beats): drop member fan-out in beat-activity queries

biwasxyz:
Follow-up to #857. That PR fixed the `/status` beat-activity fan-out; this removes the **same copy-pasted pattern** from every other place it lived.

## The pattern

Each beat-activity read computed "last signal in each beat" by joining **all active members of the beat to their signals** just to derive one `MAX(created_at)` per beat — O(members × signals) rows (~250 members/beat, ~31k signals). Sites:

| Handler | Hot? | Cache |
|---|---|---|
| `/beats` (list) | yes | per-colo edge only |
| `/correspondents-bundle` | medium | per-colo SWR (heaviest single call, ~181 CPU) |
| `/beats/:slug` | medium | per-colo edge |
| `/init` beats block | — | global KV (already off the DB most of the time) |
| `POST /beats` (reclaim check) | rare | write path |

## Fix

Per-beat correlated `MAX` backed by the existing `idx_signals_beat_created (beat_slug, created_at DESC)` — ~1 row per beat. Removing the joins also drops the now-unnecessary `GROUP BY`. Identical transform to #857.

**Consistency bonus:** all beat-activity reads (including `/status`) now share one beat-level "last signal in the beat" definition, so the homepage, `/beats`, and `/status` can no longer report different activity/expiry for the same beat.

## Measured (regression test, actual `rowsRead`)

| query | before | after |
|---|---|---|
| all-beats list (4 beats / 1,920 signals) | **4,989** | **41** (~120×) |
| `/status` shape (unchanged, still guarded) | 4,803 | 5 |

Both tests assert the cost does **not** scale with signal count.

## Semantic note
`last_signal_at` is now "last signal in the beat" (any author) rather than "last signal by a still-active member" — same as #857. Signals carry `beat_slug`, so these agree for activity/expiry except when the most-recent signal in a beat is from a since-departed member, where counting it is at least as correct. This now matches `/status`, removing a latent cross-page inconsistency.

## Verification
- `npm test` — **429/429** (2 efficiency tests) ✅
- `biome lint` — clean ✅
- Post-deploy: re-rank DO endpoints by CPU via tail to confirm `/beats` + `/correspondents-bundle` drop.

## Remaining follow-ups (from the audit, not in this PR)
- `/classifieds/rotation` `ORDER BY RANDOM()` (uncached, most-frequent endpoint — latent hog once ad volume grows)
- `/report` full-table scans (serve from materialized `correspondent_stats`)

### DIFF
--- a/src/__tests__/status-query-efficiency.test.ts
+++ b/src/__tests__/status-query-efficiency.test.ts
@@ -141,4 +141,89 @@ describe("/status beat-activity query efficiency", () => {
     // signal fan-out is gone.
     expect(result.newRows).toBeLessThan(result.signals / 10);
   });
+
+  // Same fan-out lived in the "all beats" list query used by /beats,
+  // /correspondents-bundle and the /init beats block. This guards that rewrite.
+  it("all-beats list reads O(beats), not O(members × signals)", async () => {
+    const id = testEnv.NEWS_DO.idFromName("news-singleton");
+    const stub = testEnv.NEWS_DO.get(id);
+
+    const result = (await runInDurableObject(stub, (_instance, state) => {
+      const sql = state.storage.sql;
+      const now = new Date().toISOString();
+      const BEATS = 4;
+      const MEMBERS = 60;
+      const PER = 8;
+
+      for (let b = 0; b < BEATS; b++) {
+        const beat = `list-beat-${b}`;
+        sql.exec(
+          "INSERT OR IGNORE INTO beats (slug, name, created_by, created_at, updated_at) VALUES (?, ?, ?, ?, ?)",
+          beat,
+          `List Beat ${b}`,
+          "creator",
+          now,
+          now
+        );
+        for (let m = 0; m < MEMBERS; m++) {
+          const addr = `bc1qlist${b}member${String(m).padStart(28, "0")}`;
+          sql.exec(
+            "INSERT OR IGNORE INTO beat_claims (beat_slug, btc_address, claimed_at, status) VALUES (?, ?, ?, 'active')",
+            beat,
+            addr,
+            now
+          );
+          for (let s = 0; s < PER; s++) {
+            sql.exec(
+              `INSERT OR IGNORE INTO signals
+                 (id, beat_slug, btc_address, headline, sources, created_at, updated_at, status, correction_of)
+               VALUES (?, ?, ?, ?, '[]', ?, ?, 'approved', NULL)`,
+              `lsig-${b}-${m}-${s}`,
+              beat,
+              addr,
+              `h ${b}-${m}-${s}`,
+              new Date(Date.now() - (m * PER + s) * 60_000).toISOString(),
+              now
+            );
+          }
+        }
+      }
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Removes the beats × active-members × signals fan-out from `/beats`, `/beats/:slug`, the `POST /beats` reclaim check, `/correspondents-bundle`, and `/init` — the same rewrite #857 applied to `/status`, just finishing the job everywhere the pattern was copy-pasted.

**What works well:**
- Correlated `SELECT MAX(s.created_at) FROM signals s WHERE s.beat_slug = b.slug AND s.correction_of IS NULL` reuses the existing `idx_signals_beat_created (beat_slug, created_at DESC)` index — no new index needed, and the measured numbers back it up (4,989 → 41 rows read on the all-beats list with 1,920 seeded signals).
- Dropping the `GROUP BY` along with the join is the right call once the join is gone — no dead aggregation left behind.
- New regression test (`all-beats list reads O(beats), not O(members × signals)`) mirrors the existing `/status` test's structure and asserts both a relative bound (`< oldRows/5`) and an absolute one that scales with seeded signal count — good, this catches a regression even if someone reintroduces a per-member join later.
- All five call sites got the identical transform — no site was missed (I checked `/beats`, `/beats/:slug`, the reclaim-check in `POST /beats`, `/correspondents-bundle`, and `/init` against the diff).

**Correctness check on the semantic note in the PR description:**
You flagged that `last_signal_at` changes from "last signal by a still-active member" to "last signal in the beat, any author." I traced this against the current schema/code and it's a no-op in practice, not just "close enough": `beat_claims.status` is only ever inserted/upserted as `'active'` (line ~2299-2301) — there's no code path in `news-do.ts` that ever sets a `beat_claims` row to a non-active status (unlike `beat_editors`, which does have a real active→inactive transition). Signal creation also hard-requires an active claim to exist at write time (line ~2864-2876). So every signal's `(beat_slug, btc_address)` always has a matching active claim, permanently — the old join was filtering against a condition that can never be false. This rewrite is a true equivalence in the current codebase, not an approximation, which is worth stating more confidently than "at least as correct" in the PR body. If `beat_claims` ever grows a real deactivation path in the future, this is the query to revisit.

**[question]** The PR mentions `/report` full-table scans and `/classifieds/rotation` `ORDER BY RANDOM()` as known follow-ups not included here — are those tracked as separate issues yet, or should I expect a #859?

**Code quality notes:**
- No reuse or simplification concerns — the five sites are mechanically identical edits, and keeping them as repeated inline SQL (rather than extracting a shared query builder) is reasonable here since Durable Object SQL call sites in this file are already written this way throughout.

**Operational context:**
We don't run this DO ourselves, but we do read `/beats` and `/correspondents-bundle` from Arc's sensors on a schedule — this should show up as lower CPU/rows-read on our side too once deployed. No concerns from our end about downstream consumers of `last_signal_at`; we only use it for display/staleness checks, not correctness-critical logic.

Approving — the fix is correct, well-tested, and matches the precedent set by #857.

PR #859, opened 2026-07-06
perf(beats): fetch member roster only on ?include=members

biwasxyz:
Follow-up to the beat-activity fan-out fixes (#857/#858). After those, the remaining cost of the hot `/beats` endpoint was a second query that **always fetched every active claim across all beats** (~thousands of rows) and materialised them into JS objects — just so the default `/api/beats` could report `memberCount`. The full roster is only needed by `?include=members` (Bureau page) and `/api/leaderboard`.

## Change
- **Default path**: per-beat `COUNT(*) GROUP BY beat_slug` (index-only via `idx_beat_claims_status_beat_address`), returned as `member_count`.
- **Full roster**: fetched only when `?include=members` is passed.
- `do-client.listBeats(env, includeMembers=false)` — `/api/leaderboard` opts in (it needs per-member addresses); skills + sitemap keep the cheap default.

## Measured (regression test, actual `rowsRead`)
| query | rows read |
|---|---|
| full-roster fetch (600-claim seed) | 1,200 |
| count `GROUP BY` (default) | 600 |

~2× fewer rows, plus it skips materialising every claim row into JS/JSON on a hot endpoint. A second test guards the contract (`member_count` on default, `members` array on `?include=members`).

## Confirmed against Cloudflare docs
The DO storage docs state indexes "reduce how much data is scanned … reduction in rows read." The count path is served index-only by `idx_beat_claims_status_beat_address (status, beat_slug, …)`, consistent with that guidance — and the `rowsRead` measurement above confirms the actual reduction rather than assuming it.

## Verification
- `npm test` — **431/431** (2 new roster tests) ✅
- `biome lint` — clean ✅

## Note
This is a rows-read + CPU/payload trim (~2×), not the ~100× of the fan-out fixes — `COUNT … GROUP BY` still reads one index row per active member. Making it O(beats) would need a materialized `member_count` column maintained on join/leave; deferred as a bigger change.

### DIFF
--- a/src/__tests__/beats-roster-gating.test.ts
+++ b/src/__tests__/beats-roster-gating.test.ts
@@ -1,0 +1,105 @@
+/**
+ * Guards the /beats member-roster gating.
+ *
+ * The DO /beats handler fetches the full active-member roster (thousands of rows
+ * across all beats) only when ?include=members is requested; the default path
+ * returns per-beat `member_count` instead, so the hot /api/beats endpoint does
+ * not materialise every claim just to report a count.
+ */
+import { env, runInDurableObject } from "cloudflare:test";
+import { describe, it, expect } from "vitest";
+import type { Env } from "../lib/types";
+
+const testEnv = env as unknown as Env;
+
+interface BeatRow {
+  slug: string;
+  member_count?: number;
+  members?: unknown[];
+}
+
+describe("/beats member-roster gating", () => {
+  it("default returns member_count only; include=members returns the roster", async () => {
+    const id = testEnv.NEWS_DO.idFromName("news-singleton");
+    const stub = testEnv.NEWS_DO.get(id);
+    const MEMBERS = 7;
+
+    await runInDurableObject(stub, (_instance, state) => {
+      const sql = state.storage.sql;
+      const now = new Date().toISOString();
+      sql.exec(
+        "INSERT OR IGNORE INTO beats (slug, name, created_by, created_at, updated_at) VALUES ('roster-beat', 'Roster', 'creator', ?, ?)",
+        now,
+        now
+      );
+      for (let m = 0; m < MEMBERS; m++) {
+        sql.exec(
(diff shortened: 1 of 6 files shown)


arc0btc (approved):
Good follow-up to #857/#858 — trims the last hot-path cost on `/beats` by gating the full member-roster fetch behind `?include=members` and reporting a cheap per-beat `member_count` otherwise.

**What works well:**
- The `GROUP BY beat_slug` count query is index-only via `idx_beat_claims_status_beat_address`, and the new regression test asserts `countRows < fullRows` using the real `rowsRead` counter rather than assuming the win — that's the right way to validate a "fewer rows read" claim.
- Checked all four `listBeats()` call sites (`beats.ts`, `leaderboard.ts`, `skills.ts`, `seo.ts`, plus `getActiveBeatSlugs`) — only `leaderboard.ts` needs the full roster (for `buildBeatsByAddress`) and it correctly opts in with `listBeats(c.env, true)`. The other three only ever touched `slug`/`status`/`name`/`description`, so the default-to-count change is behavior-preserving for them.
- `member_count` defaults to `0` via `countByBeat.get(slug) ?? 0` for beats with zero active claims (excluded from the `GROUP BY` result set) — matches the old `claimsByBeat.get(slug) ?? []` `.length` semantics.
- Edge-cache key already varies on the full request URL, so `?include=members` and the default path get independent cache entries — no cache-poisoning risk from serving one variant's response for the other.
- Honest PR description: calls out this is a ~2x rows-read/CPU/payload trim, not the ~100x of the earlier fan-out fix, and explains why (`COUNT ... GROUP BY` still reads one index row per active member) rather than overselling it.

**Code quality notes:**
- Nit only: in the `!includeMembers` branch of `news-do.ts`, `claimsByBeat` is declared and never populated (only used inside the `includeMembers` branch) — harmless, just a slightly wider variable scope than needed. Not blocking.

No functional, security, or architectural concerns. Tests (431/431 incl. 2 new), lint, and dry-run deploy all reported green.

PR #860, opened 2026-07-08
fix(identity-gate): negative-cache fail-closed result to stop timeout storms (#826)

biwasxyz:
Closes #826.

## Problem
`checkAgentIdentity` fails closed (`shouldBlock=true`) when the identity API is unreachable, but did **not cache** that result. During a sustained outage every incoming request re-ran the full initial+retry fetch (up to ~6s of hung sockets each), producing the repeated timeout storms reported in #826.

## Fix
Negative-cache the block for a short window (60s — Cloudflare KV's minimum `expirationTtl`). This collapses the storm to at most one probe per address per minute while submissions still unblock within a minute of the API recovering. Happy-path and 404 caching are unchanged.

## Verification
- Rewrote the stale "does not cache on failure" unit test to assert the new negative-cache behavior (block is cached with a 60s TTL; a follow-up request within the window hits cache instead of re-storming).
- Full suite green (433 tests); `tsc --noEmit` clean.

## Notes
- 60s is the floor KV allows; shorter TTLs are rejected at runtime.
- Confirmed live in code before fixing: the fail-closed branch returned without any `kv.put`.

### DIFF
--- a/src/__tests__/identity-gate.test.ts
+++ b/src/__tests__/identity-gate.test.ts
@@ -169,33 +169,43 @@ describe("checkAgentIdentity — fail closed on API errors", () => {
     expect(result.level).toBe(2);
   });
 
-  it("does not cache the result on API failure", async () => {
-    const kv = makeKV();
-    // Both attempts fail — fail-closed result must NOT be cached
+  it("negative-caches the block briefly so an outage doesn't storm the API (#826)", async () => {
+    const store = new Map<string, string>();
+    const putArgs: Array<{ key: string; options?: { expirationTtl?: number } }> = [];
+    const kv = {
+      async get(key: string) {
+        return store.get(key) ?? null;
+      },
+      async put(key: string, value: string, options?: { expirationTtl?: number }) {
+        store.set(key, value);
+        putArgs.push({ key, options });
+      },
+    } as unknown as KVNamespace;
+
+    // Both attempts fail — fail-closed result must be cached with a short TTL
     const firstSpy = vi
       .spyOn(globalThis, "fetch")
       .mockRejectedValueOnce(new Error("timeout"))
       .mockRejectedValueOnce(new Error("timeout"));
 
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Solid, targeted fix for #826. Confirmed the root cause matches the diff: the fail-closed branch in `checkAgentIdentity` built the `blocked` result but never wrote it to KV, so every request during an API outage re-ran the full 2×3s-timeout retry loop — that's the storm.

**What works well:**
- Reuses the existing cache-read path at the top of the function (`kv.get(cacheKey)` → parse → return) — no new branching needed on the read side, since the negative-cache result is just another `IdentityCheckResult` in the same shape as the happy-path cache.
- 60s TTL is a good choice: short enough that a real recovery unblocks quickly, and it's KV's floor anyway so there's no cheaper option.
- Test rewrite is thorough — it now asserts the actual `expirationTtl` value passed to `kv.put`, not just call counts, and explicitly checks that a follow-up request within the window skips `fetch` entirely. That's the exact regression this PR prevents from being silent.
- Comments on both the constant and the `kv.put` call explain the *why* (storm mechanics, KV minimum), not just what the line does.

**Operational context:** we've hit fail-closed identity gates from our own submission paths before — an unrecoverable-until-cache-expires block is the right tradeoff here since it caps worst-case blocked time at 60s, which lines up with how we'd want a downstream 503+Retry-After consumer to behave.

**[question]** Not blocking, but: if `kv.put` itself throws (KV write failure/timeout), the function will now throw instead of returning the `blocked` result, since the `await kv.put(...)` sits before `return blocked` with no try/catch. That's an existing pattern already present on the happy-path cache write above it, so it's not a regression introduced by this PR — just flagging in case it's worth a follow-up hardening pass across both write sites.

**Code quality notes:** nothing to add beyond the above — the change is minimal, reuses existing types (`IdentityCheckResult`), and doesn't touch the happy-path or 404 caching as claimed.

Nice clean fix for a real production issue. Approving.

PR #861, opened 2026-07-08
fix(status): surface active beat, don't report file-gate on retired beats (#832)

biwasxyz:
Closes #832.

## Problem
`GET /api/status/{address}` exposed the agent's **oldest** claim (`ORDER BY claimed_at`) as the singular `beat`/`beatStatus`. For the pre-consolidation cohort that oldest claim is a since-retired beat with no recent activity, so it read `beatStatus:"inactive"` while `canFileSignal:true` and the `file-signal` action were correct for the active beat they migrated to — the contradiction reported in #832 (reproduced live on `deal-flow`).

## Fix
- Prefer the agent's first non-retired **active** beat (then any non-retired beat) as the backward-compat `beat`, instead of the oldest claim. Keeps the `active|inactive` contract and avoids flapping `beatStatus` on cooldown (which sonic-mast's literal option-1 would have done).
- Set `canFileSignal=false` when **every** claimed beat is retired, matching `POST /api/signals` which rejects retired-beat filing, with a `claim-beat` action.
- Expose per-beat `retired` in the `beats` array.

## Verification
- New `status-beat-consistency.test.ts` seeds the exact reported scenario (retired beat claimed first + active beat migrated to) and asserts the exposed beat is the active one with a consistent gate; plus a retired-only agent gets `canFileSignal:false`.
- Full suite green (433 tests); `tsc --noEmit` clean.

### DIFF
--- a/src/__tests__/status-beat-consistency.test.ts
+++ b/src/__tests__/status-beat-consistency.test.ts
@@ -1,0 +1,115 @@
+/**
+ * Regression guard for #832 — /status/:address returned a contradictory
+ * `beatStatus: "inactive"` alongside `canFileSignal: true` for agents who
+ * registered on a since-retired beat but migrated to an active one.
+ *
+ * The backward-compat `beat`/`beatStatus` fields blindly exposed the agent's
+ * oldest claim (ORDER BY claimed_at), which for the pre-consolidation cohort was
+ * a retired beat with no recent activity — reading "inactive" while the filing
+ * gate said they could file. These tests assert the display field now tracks the
+ * agent's fileable beat, and that an all-retired agent is not told they can file.
+ */
+import { env, runInDurableObject } from "cloudflare:test";
+import { describe, it, expect } from "vitest";
+import type { Env } from "../lib/types";
+
+const testEnv = env as unknown as Env;
+
+interface StatusBody {
+  data: {
+    beat: { slug: string; beatStatus: string; retired: boolean } | null;
+    beatStatus: string | null;
+    canFileSignal: boolean;
+    beats: Array<{ slug: string; beatStatus: string; retired: boolean }>;
+    actions: Array<{ type: string; description: string }>;
+  };
+}
+
+async function seed(stub: DurableObjectStub, fn: (sql: SqlStorage) => void) {
+  await runInDurableObject(stub, (_instance, state) => fn(state.storage.sql));
+}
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Fixes the exact contradiction from #832 — an agent's oldest claim (often a retired pre-consolidation beat) was shown as the display `beat` even when `canFileSignal`/`file-signal` correctly pointed at the active beat they'd migrated to. Nice catch, and the fix lines up with what `POST /api/signals` already enforces (retired beats reject filing at src/objects/news-do.ts:2886-2888), so the status/gate contract stays consistent across endpoints.

**What works well:**
- Selection order (`active non-retired` → `any non-retired` → oldest fallback) fixes the display without touching the existing `beatStatus` active/inactive computation (still purely activity-based) — the two concepts (retired vs stale-but-not-retired) stay properly separate.
- `hasFileableBeat` gate matches the real constraint (retired beats can't accept new signals) rather than inferring it from `beatStatus`, which would have been wrong — a beat can be non-retired but "inactive" (no recent signals) and still fileable.
- Test coverage is exactly on point: seeds the reported repro (retired-first-claim + active-migrated-to) and a retired-only agent, both asserting the previously-contradictory fields.
- Exposing `retired` per-beat in the `beats` array is a good call — gives clients enough to build their own UI messaging without re-deriving it.

**[nit]** The final fallback in the `beat` selection is slightly redundant:
```suggestion
      const beat =
        agentBeats.find((b) => !b.retired && b.beatStatus === "active") ??
        agentBeats.find((b) => !b.retired) ??
        agentBeats[0] ??
        null;
```
`agentBeats.length > 0 ? agentBeats[0] : null` and `agentBeats[0] ?? null` are equivalent here since `find` already returns `undefined` when the array is empty — just a shorter way to say the same thing, not blocking.

**[question]** In the actions block, the new `!hasFileableBeat` branch is checked before the `signalsToday >= MAX_SIGNALS_PER_DAY` branch. For an agent who is both over the daily limit and retired-only, they'll now see "beat retired" instead of "daily limit reached." Given retired-only already means they can't file regardless of the daily count, this seems fine/arguably more useful — just flagging that the message priority changed in case that wasn't intentional.

Verification (433 tests green, `tsc --noEmit` clean) plus a diff that's fully scoped to the reported bug — no scope creep. Approving.

PR #864, opened 2026-07-13
feat(review): let editors override the auto-scorer quality_score (#810)

biwasxyz:
## What & why

Fixes #810. The signal `quality_score` is stamped once by the pure auto-scorer at insert and is otherwise **immutable** — the review PATCH only accepts `status`, `feedback`, `displace_signal_id`. So when an editor rejects a signal for fabricated sources (e.g. `quality_score=100` on three GitHub release URLs that all 404), the misleading score stays on the record and on `/api/signals`. Editors can void the *signal* but not the *score*.

This gives editors the one lever they were missing.

**Deliberately not** an automated URL-existence checker: source truth remains an editorial judgement and the system never makes an outbound fetch. (We explored a background `alarm()` verifier and ruled it out — kept scope on the human-in-the-loop fix.)

## Change

Optional `quality_score` field on `PATCH /api/signals/:id/review`:

| Case | Result |
|---|---|
| Field omitted | Unchanged — the no-override `UPDATE` path is byte-identical |
| `quality_score: 15` | Score replaced; original preserved in `score_breakdown.override` (`by`, `at`, `previous_score`, `reason`) for audit |
| Out of range / non-integer | `400` |
| Non-editor caller | `401/403` from the existing `verifyEditorOrPublisher` gate — unchanged |

- No migration — reuses the existing `quality_score` / `score_breakdown` columns.
- Validation runs at the edge (fail-fast, pre-auth) **and** in the DO handler (backstop).
- BIP-322 auth unaffected: `verifyAuth` signs over method + path, not the body.
- Override is orthogonal to `status` — no auto-reject, no queue-behavior change.

## Files
- `src/routes/signal-review.ts` — edge validation + pass-through
- `src/lib/do-client.ts` — `ReviewSignalInput.quality_score`
- `src/objects/news-do.ts` — DO handler: destructure/validate, widen existence SELECT, conditional UPDATE
- `src/lib/signal-scorer.ts` — new pure `withScoreOverride()` helper + `SignalScoreBreakdown.override` type

## Tests
- 7 edge-validation tests (range/type boundaries, omission)
- 5 provenance-merge tests (preserve axes, null/legacy-JSON degrade, re-override replaces not nests)
- Full suite: **444 pass**, typecheck clean

## Known gap
The DO write happy-path (persisting the new score) isn't covered by an integration test — the test env can't generate BIP-322 auth (same limitation noted in `retraction.test.ts:9-11`). Mitigated by extracting the risky JSON-merge logic into the fully-unit-tested `withScoreOverride()`. A `runInDurableObject` test that seeds a publisher and drives the DO router directly would close this — proposed as a follow-up.

## Not included (follow-up)
The editor/UI client needs a field added to actually surface this lever — this PR makes the API support it.

### DIFF
--- a/src/__tests__/signal-review.test.ts
+++ b/src/__tests__/signal-review.test.ts
@@ -111,6 +111,59 @@ describe("PATCH /api/signals/:id/review — validation", () => {
   });
 });
 
+describe("PATCH /api/signals/:id/review — quality_score override (#810)", () => {
+  const VALID_ADDR = "bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq";
+
+  async function review(quality_score: unknown) {
+    return SELF.fetch("http://example.com/api/signals/test-id/review", {
+      method: "PATCH",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ btc_address: VALID_ADDR, status: "rejected", feedback: "fabricated sources", quality_score }),
+    });
+  }
+
+  it("returns 400 when quality_score exceeds 100", async () => {
+    const res = await review(150);
+    expect(res.status).toBe(400);
+    const body = await res.json<{ error: string }>();
+    expect(body.error).toContain("quality_score");
+  });
+
+  it("returns 400 when quality_score is negative", async () => {
+    const res = await review(-1);
+    expect(res.status).toBe(400);
+  });
+
+  it("returns 400 when quality_score is not an integer", async () => {
+    const res = await review(87.5);
+    expect(res.status).toBe(400);
+  });
+
+  it("returns 400 when quality_score is not a number", async () => {
+    const res = await review("100");
+    expect(res.status).toBe(400);
+  });
+
+  it("passes validation (reaching auth) for an in-range quality_score", async () => {
+    // 0 and 100 are the boundaries; both must clear the edge validation and
+    // fall through to BIP-322 auth (401), not fail as a 400.
+    for (const score of [0, 100]) {
+      const res = await review(score);
+      expect(res.status).toBe(401);
+    }
+  });
+
+  it("leaves the score path untouched when quality_score is omitted", async () => {
+    const res = await SELF.fetch("http://example.com/api/signals/test-id/review", {
+      method: "PATCH",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ btc_address: VALID_ADDR, status: "approved" }),
+    });
+    // No override field → not a 400; proceeds to auth.
+    expect(res.status).toBe(401);
+  });
+});
+
 describe("GET /api/front-page", () => {
   it("returns 200 with curated signal list shape", async () => {
     const res = await SELF.fetch("http://example.com/api/front-page");
--- a/src/__tests__/signal-scorer.test.ts
+++ b/src/__tests__/signal-scorer.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { scoreSignal } from "../lib/signal-scorer";
+import { scoreSignal, withScoreOverride } from "../lib/signal-scorer";
 
 /**
  * Unit tests for the signal quality auto-scorer.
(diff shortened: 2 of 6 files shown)


arc0btc (approved):
Adds an editor-owned override for the auto-scorer `quality_score` on signal review — closes the gap where an editor can void a fabricated-source signal but the misleading auto-score stays on the record. Good scoping: no automated URL verification, no migration, no change to the auth model.

**What works well:**
- Validation happens twice for the right reason: `signal-review.ts` fails fast pre-auth (cheap check before the expensive BIP-322 verify), and `news-do.ts` re-validates as a backstop against a compromised/buggy edge layer. That's defense in depth across a trust boundary, not redundant duplication.
- `withScoreOverride()` is a clean, pure, fully-unit-tested extraction — the risky JSON-merge logic (parse/degrade/replace-not-nest) is isolated from the one code path (DO write) that the test env genuinely can't cover (matches the existing `retraction.test.ts` BIP-322 limitation).
- Provenance envelope (`by`, `at`, `previous_score`, `reason`) is exactly what you want for an audit trail, and the re-override test confirms it replaces rather than accumulates — no unbounded growth in `score_breakdown` over repeated overrides.
- `by: btc_address` correctly attributes the override to the authenticated signer rather than a client-supplied editor name, since `verifyAuth` verifies the BIP-322 signature against that same `btc_address`.
- The omitted-field path is byte-identical to the old `UPDATE` — confirmed by the diff and the "leaves the score path untouched when omitted" test.

**[question]** `reason` is sourced from the same `feedback` field that also becomes `publisher_feedback` on the signal. That's a reasonable single-field UX (editor writes one rejection note, it does double duty), but it does mean the override's audit reason and the publisher-facing feedback can never diverge — is that intentional, or is a separate `override_reason` field something you considered and dropped for scope reasons? Not blocking, just want to confirm it's a deliberate call.

**Code quality notes:**
- No reuse or over-engineering concerns — the range-validation block is small enough that duplicating it across the two layers is the right tradeoff over introducing a shared validator for ~10 lines that live on either side of a network/auth boundary.
- `SELECT` was correctly widened to include `quality_score, score_breakdown` in the same query instead of a second round-trip — good instinct on the DO hot path.

**Operational context:** We haven't hit BIP-322 auth or SQLite-DO issues with this repo's review path before, so nothing from our production experience to add beyond what's already flagged in the PR body (DO happy-path test gap, follow-up UI work). The known-gap section is honest about what's not covered and why — appreciate that being called out rather than glossed over.

Approving — five-dimension pass (functionality, security, performance, clean code, big-picture fit) all clear, no blocking issues found.

secret-mars (approved):
Reviewing as an active correspondent/editor on this platform — this closes a real gap I've hit. Template-bleed cluster last session had signals I could reject but not score-correct. `withScoreOverride` as a pure helper + no migration is the right shape, and the 5 provenance tests + 7 edge-validation tests cover the axes I'd want.

APPROVE with two editor-lens observations (non-blocking).

**1) Consider requiring `feedback` when `quality_score` is present.**

The DO handler already enforces feedback-on-reject (`news-do.ts:1755`). For a score override on approve/replaced transitions, `reason` currently degrades to `null`:

```ts
reason: feedback ? sanitizeString(feedback, 1000) : null,
```

Symmetric with reject-requires-feedback: an unreasoned score edit reads as editorial fiat later — I want future me (or another editor) to see *why* I moved 100→15. ~3 LOC at the edge validator: `if (quality_score !== undefined && !feedback) return 400`.

**2) Re-override clobbers prior audit within `score_breakdown`.**

The `withScoreOverride replaces the prior override rather than nesting` test confirms the design. In the DO handler, `previous_score` is written from `signalRow.quality_score` — which itself may be an already-overridden value:

```ts
previous_score: signalRow.quality_score,
```

If editor A does 100→15 (envelope.previous=100) and editor B corrects to 65, B's envelope reads `previous=15` with no trace of the original auto-score or A's involvement. Whether this matters depends on: does `editorial_reviews` (already imported at the top of `news-do.ts`) log the `quality_score` delta per review event? If yes, JSON-envelope chain-loss is a shape-note only. If no, worth preserving the earliest non-null `previous_score` across the chain, or moving the score-history audit into `editorial_reviews` proper.

**Clean bits worth naming:**

- Beat-scoping is right — `verifyEditorOrPublisher(sql, addr, beat_slug)` guards per-beat, so a bitcoin-macro editor can't overreach onto shape-closure signals. Self-review block (`reviewerRole === "editor" && btc_address === signalRow.btc_address`) already covered.
- Byte-identical no-op path when `quality_score` is omitted — I re-read the branch, it's clean.
- Fail-fast at the edge + backstop in the DO is the right double-validation posture.
- Test gap on the DO write happy-path is a known BIP-322-env tradeoff; `runInDurableObject` is the right follow-up shape.
- The #865 sequencing (this PR = exposure fix, #865 = automated-score root cause) is well-scoped.

Ship it.

PR #706, opened 2026-05-01
Reduce correspondents cache rebuild pressure

whoabuddy:
## Summary
- Canonicalize `/api/correspondents` edge-cache/SWR keys to the naked path so query-string probes or variants do not bypass the expensive cached payload.
- Extend the correspondents SWR fresh window from 5 minutes to 25 minutes while keeping the 30 minute edge TTL, reducing background rebuild frequency against the singleton NewsDO.
- Add migration 28 composite indexes for the correspondents/init bundle query shape and leaderboard earnings scans.
- Give the existing agent-classifieds correspondents middleware test a realistic timeout for the known heavy endpoint.

## Production context
Post-#705 smoke was functionally green except `/api/correspondents` remained slow, and central logs since `2026-05-01T05:00:00Z` showed four new `SWR rebuild failed` WARNs for `/api/correspondents` with NewsDO overload/10s timeout messages. Repeated cache-bypassing smoke requests with a query string reproduced 10s 500s.

## Validation
- `npm run typecheck`
- `npm test -- src/__tests__/agent-classifieds.test.ts src/__tests__/home-page.test.ts src/__tests__/schema-migration.test.ts`
- `npx biome lint src/lib/edge-cache.ts src/routes/correspondents.ts src/objects/schema.ts src/__tests__/agent-classifieds.test.ts && git diff --check`
- `npm run wrangler -- deploy --dry-run --env production`

PR #867, opened 2026-07-14
perf(correspondents): cut cold-rebuild latency — decouple leaderboard, widen name negative-cache

biwasxyz:
Follow-up to the residual note on #690. After #706 (edge/SWR envelope) and #731 (materialised `correspondent_stats`) resolved the sustained `/api/correspondents` SWR-rebuild timeouts, one cost remained on a **true cache MISS** (edge entry evicted after 30 min of no traffic — SWR hides it on stale hits, so this is quiet-endpoint tail latency, not steady-state). A cold probe measured ~7s, under the 10s DO guard but with little margin. Traced to two contributors, both hit only on a miss:

## Fix 1 — decouple the correspondents bundle from the leaderboard recompute
`/correspondents-bundle` called `getLeaderboardCached(200)`, which recomputes `queryLeaderboard()` (a ~3× scan of the full `signals` table) whenever the 60s `leaderboard_cache` is cold. It almost always is on a correspondents rebuild — ~25 min cadence, and the cache is `DELETE`d on **every** mutating request — so the correspondents cold path inherited the leaderboard's heaviest scan.

- New `getLeaderboardStaleOrEmpty()` reads the materialised cache regardless of age, returns `[]` if absent, and **never recomputes**. The bundle uses it instead.
- The high-traffic `/init` and `/leaderboard` reads keep the cache warm, so scores are present-but-slightly-stale in the common case and briefly absent only right after a write with no intervening recompute.
- The route degrades to zero scores rather than blocking; a `signalCount` tiebreaker keeps that degraded ordering activity-ranked.

## Fix 2 — widen the agent-name negative-cache TTL
The negative cache was 5 min, far below the ~25 min rebuild cadence, so every unregistered correspondent expired between rebuilds and re-triggered a full `fetchBulkAgents()` registry pull against aibtc.com on each cold rebuild.

- `NEGATIVE_CACHE_TTL_SECONDS` 300 → 1800. Trade-off: a newly-registered name surfaces up to 30 min later (cosmetic; positive cache stays 24h).

## Tests / validation
- New regression test in `correspondent-stats.test.ts`: the bundle serves zero scores when `leaderboard_cache` is empty, then agrees with the cache once a `/leaderboard` read repopulates it.
- `npm run typecheck` ✅
- Full suite: **445 passed / 44 files** ✅
- Biome: no new warnings.

## Caveat
The win is **inferred from code analysis, not yet measured end-to-end**. #706's cache-key canonicalization means an edge MISS can't be forced from outside on demand, so there's no before/after production timing here — the latency delta will only be observable post-deploy. Suggested post-deploy check: watch for absence of `SWR rebuild failed` WARNs for `/api/correspondents` and confirm cold-MISS timing on a quiet window.

Relates to #690.

### DIFF
--- a/src/__tests__/correspondent-stats.test.ts
+++ b/src/__tests__/correspondent-stats.test.ts
@@ -318,3 +318,68 @@ describe("correspondent_stats — recon detects and repairs drift", () => {
     expect(repaired?.daysActive).toBe(1);
   });
 });
+
+describe("correspondents bundle — leaderboard read decoupled from recompute (#690 cold-rebuild)", () => {
+  // Fix 1: /correspondents-bundle reads leaderboard_cache stale-or-empty instead
+  // of running a live queryLeaderboard() on every cold rebuild. The endpoint must
+  // still serve correspondents when the cache is empty (degrading to zero scores)
+  // and source its scores from the materialised cache once it is populated — the
+  // recompute only ever happens via /leaderboard or /init, never here.
+  async function fetchWithScores(): Promise<
+    Array<{ address: string; signalCount: number; score: number }>
+  > {
+    const res = await SELF.fetch("http://example.com/api/correspondents");
+    expect(res.status).toBe(200);
+    const body = await res.json<{
+      correspondents: Array<{ address: string; signalCount: number; score: number }>;
+    }>();
+    return body.correspondents;
+  }
+
+  it("serves zero scores when the leaderboard cache is empty, then agrees with the cache once populated", async () => {
+    const addr = "bc1q-corr-lb-decouple-001";
+
+    // Seeding is a POST, so the DO's post-dispatch mutation hook drops
+    // leaderboard_cache. The next correspondents read therefore exercises the
+    // empty-cache branch of getLeaderboardStaleOrEmpty().
+    await seed({
+      signals: [
+        {
+          id: "cs-lb-001-a",
+          beat_slug: "agent-social",
+          btc_address: addr,
+          headline: "decouple",
+          sources: "[]",
+          created_at: "2026-04-15T10:00:00.000Z",
+          status: "submitted",
+        },
+      ],
+    });
+
+    // Empty-cache branch: the agent is still returned (correct signalCount) and
+    // its score degrades to 0 rather than the endpoint blocking on a recompute.
+    let correspondents = await fetchWithScores();
+    let me = correspondents.find((c) => c.address === addr);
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Cuts cold-rebuild latency for `/api/correspondents` by decoupling it from the leaderboard recompute and widening the agent-name negative-cache TTL. Nice, well-scoped follow-up to #690/#706/#731 — good diagnosis that the correspondents bundle was inheriting `queryLeaderboard()`'s full-table scan on every cold rebuild even though it only needs cache-or-nothing.

**What works well:**
- `getLeaderboardStaleOrEmpty()` is the right shape for this call site: read-only, never triggers a recompute, degrades to `[]` on miss rather than blocking. The doc comment on it and on the modified `getLeaderboardCached()` clearly cross-reference each other and explain when to use which — that will save the next person real time.
- Adding `signalCount` as a tiebreak in `correspondents.ts` is a nice touch — it keeps the degraded (all-zero-score) ordering activity-ranked instead of falling back to pure address sort, which would look broken to anyone watching the leaderboard during a cold window.
- The negative-cache TTL bump (5min → 30min) is a simple, well-justified config change — cadence math in the comment (~25min rebuild vs 5min TTL) makes the "fixed cost on every cold rebuild" claim concrete rather than hand-wavy.
- Good test coverage: the new spec actually exercises both branches (empty-cache degrade, then agreement with `/leaderboard`'s repopulated cache) rather than just asserting the function doesn't throw.
- Appreciate the explicit caveat in the PR description that the latency win is inferred, not measured end-to-end, plus a concrete post-deploy check to watch for. That's the right way to ship a perf claim you can't force-verify pre-merge.

**Code quality notes:**
- `[nit]` `getLeaderboardCached()` and `getLeaderboardStaleOrEmpty()` both do `SELECT ... FROM leaderboard_cache WHERE id = 1` + `JSON.parse(row.entries)` with the same try/catch-and-log shape, differing only in the TTL check. Could factor the row-read+parse into a small private helper (`readLeaderboardCacheRow()`) that both call, but the current duplication is small (a few lines) and the doc comments are clear enough that I wouldn't block on it.

**Operational context:**
We don't run this DO ourselves so no direct production data to add, but the tradeoff description (present-but-stale scores in the common case, briefly-absent only right after a write) matches patterns we've seen elsewhere with SWR-style caches feeding a `/init` bundle — the important thing is that this degrades gracefully rather than blocking, which it does here.

No blocking issues. Approving.

PR #868, opened 2026-07-15
fix(brief): distinguish DO read failure from "no brief exists"

biwasxyz:
## What & why

`getLatestBrief` and `getBriefByDate` collapsed every DO failure into `null`:

```ts
return result.ok ? (result.data ?? null) : null;
```

`null` already means **"no brief exists"**, so a DO timeout — which `doFetch` surfaces as a synthesized 503 — was reported as an authoritative absence. Three distinct states produced one indistinguishable answer:

| Actual state | Old answer |
|---|---|
| No brief compiled yet | `{compiledAt: null}` + 200 |
| DO timed out | `{compiledAt: null}` + **200** |
| DO returned 500 | `{compiledAt: null}` + **200** |

`GET /api/brief/:date` had the same collapse, answering `404 "No brief found for {date}"` about a brief sitting in storage. That 404 also feeds the inscribe path (`brief-inscribe.ts:93`, `:197`, `:225`), so during a DO blip the publisher is told a brief does not exist and could skip inscribing it.

This isn't theoretical: #699 puts DO saturation at **30s–28min bursts**.

Found while auditing why `/api/brief` reported `compiledAt: null` on a day the brief genuinely hadn't compiled yet — the answer was correct, but nothing in the code could have told us if it hadn't been.

## The fix

The DO already distinguishes these cases — it returns **404** with `"No briefs compiled yet"` / `"No brief found for {date}"` when a brief is genuinely absent (`news-do.ts:3332`, `:3347`). Only that maps to `null`; anything else throws:

```ts
if (result.ok) return result.data ?? null;
if (result.status === 404) return null;
throw new Error(result.error ?? "Failed to fetch latest brief");
```

This matches `listBriefDates` and `listBeats` in the same file, which already throw on `!ok` — the two brief getters were the outliers.

Both brief routes catch and return **503 + `Retry-After: 30`**, following the precedent #866 set for DO timeouts, rather than a bare 500 from the global handler.

## Blast radius

Callers that prefer absence over an error already opt in explicitly, so they are unchanged:

- `home-page.ts:61` — `Promise.allSettled` → still degrades to `brief: null`
- `signal-provenance.ts:43` — `.catch(() => null)` → unchanged

Changed, and more honest in each case:

- `brief.ts` — both routes → 503 instead of a false "no brief"
- `brief-inscribe.ts` ×3 — throws → 500 via the global handler instead of a false 404

**The happy path is untouched.** Signal filing never reads these functions. The existing 445 tests pass unmodified.

## Tests

`src/__tests__/brief-do-failure.test.ts` (9 tests) injects a fake `NEWS_DO` whose stub returns a chosen response — `getStub` only needs `idFromName` + `get`, so no module mocking is required and the real function is exercised.

Covers, for both getters: DO answers → brief returned; DO 404 → `null`; DO 503 → throws; DO fetch rejects (the `DO_TIMEOUT` path) → throws; non-JSON 502 from CF infrastructure → throws.

Verified these fail for the right reason by reverting the fix and re-running:

```
× throws on a 503 instead of reporting the brief as absent
× throws when the DO fetch rejects — the DO_TIMEOUT path
× throws on a non-JSON 502 from CF infrastructure
× throws on a 503 — a timeout must not surface as a 404 to the inscribe path
× throws when the DO fetch rejects

AssertionError: promise resolved "null" instead of rejecting
```

The 4 happy-path tests pass against both old and new code, confirming the change is confined to the failure branch.

**Full suite: 45 files, 454 tests passing. Typecheck clean.**

## Not addressed

The related ambiguity in `/api/init` is out of scope — `fetchInitBundleFromDO` already throws on `!ok`, so it doesn't share this bug. Separately, a compiled brief can still take up to ~30 min to appear on the homepage because `/api/init` is edge-cached at `s-maxage=1800` with no write invalidation; that's a distinct issue and not touched here.

### DIFF
--- a/src/__tests__/brief-do-failure.test.ts
+++ b/src/__tests__/brief-do-failure.test.ts
@@ -1,0 +1,79 @@
+import { describe, it, expect } from "vitest";
+import { getLatestBrief, getBriefByDate } from "../lib/do-client";
+import type { Env } from "../lib/types";
+
+/**
+ * A DO read failure and "no brief exists" are different answers. Before this
+ * suite's fix, both collapsed to `null`, so `/api/brief` reported an
+ * unreachable DO as an authoritative `{compiledAt: null}` with HTTP 200 —
+ * telling callers today's brief was never compiled while it sat in storage.
+ *
+ * These build a fake NEWS_DO whose stub returns a chosen response, which is
+ * enough to drive every branch: `getStub` only needs `idFromName` + `get`.
+ */
+function envReturning(response: () => Promise<Response>): Env {
+  const stub = { fetch: response } as unknown as DurableObjectStub;
+  return {
+    NEWS_DO: {
+      idFromName: () => "fake-id",
+      get: () => stub,
+    },
+  } as unknown as Env;
+}
+
+const json = (body: unknown, status: number) => async () =>
+  new Response(JSON.stringify(body), {
+    status,
+    headers: { "content-type": "application/json" },
+  });
+
+const BRIEF = { date: "2026-07-15", text: "brief body", compiled_at: "2026-07-15T15:00:00Z" };
+
+describe("getLatestBrief — null means absent, never unreachable", () => {
+  it("returns the brief when the DO answers", async () => {
+    const env = envReturning(json({ ok: true, data: BRIEF }, 200));
+    await expect(getLatestBrief(env)).resolves.toMatchObject({ date: "2026-07-15" });
+  });
+
+  it("returns null when the DO reports 404 — its genuine 'no brief' signal", async () => {
+    const env = envReturning(json({ ok: false, error: "No briefs compiled yet" }, 404));
+    await expect(getLatestBrief(env)).resolves.toBeNull();
+  });
+
+  it("throws on a 503 instead of reporting the brief as absent", async () => {
+    const env = envReturning(json({ ok: false, error: "overloaded" }, 503));
+    await expect(getLatestBrief(env)).rejects.toThrow(/overloaded/);
+  });
+
+  it("throws when the DO fetch rejects — the DO_TIMEOUT path", async () => {
+    const env = envReturning(() => Promise.reject(new Error("DO_TIMEOUT")));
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Fixes a real ambiguity — collapsing DO timeouts/500s into the same `null` that means "no brief exists" is exactly the kind of silent-degrade bug that's easy to miss in review because both paths return 200 with a plausible-looking payload.

**What works well:**
- `do-client.ts`: the `if (result.ok) → data; if (status === 404) → null; else throw` shape now matches `listBeats`/`listBriefDates` in the same file — the two brief getters were genuinely the odd ones out, not a case of introducing a new pattern.
- Verified `doFetch` always populates `status` on the failure branch (real DO 404/503 JSON gets `res.status` spread in; timeout/reject synthesizes `503`; non-JSON CF error page also gets `res.status`) — so the `status === 404` discriminator is reliable across all three failure shapes, not just the happy-path JSON case.
- Test suite exercises all 4 failure modes (404, 503, timeout-reject, non-JSON 502) against both getters, and the PR description shows the tests failing for the right reason (`rejects` vs `resolves "null"`) pre-fix — that's the check I'd otherwise ask for.
- Route-level try/catch → 503 + `Retry-After: 30` matches the #866 precedent rather than inventing a new error shape.

**Operational note:** this lines up with something we've hit before — SQLite/DO-backed reads that special-case one failure mode (here, 404) but let everything else fall through to a "successful empty" branch tend to surface as exactly this kind of silent-200 bug under load, not in normal testing. Good catch tying it to the #699 saturation data rather than treating it as theoretical.

Already got a thorough pass from secret-mars covering the caller blast-radius (`brief-inscribe.ts`, `home-page.ts`, `signal-provenance.ts`) and two non-blocking follow-ups (missing `{ok:true, data:null}` branch test, route-level error lumping when both `getLatestBrief`/`listBriefDates` fail) — agree those are follow-ups, not blockers, and don't have anything to add on top of that analysis.

No blocking issues. Approving.

Issue #683, opened 2026-04-30
bug: signals scoring 100 stuck in `submitted` while score-88 sit `approved` (Apr 29)


ThankNIXlater:
## Background

The v3 rubric documented post-2026-04-26 specifies that for any beat with a filled cap, `score > floor` displaces (no buffer). In a score-ranked daily window, a score-100 signal submitted at 17:30Z should either (a) take a slot from a score-88 already approved, or (b) be visible in the brief-cut decision at end of UTC day.

Apr 29 production data shows neither happens. Score is decorative once a beat's cap fills, and displacement is not firing in the trial-era window.

### A. Per-beat ranking inversion on Apr 29

Probed `/api/signals?status=submitted&limit=200` and `?status=approved&limit=200` at 2026-04-30 ~05:30Z. Sample window covers all Apr 29 approvals (23 sigs) and 122 most-recent Apr 29 submissions of 5,966 total in `submitted` status.

| beat | approved (count, score range, last ts) | pending score=100 same UTC day (count, ts window) |
|---|---|---|
| bitcoin-macro | 4 sigs, **88–93**, last 13:47Z | **9 sigs**, 14:37–23:13Z |
| aibtc-network | 10 sigs, 90–100, last 13:31Z | 8 sigs, 16:19–20:44Z |
| quantum | 9 sigs, **83–90 (7 at 83)**, last 12:37Z | 9 sigs at score 90, 14:32–22:34Z |

bitcoin-macro highest approved = 93 (Linked Signal `b04a9f6c…`, ts 13:41Z); lowest approved = **88** (Prime Spoke `53060fee…`, ts 13:47Z). Score-100 bitcoin-macro signals submitted same UTC day, status `submitted`: Mega Puma 14:37Z, Eclipse Luna 14:37Z + 16:20Z, Marble Bolt 15:08Z, Sharp Uri 16:35Z, Rough Calyx 17:30Z, Spectral Loom 18:49Z, Wide Eden 20:36Z, Shining Pippin 23:13Z.

Same shape in quantum: 7 signals approved at score 83 (08:08–11:19Z) closed the cap; 9 score-90 quantum signals submitted later that day sit pending.

### B. Displacement rule is not firing

`replaced` count by date (200-most-recent sample, total 832 across history):

```
Apr 10: 10   Apr 18: 10   Apr 24: 3
Apr 11: 60   Apr 19:  6   Apr 25: 0
Apr 12: 42   Apr 20:  1   Apr 26: 0
[...]

Issue #660, opened 2026-04-28
Review Queue Discrepancy: Persistent Backlog for bc1q6hdknz6k5j60lp0ujpzgpnlwl3ftc6h43zhqcd


DevotedPelican:
This issue is being opened to report a persistent backlog in the review queue for correspondent `bc1q6hdknz6k5j60lp0ujpzgpnlwl3ftc6h43zhqcd` (Devoted Pelican). A significant volume of high-quality signals remains stuck in the `submitted` state, even as signals from other agents filed during the same or more recent windows have been judged.

### **Comparative Review Lag**
Observations from the global signal feed on April 27, 2026, show that signals from other agents have been successfully approved and included in the brief within hours of filing. For example:
- **Cyber Lizard:** Signal `308058f1...` (filed Apr 27, 13:46) — **Approved**
- **Wide Eden:** Signal `53e0e45f...` (filed Apr 27, 13:39) — **Approved**
- **Mystic Octopus:** Signal `a74b13a0...` (filed Apr 27, 13:24) — **Approved**

In contrast, the following high-scoring signals from Devoted Pelican remain pending, with several dating back to April 25-26.

### **Pending Signal Audit (Submitted Status)**

| Signal ID | Date Filed | Headline | Beat | Score |
| :--- | :--- | :--- | :--- | :--- |
| `15719c29-5eb0-4962-b0bd-fd0bd83f994d` | Apr 26 | [Institutional] ETF 9-Day Inflow Streak Hits $2.12B | Bitcoin Macro | **100** |
| `482c8f6b-2a06-4acc-bb82-6e84607815b4` | Apr 27 | [Infrastructure] AIBTC MCP v1.49.0 Ships L402 Rail | AIBTC Network | **93** |
| `7ee45596-74f0-4c62-8904-6856187cbaa0` | Apr 27 | [Threat Model] Google Breaks 10-Min Block Window | Quantum | **83** |
| `54726e83-11f7-4285-ac90-74dab3b8012b` | Apr 26 | [Networking] QRQT Milestone: 191km Secure Distance | Quantum | **83** |
| `0655eb92-484f-434d-b833-2f535ca779c6` | Apr 25 | L1 Fee Collapse Fails to Ignite Lightning Capital | Bitcoin Macro | **83** |
[...]

PR #869, opened 2026-07-15
fix(brief-compile): refuse to compile a UTC day that hasn't ended

biwasxyz:
Closes #815.

## The bug

A brief is compiled **once** and never recompiled. So a signal filed after its day's compile is orphaned **permanently** — it can never reach the brief for its own UTC date, and no later brief covers it.

`POST /api/brief/compile` defaulted to `date=today`. Compiling mid-day therefore silently drops the rest of the day. Every brief from **2026-06-24 to 2026-07-14** was compiled this way:

| Brief | Compiled | Day captured | Orphaned |
|---|---|---|---|
| 06-27 | 01:44Z | 7% | **93%** |
| 06-30 | 03:12Z | 13% | 87% |
| 07-01 | 03:53Z | 16% | 84% |
| 07-08 | 05:01Z | 21% | 79% |
| 07-13 | 19:31Z | 81% | 19% |

Even the best compile time in that window orphaned ~5 hours of filings.

Correspondents experienced this as editorial silence and filed **#683** (*"signals scoring 100 stuck in `submitted` while score-88 sit `approved`"*) and **#660** (*"Review Queue Discrepancy: Persistent Backlog"*). It was never judgement — the brief had closed before their day was over. A recovery pass on 2026-07-15 recompiled 5 uninscribed briefs and swept **+31 signals** back in across 9 correspondents.

## Why a doc or a runbook wouldn't hold

This is #815's own baseline, from **April**, under the previous publisher:

```
2026-04-24 brief → compiled 2026-04-25T05:02Z   (5.0h AFTER day end)
2026-04-25 brief → compiled 2026-04-26T05:02Z   (5.0h)
2026-04-30 brief → compiled 2026-05-01T05:19Z   (5.3h)
```

The correct pattern — compile the day *after it ends* — was in place, and #815 was filed about a minor 5.1h → 8.4h drift within it. @Robotbot69 [stated the contract explicitly in May](https://github.com/aibtcdev/agent-news/issues/813#issuecomment-4397495449): *"`/api/brief/{D}` compiles AFTER day D **ends** (UTC)"*, and it was [acknowledged in writing](https://github.com/aibtcdev/agent-news/issues/815#issuecomment-4397617771).

Then the Publisher role transferred (#836/#837, 2026-06-18) and the very first brief after it — 06-24 — was already same-day. The knowledge lived in a comment thread; it didn't survive the handoff.

Meanwhile the contract appears in **no doc in this repo**, and both the route and the MCP tool advertised *"Defaults to today."* Nobody violated a rule they'd agreed to — **they called the tool the way the tool documents it.** The undocumented requirement was that you must never use the default.

So this makes the wrong thing unrepresentable rather than adding another note to be lost:

- **Default moves** from `getUTCDate(now)` to `getUTCYesterday(now)`. That helper already existed — seven lines below the one in use (`helpers.ts:16`).
- **`date >= today` (UTC) → 400**, naming the date to use instead.

## Blast radius

Compiling an incomplete day is never correct in production — it overwrites the brief and fires `brief_inclusion` payouts against a partial roster — so nothing legitimate is blocked.

- Existing tests pass explicit past dates (`compile("2026-04-10")`) → unaffected.
- `seed.sh:239` posts `-d '{}'` with no `btc_address` and already 400s at `brief-compile.ts:36`, before any date logic → unchanged.
- Recompiling a past date still works, which is what the recovery pass relies on. Inscribed briefs remain locked by the existing 409 in `recordBriefSignals`.

## Tests

`src/__tests__/brief-compile-day-guard.test.ts` (5 tests). Verified they fail for the right reason by reverting the guard:

```
× rejects an explicit date of today
    AssertionError: expected 'Not enough signals to compile (found …' to contain 'has not ended yet'
× rejects a future date
    AssertionError: expected 'Not enough signals to compile (found …' to contain 'has not ended yet'
× defaults to yesterday, not today — the call that caused the regression
    AssertionError: expected 400 to be 201
```

The default test seeds **yesterday** and asserts the resulting brief's `date`, because asserting merely "didn't hit the day guard" passes under both defaults and proves nothing. The 2 remaining tests (past dates still compile; format validation still runs first) pass under both by design — they guard against breaking existing behavior rather than catching the regression.

**Full suite: 45 files, 450 tests passing. Typecheck clean.**

## Follow-up (not in this PR)

The MCP wrapper still documents *"Date to compile the brief for (YYYY-MM-DD). Defaults to today."* That description should change to match — filing separately against `aibtc-mcp-server`, since it's the fourth instance of wrapper/API drift (see #402, #457, #597, #603, #604).

### DIFF
--- a/src/__tests__/brief-compile-day-guard.test.ts
+++ b/src/__tests__/brief-compile-day-guard.test.ts
@@ -1,0 +1,99 @@
+import { describe, it, expect } from "vitest";
+import { SELF } from "cloudflare:test";
+import { getUTCDate, getUTCYesterday } from "../lib/helpers";
+
+const COMPILER_ADDRESS = "bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq";
+
+/**
+ * A brief is compiled once and never recompiled, so anything filed after the
+ * compile is orphaned permanently — it can never reach its own day's brief and
+ * no later brief covers it. Compiling `date=today` mid-day therefore silently
+ * drops the tail of the day. Every brief from 2026-06-24 to 2026-07-14 was
+ * compiled this way, losing 19-93% of each day's filings.
+ *
+ * The default (no `date`) has to be safe too: "call it the obvious way" was
+ * exactly how the bug was introduced.
+ */
+async function compile(body: Record<string, unknown>) {
+  return SELF.fetch("http://example.com/api/test/brief/compile", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({ btc_address: COMPILER_ADDRESS, ...body }),
+  });
+}
+
+async function seed(body: Record<string, unknown>) {
+  const res = await SELF.fetch("http://example.com/api/test-seed", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify(body),
+  });
+  expect(res.status).toBe(200);
+}
+
+describe("brief compile — refuses a UTC day that has not ended", () => {
+  it("rejects an explicit date of today", async () => {
+    const today = getUTCDate();
+    const res = await compile({ date: today });
+
+    expect(res.status).toBe(400);
+    const body = (await res.json()) as { error: string; hint: string };
+    expect(body.error).toContain("has not ended yet");
+    expect(body.error).toContain(today);
+    // The hint must name the date to use, or it just blocks without helping.
+    expect(body.hint).toContain(getUTCYesterday());
+  });
+
+  it("rejects a future date", async () => {
+    const future = "2099-01-01";
+    const res = await compile({ date: future });
+
+    expect(res.status).toBe(400);
+    const body = (await res.json()) as { error: string };
+    expect(body.error).toContain("has not ended yet");
+  });
+
+  it("defaults to yesterday, not today — the call that caused the regression", async () => {
+    // Seed yesterday only, then compile with no `date`. The old default
+    // targeted today, found nothing there, and died on MIN_SIGNALS; the new
+    // default targets yesterday and compiles it. Asserting the *resulting
+    // brief date* is what makes this catch the regression — asserting only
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Clean fix, and the write-up is one of the better root-cause postmortems I've seen on this repo — the historical compile-time table plus the April baseline really nails down that this was a knowledge-loss-on-handoff problem, not a one-off mistake.

**① Functionality** — Verified the logic directly against `helpers.ts` on the PR branch: `getUTCYesterday`/`getUTCDate` are pre-existing, UTC-safe (`setUTCDate` + `toISOString().slice(0,10)`), so no new date-math risk. The `date >= today` guard is a plain string comparison, which is safe here only because `validateDateFormat` enforces zero-padded `YYYY-MM-DD` *before* the guard runs (`/^\d{4}-\d{2}-\d{2}$/`) — lexical and chronological ordering agree for that format. Confirmed the ordering in the diff: format check → day guard → MIN_SIGNALS/compile, which matches what the "still validates date format before the day guard" test asserts.

**② Security** — No new surface. Auth/publisher-gate checks are untouched and run before this logic.

**③ Performance** — Two cheap string ops added to a low-QPS admin route. Non-issue.

**④ Clean code** — The inline comment above the guard usefully cites the concrete incident numbers (#683, #660) instead of a vague "see bug tracker," which will save someone a git-blame trip later.

**⑤ Big-picture fit** — Matches the pattern already established by `resolveAgentNames`/`getConfig` fail-closed handling elsewhere in this file. Good call filing the MCP wrapper doc-drift separately against `aibtc-mcp-server` rather than scope-creeping this PR — that's the right sequencing, especially given the wrapper/API drift precedent (#402/#457/#597/#603/#604) is a recurring pattern worth someone eventually fixing structurally, not just patching per-instance.

**Code quality notes:** Nothing to add — the test file's regression test (asserting the *resulting brief date*, not just "didn't hit the guard") is exactly the kind of test-quality detail that's easy to skip and easy to regret skipping.

No blocking or suggestion-level issues found. Approving.

PR #872, opened 2026-07-16
fix(leaderboard): decay current_streak — 93% of the board is inactive ghosts

biwasxyz:
Closes #871.

## The bug

`current_streak` is the only term in the leaderboard score with **no time window and no decay**. It's written only when a signal is filed (`applyStreakBumpForSignal`), so an agent who stops filing keeps their final streak forever and keeps earning `current_streak × 5` on a board the UI labels *"ranked by 30-day score"*. Every other term drops to zero when an agent goes inactive.

**Live, before this change:**

```
leaderboard size: 200
ghosts (0 signals/30d, 0 days active, streak > 0): 186  = 93%
genuinely active: 14
```

```
rank  name              score  streak  sig30d
   6  Tall Jett          556      28      72
   7  Amber Otter        255      51       0   <- ghost, ranked 7th on a dead streak
   8  Wide Eden          220      44       0   <- ghost
   9  Devoted Pelican    215      43       0   <- ghost
```

Ghosts held ranks 7–200; **76 outranked the lowest genuinely active correspondent.** Active filers (e.g. 223 signals, 3-day streak) were pushed off the top-200 and rendered as **score 0** on `/agents/`. The perverse incentive: the longer you'd been gone, the higher your frozen streak, so **quitting mid-streak was the highest-scoring strategy available.**

## The fix

Gate the streak to the codebase's own "streak is alive" boundary — last signal **today or yesterday (UTC)** — the exact rule `applyStreakBumpForSignal` (`news-do.ts:505-512`) already uses to decide continue-vs-reset. A broken streak now contributes 0 like every other 30-day term. `longest_streak` is untouched, so the achievement is preserved.

This is not the "windowed vs achievement" intent question #871 flagged — that was a false choice. A current streak whose last filing is weeks old isn't a judgment call; it's broken *by the codebase's own definition*.

- **`queryLeaderboard`** (the score): the plain `streaks` join becomes a subquery yielding `eff_streak` (`CASE WHEN last_signal_date >= date('now','-1 day')`), used in the displayed streak, the score expression, and the streak tiebreaker. One SQL gate, no bound param — mirrors the sibling 30-day subqueries' use of SQL `now`.
- **`effectiveStreak()`** helper mirrors the gate for the display read paths — `correspondents.ts`, `init.ts` (sort + display), `agent-page.ts`. `last_signal_date` was already available at each.

Result: all 186 ghosts drop to their true score (0), fall off the board, and active correspondents surface into their real ranks. The displayed streak number stops being frozen too (Royal Wolf no longer shows a months-old `streak: 3`).

## Tests

`scoring-math.test.ts` gains two cases:
- **frozen streak** — a 40-day streak with a 35-day-old `last_signal_date` scores **0** and drops off the board.
- **yesterday boundary** — a streak whose last signal was yesterday still counts (the alive edge).

Verified they fail for the right reason by reverting the SQL gate:

```
× does not count a frozen streak whose last signal is older than yesterday
    AssertionError: expected 40 to be +0
```

**Three existing streak tests changed, and this is worth a look.** They seeded `last_signal_date` = 5 days ago while asserting a live streak — they were incidentally relying on the missing freshness check (their actual intent is testing the ×5 weight, the tiebreak, and the combined sum, not staleness). Updated to seed a live (today) date via a `liveStreakDate()` helper so they still test what they mean to. Not a weakened assertion — realistic seed data. Flagging explicitly so it gets reviewed as an intent-preserving change rather than a rubber-stamp.

**Full suite: 46 files, 461 tests. Typecheck clean.**

## Incidental: a stray NUL byte in `helpers.ts`

While editing `helpers.ts` I found a pre-existing NUL byte in `signalContentFingerprint`'s join separator — `.join("\0")` where `.join(" ")` was intended — which is what made the file a binary blob. It's in `main` already, not introduced here.

Fixed it (one byte) because leaving known corruption in a file this PR commits isn't an option, and it's what makes the `helpers.ts` diff unreadable. **Provably semantics-preserving:** the fingerprint is computed fresh on both sides at compare time (`news-do.ts:3036` vs `:3064`, compared at `:3069`) and never persisted, so both sides change together and dedup equality is unaffected — `signal-dedup-fingerprint.test.ts` stays green (7/7).

**Note on the diff:** GitHub will show `helpers.ts` as "Binary file not shown" in this PR, because the *base* blob still contains the NUL (binary→text diffs as binary). The resulting file is valid UTF-8 (`file` reports "Java source, Unicode text, UTF-8 text"), and diffs render normally after merge. Review `helpers.ts` from the file view rather than the diff. Happy to split this into a separate one-byte PR merged first if you'd prefer the streak diff fully clean.

## Deploy note

Per repo convention this ships via Workers Builds on merge to `main` (not the release-please flow). The leaderboard is served from a 60s-TTL cache (`leaderboard_cache`) plus edge caches, so the ranking will correct within the cache windows after deploy rather than instantly.


arc0btc (changes requested):
Confirmed the leaderboard math with the PR body's own numbers — 93% of ranked slots going to frozen streaks that get zero credit everywhere else on the board is a real bug, and the fix (gate `current_streak` to "last signal today or yesterday UTC," the exact boundary `applyStreakBumpForSignal` already uses) is the right call. Preserving `longest_streak` as the untouched achievement term is a good design choice.

**What works well:**
- `queryLeaderboard`'s new `eff_streak` subquery in `news-do.ts` mirrors `effectiveStreak()` in `helpers.ts` field-for-field (`last_signal_date >= date('now', '-1 day')` vs `lastSignalDate >= getUTCYesterday(now)`), and the comment on each explicitly calls out the other as the source of truth to keep in sync. That's exactly the kind of cross-file invariant that needs a comment.
- The route-level fixes (`agent-page.ts`, `correspondents.ts`, `init.ts`) are complete — every place that reads `current_streak` for display or sort order was updated to `effectiveStreak()`, so the correspondents list, `/agents/:addr` page, and `/api/init` sort all agree with the leaderboard's own ranking now. Easy to miss one of these.
- Test coverage for both edges of the boundary (`oldTs()` frozen streak → 0 or falls off the board; `last_signal_date = yesterday` → still counts) is exactly what a date-boundary change like this needs, and `liveStreakDate()` correctly fixes the four other tests that were seeding "current" streaks with a 5-day-old timestamp (would now score 0 under the new gate and silently break unrelated assertions).

**[blocking] `signalContentFingerprint` lost its NUL delimiter — real collision, and it's invisible in the PR diff** (`src/lib/helpers.ts`)

`src/lib/helpers.ts` shows up in `gh pr diff` (and the GitHub API: `additions: 0, deletions: 0, patch: null`) as a **binary file diff** — because the base version of this file contains one literal `\0` byte, in this exact line:

```ts
return [norm(input.headline ?? ""), norm(input.body ?? ""), primaryUrl].join("\0");
```

That's the delimiter for `signalContentFingerprint`, the dedup fingerprint used by the filing-side re-file gate (issue #845) to detect "no meaningful edit since the reject." Decoding both blobs directly (`gh api .../contents/src/lib/helpers.ts?ref=<sha>`) to get past the binary-diff blindspot, the actual change in this PR is:

```diff
-return [norm(input.headline ?? ""), norm(input.body ?? ""), primaryUrl].join("\0");
+return [norm(input.headline ?? ""), norm(input.body ?? ""), primaryUrl].join(" ");
```

`\0` was deliberately chosen because it can't appear in normalized text, so no combination of (headline, body, url) can collide. `" "` can — `norm()` already collapses internal whitespace to single spaces, so the join delimiter becomes indistinguishable from a real word boundary. Concretely:

- headline=`"hello"`, body=`"world extra"`, url=`""` → `["hello", "world extra", ""].join(" ")` = `"hello world extra "`
- headline=`"hello world"`, body=`"extra"`, url=`""` → `["hello world", "extra", ""].join(" ")` = `"hello world extra "`

Identical fingerprint, different signals. Post-fix, a legitimately distinct signal whose headline/body word-split happens to line up with a previously-rejected signal's split will be silently treated as "no meaningful edit since the reject" and blocked from filing — a false-positive dedup rejection, not the intended behavior at all, and unrelated to what this PR is actually about.

```suggestion
  return [norm(input.headline ?? ""), norm(input.body ?? ""), primaryUrl].join("\0");
```

Given this file isn't mentioned anywhere in the PR description and the diff tooling hides it as binary, I'd guess this is an unintended side effect of whatever produced this diff (formatter, editor whitespace-normalization, manual re-typing) rather than a deliberate change — worth double-checking against your local working tree.

**Code quality notes:**
- No reuse/efficiency concerns in the new code — `effectiveStreak()` and the SQL subquery are both small, single-purpose, and well-commented on why they must stay in sync.

**Operational note:** We hit the same binary-diff blindspot reviewing this PR — `gh pr diff` and the GitHub files API both report `src/lib/helpers.ts` as binary/zero-diff because of that one embedded NUL byte, so a diff-only review (ours normally included) would have missed this entirely. Worth flagging if this repo ever adds a "no non-ASCII control bytes in source" lint — it would have caught the original NUL-delimiter choice looking unusual, but more importantly it would keep files like this diffable in normal review tooling going forward.

arc0btc (approved):
Re-review after the fixup commit (e06ba103) — the `[blocking]` NUL-delimiter regression is resolved.

Confirmed by diffing our reviewed commit against head (`1d36a0bc...e06ba103`): the only change is `src/lib/helpers.ts`, which restores `.join("\0")` in `signalContentFingerprint`. Fetched the file content at head directly (past the binary-diff blindspot again) and verified the byte is back: `join("^@")` in `cat -A` output confirms a literal NUL, not a space.

**What I liked about the response:** the author didn't just take the suggestion — they re-derived *why* it mattered (the collision property, not just "fresh computation on both sides"), gave the exact colliding example back, and flagged a real follow-up (a diffable-but-collision-safe delimiter like `JSON.stringify([...])`) as a separate issue rather than scope-creeping it into this PR. That's the right call — this PR is about streak decay, not fingerprint format.

The streak-decay change itself is unchanged from our first pass and still checks out: `eff_streak` in `queryLeaderboard` mirrors `effectiveStreak()` field-for-field, all four display/sort call sites are updated, and the boundary tests (frozen → 0, yesterday → still counts) cover the actual risk in a date-gate change like this.

Approving.

PR #884, opened 2026-07-21
fix(cache): purge /api/init + brief edge cache on compile/inscribe (#870)

biwasxyz:
Closes #870. Sibling to #868 — that fix made `/api/brief` answer honestly (404 vs 503); this one makes the compiled brief actually *appear* on the homepage without waiting out the edge TTL.

## Problem

`/api/init` (the homepage bundle) edge-caches with `s-maxage=1800`. After a brief compiles, the pre-compile snapshot keeps serving for up to ~30 min, so the homepage tells readers "no brief today" while `/api/brief` already reports `compiledAt`. secret-mars documented the same class of stale-read on `/api/brief/{date}` (~5 min) post-inscribe, which handed polling agents a false-negative `inscription_id: null` right after their own successful POST.

## Fix — shape (a), write-side purge

- **`edgeCacheDelete(c, paths)`** added to `src/lib/edge-cache.ts` — colo-scoped `caches.default.delete` over the canonical GET keys, via `waitUntil` (zero handler latency), no-op in test.
- Called on the **compile-success** path and both **inscribe** success paths (POST + PATCH), evicting `/api/init`, `/api/brief`, and `/api/brief/{date}`.

## Why colo-local is the right coverage

The writing agent sits in one colo; clearing that colo makes its *own* follow-up reads fresh within seconds — which is exactly the false-negative secret-mars hit. Far-colo human readers still refresh on the normal TTL.

## Cost

Strictly better than shrinking the TTL (fix b). The delete is free (a colo cache op, not a DO read); the only cost is one rebuild on the next read in that colo — i.e. we pay a rebuild **only when the brief actually changed**, not on a timer. `/api/init` DO row-reads dominate Cloudflare spend, so a timer-based refresh was a non-starter. The 1800s TTL stays intact for the ~all reads where nothing changed.

## Scope

Per the issue, limited to the brief-compile/inscribe → homepage path. The broader write-triggered surfaces (classifieds create/expire, editor register/deactivate) share the pattern but are out of scope here.

## Test / verify

- `npm run typecheck` + `biome lint` clean.
- Existing brief tests pass (17/17). `edgeCacheDelete` is a no-op in the test runtime by design (shared `caches.default`), matching the other edge-cache helpers.
- Post-deploy manual check (from the issue): compile a brief, then `curl -s https://aibtc.news/api/init | jq .brief.compiledAt` should agree with `/api/brief` within seconds from the same region, not 30 min.

### DIFF
--- a/src/lib/edge-cache.ts
+++ b/src/lib/edge-cache.ts
@@ -200,6 +200,27 @@ export function edgeCacheDelete(c: AppContext, paths: string[]): void {
 }
 
 /**
+ * Evict the homepage bundle (/api/init) from the local colo's edge cache.
+ *
+ * Shared by the brief compile and inscribe success paths: both mutate brief
+ * state that /api/init embeds behind its 30-min s-maxage TTL — compile changes
+ * the brief body, inscribe changes the inscription id/txid (init.ts folds
+ * inscription state into its brief payload). Purging on write lets the writing
+ * agent's own follow-up reads go fresh within seconds instead of waiting out
+ * the TTL (#870).
+ *
+ * Only /api/init is purged. The /api/brief and /api/brief/:date read endpoints
+ * set Cache-Control but never call edgeCachePut, so they are not stored in this
+ * Workers cache — there is nothing here to evict, and listing them would imply
+ * a caching path that this codebase doesn't create. (If a Cloudflare zone Cache
+ * Rule caches them at the dashboard level, that's a separate namespace/config
+ * this Cache-API delete wouldn't reliably reach — tracked separately.)
+ */
+export function purgeHomepageBundle(c: AppContext): void {
+  edgeCacheDelete(c, ["/api/init"]);
+}
+
+/**
  * Fire a background rebuild for a stale cache entry, guarded by a short KV
  * lock so concurrent stale hits don't hammer the upstream (e.g. a cold
  * Durable Object) with duplicate rebuilds.
--- a/src/lib/edge-cache.ts
+++ b/src/lib/edge-cache.ts
@@ -166,6 +166,40 @@ export function edgeCachePut(
 }
 
 /**
+ * Purge one or more cached paths from the local colo's edge cache.
+ *
+ * Cloudflare's Cache API is colo-scoped: this evicts entries only in the data
+ * center where the current (write) request runs, not globally. That is exactly
+ * the coverage a write-side purge wants — the agent that just POSTed a compile
+ * or inscribe sits in one colo, and clearing that colo makes its own follow-up
+ * reads (and any co-located readers) see the new state within seconds instead
+ * of waiting out the s-maxage TTL. Far-colo readers still refresh on the normal
+ * TTL, which is acceptable for the glance-info the homepage bundle carries.
+ *
+ * Cost note: the delete itself is free (a colo cache op, not a DO read). The
+ * only downstream cost is that the next read in this colo is a MISS and rebuilds
+ * once — i.e. we pay one rebuild exactly when the underlying data changed, not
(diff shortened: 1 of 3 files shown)


arc0btc (commented):
Fixes the real problem cleanly — `/api/init` edge-caches at `s-maxage=1800`, and a write-side colo purge is strictly better than shrinking that TTL: you pay a rebuild only when the brief actually changed, not on a timer. `edgeCacheDelete`'s doc comment does a good job explaining the colo-scoping tradeoff (writer's own colo goes fresh in seconds, far-colo readers still ride the normal TTL) — that's the right call for this endpoint.

**What works well:**
- `edgeCacheDelete` mirrors the existing `edgeCachePut`/`edgeCacheMatch` shape (same `isTestEnv` guard, same `buildCacheKey`-style canonical URL construction, same `waitUntil` fire-and-forget) — no new pattern introduced, just the missing inverse of `put`.
- Placement in `brief-compile.ts` is on the success path only (after `saveResult.ok`, before the final `return`), so a failed compile doesn't spuriously evict a still-valid cache.
- Same correct placement in both the POST and PATCH inscribe handlers in `brief-inscribe.ts` — after `result.ok`, before returning.

**[question] Are `/api/brief` and `/api/brief/${date}` actually cached via `caches.default`?**
I traced both cached targets through the codebase and don't see either one calling `edgeCachePut`:
- `src/routes/brief.ts` sets `Cache-Control: public, max-age=60, s-maxage=300` on `GET /api/brief` and `GET /api/brief/:date`, but never imports `edge-cache.ts` or calls `caches.default.put()` anywhere.
- A repo-wide `grep -rn "caches.default" src/` only turns up hits inside `edge-cache.ts` itself.

Per `edge-cache.ts`'s own module docstring: *"Workers responses don't automatically populate Cloudflare's edge cache ... no downstream cache is ever asked unless the response is explicitly stored via `caches.default.put()`."* If that's accurate (and it matches how every other cached route in this repo — `init.ts`, `home-page.ts`, `beats.ts`, etc. — explicitly pairs `edgeCacheMatch`/`edgeCachePut`), then `/api/brief` and `/api/brief/${date}` were never populated in this cache to begin with, and `edgeCacheDelete(c, ["/api/brief", ...])` is a no-op against a key that's never written — harmless (per your own comment, "harmless for any path that was never edge-cached"), but it also means this PR doesn't actually address the secret-mars `/api/brief/{date}` staleness case cited in the PR description as the reason for including those two paths. Worth double-checking whether there's a Cloudflare zone-level Cache Rule (outside this repo, e.g. dashboard config) honoring the `Cache-Control` header independently of the Workers Cache API — if so, `caches.default.delete()` wouldn't reach that cache anyway, since it's a different cache namespace. Either way, if `/api/brief`/`/api/brief/{date}` genuinely aren't cached anywhere, the `/api/init` purge alone fully covers the homepage staleness issue (#870) and the other two paths could be dropped from the purge list to avoid implying a caching behavior that doesn't exist.

**Code quality notes:**
- `brief-compile.ts` inlines `edgeCacheDelete(c, ["/api/init", "/api/brief", \`/api/brief/${date}\`])` directly, while `brief-inscribe.ts` wraps the identical array in a local `purgeBriefCaches(c, date)` helper. Since both call sites purge the exact same three paths, consider hoisting one shared helper (e.g. exported from `edge-cache.ts` as `purgeBriefCaches(c, date)`) so the scope can't drift between the two files if it's ever changed.

**Operational context:** `/api/init` DO row-reads are already our dominant Cloudflare cost driver per prior ops notes, so keeping the 1800s TTL intact for the common case (nothing changed) while only paying a rebuild on real writes is the right tradeoff — glad this didn't go the "shrink the TTL" route.

PR #885, opened 2026-07-21
fix(brief): expose flat inscription_id/inscribed_txid on brief endpoints (#870)

biwasxyz:
Follow-up from the #870 / #884 investigation. **Not a cache fix** — a response-shape fix. Splitting it out of #884 (which is scoped to the `/api/init` purge) so each PR does one thing.

## What the investigation found

secret-mars reported that `GET /api/brief/{date}.inscription_id` returned `null` right after a successful inscribe, and attributed it to ~5-min `s-maxage` edge staleness. I probed the live endpoints:

- `/api/brief/{date}` returns **no** `cf-cache-status`, **no** `age`, **no** `x-edge-cache` even on back-to-back requests → it is **not edge-cached at all** (no Cloudflare Cache Rule; `brief.ts` never calls `edgeCachePut`). So caching could not have been the cause, and the data was always fresh.
- The real cause: inscription state was only exposed at the nested camelCase `.inscription.inscriptionId`. There is **no top-level `inscription_id` key** — the response spreads the compile-time `json_data` blob, which never carried one. A client polling `.inscription_id` reads a *permanent* undefined, not a lagging value.

Concretely, days after inscription, `/api/brief/2026-07-15` still shows top-level `inscription_id: null` while `.inscription.inscriptionId` and `/api/brief/2026-07-15/inscription` both return the real id. Three endpoints, three shapes for the same DB-backed fact.

## Fix

Add flat, snake_case `inscription_id` / `inscribed_txid` (read straight from the DB column) to `/api/brief` and `/api/brief/:date`, matching what `/api/brief/:date/inscription` already returns. Also added to the pre-compile empty branch as `null`, so the key is **always present** — a client polling it gets a real `null`, never `undefined`.

- Placed **after** `...jsonData` so the authoritative column value always wins over any stale inscription baked into `json_data`.
- The nested `inscription` object is untouched (frontend backward-compat).

## Test / verify

- New `src/__tests__/brief-inscription-fields.test.ts` — route-level, seeds a brief via `saveBrief`/`updateBrief` and asserts the flat fields both post-inscription and pre-inscription (present-as-null).
- Full suite green: **467/467**. `typecheck` + `biome lint` clean.

## Note

This supersedes the "secret-mars `/api/brief/{date}` sibling" rationale that #884 originally cited for purging brief URLs — that purge was dropped in review precisely because those endpoints aren't in the Workers cache. This PR fixes the actual root cause.

### DIFF
--- a/src/__tests__/brief-inscription-fields.test.ts
+++ b/src/__tests__/brief-inscription-fields.test.ts
@@ -1,0 +1,70 @@
+import { describe, it, expect } from "vitest";
+import { SELF, env } from "cloudflare:test";
+import { saveBrief, updateBrief } from "../lib/do-client";
+import type { Env } from "../lib/types";
+
+// `cloudflare:test` types env as Cloudflare.Env; the do-client helpers want the
+// app's own Env. Same cast the other DO-touching suites use.
+const testEnv = env as unknown as Env;
+
+/**
+ * /api/brief and /api/brief/:date must expose flat, snake_case inscription_id /
+ * inscribed_txid fields read straight from the DB column — the same shape the
+ * /api/brief/:date/inscription endpoint returns.
+ *
+ * Before this, inscription state on the brief endpoints was only reachable at
+ * the nested camelCase `.inscription.inscriptionId` path, while the compile-time
+ * `json_data` blob carried no flat field at all. A client polling
+ * `.inscription_id` therefore got a *permanent* undefined — not a value that
+ * lagged and cleared, but one that never existed. That was misread as a
+ * 5-minute edge-cache staleness (#870 investigation), when in fact
+ * /api/brief/:date is not edge-cached at all and the data was always fresh.
+ */
+async function seedBrief(date: string, updates?: { inscription_id: string; inscribed_txid: string }) {
+  const saved = await saveBrief(testEnv, {
+    date,
+    text: "brief body",
+    json_data: JSON.stringify({ sections: [], summary: { beats: 0, signals: 0, correspondents: 0 } }),
+    compiled_at: `${date}T01:00:00.000Z`,
+  });
+  expect(saved.ok).toBe(true);
+  if (updates) {
+    const updated = await updateBrief(testEnv, date, updates);
+    expect(updated.ok).toBe(true);
+  }
+}
+
+describe("/api/brief/:date — flat inscription fields", () => {
+  it("mirrors the DB column into flat inscription_id/inscribed_txid after inscription", async () => {
+    const date = "2026-06-10";
+    const inscriptionId = `${"a".repeat(64)}i0`;
+    const inscribedTxid = "b".repeat(64);
+    await seedBrief(date, { inscription_id: inscriptionId, inscribed_txid: inscribedTxid });
+
+    const res = await SELF.fetch(`http://example.com/api/brief/${date}`);
+    expect(res.status).toBe(200);
+    const body = (await res.json()) as Record<string, unknown>;
+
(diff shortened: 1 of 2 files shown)


arc0btc (changes requested):
Fixes the actual root cause behind #870: the `inscription_id`/`inscribed_txid` polling failure secret-mars saw wasn't edge-cache staleness — it was a response-shape gap, since `/api/brief` and `/api/brief/:date` never exposed a top-level `inscription_id`/`inscribed_txid` at all, only the nested `.inscription.inscriptionId`. Good diagnostic work narrowing that down before touching code.

**What works well:**
- Placing the flat fields *after* `...jsonData` in both handlers (`src/routes/brief.ts:96-100`, `:290-294`) is the right call — it guarantees the DB-backed column always wins over anything baked into the compiled `json_data` blob, which is exactly the kind of stale-vs-authoritative bug we've hit before with cached compile-time blobs.
- The empty-brief branch (`:64-65`) also sets `inscription_id`/`inscribed_txid` to `null` explicitly, so callers always get a real `null` instead of `undefined` — matches the PR's stated goal of never producing a permanently-missing key.
- Test coverage is solid: both the post-inscription (flat fields populated + nested object preserved for frontend compat) and pre-inscription (`null`-but-present) cases are asserted at the route level via `SELF.fetch`, not just unit-testing the helper.
- Correctly scoped — this PR doesn't touch the `/api/init` purge logic from #884, keeping it a single-purpose fix.

**[blocking] Unrelated binary file committed to repo root** (`629382b0-e971-5bac-9b92-30b5d7e5e7df.pdf`)
This PR adds a 44KB PDF with a UUID-style filename at the repo root that has nothing to do with the `inscription_id` fix — no reference to it in the diff description, code, or tests. This looks like an accidental commit (e.g. a local export/print-to-PDF artifact swept in with `git add`). Please drop it from the branch before merge — committing stray binaries to root bloats repo history permanently (removing it later doesn't shrink existing clones) and an unexplained PDF from an external contributor is worth treating with a bit of caution until its origin is confirmed.

**[nit]** The `inscription_id ?? null` / `inscribed_txid ?? null` pattern is repeated 3x across the file (empty branch, `/api/brief`, `/api/brief/:date`) with identical comments. Not blocking, but if a 4th endpoint ever needs the same shape, a small `flatInscriptionFields(brief)` helper would keep the "DB column wins" invariant in one place instead of copy-pasted comments.

*Operational note:* We don't currently poll these `inscription_id` fields ourselves, but this same "nested object only, no flat mirror" gap is a pattern worth watching for on other endpoints that expose DB-backed state — matches what we'd expect after an org-wide grep, worth keeping in mind for future changes to `brief.ts`-adjacent routes.

Once the stray PDF is dropped, this looks good to merge.

arc0btc (approved):
Re-review — both prior asks addressed.

**[blocking] Stray PDF removed** — the unexplained `629382b0-...pdf` binary is gone from the diff; only `src/routes/brief.ts` and the new test file remain.

**[nit] Dedup applied** — the repeated `?? null` pattern is now consolidated into `flatInscriptionFields()` (`src/routes/brief.ts:12-18`), called consistently from the empty branch, `/api/brief`, and `/api/brief/:date`. Cleaner than the copy-pasted version and keeps the "DB column wins over `json_data`" invariant in one place.

Logic is unchanged from the version we already reviewed: flat fields still spread after `...jsonData` so the DB column always wins, and the pre-inscription branch still emits explicit `null` rather than an absent key. Tests still cover both cases at the route level.

Approving.

PR #887, opened 2026-07-22
feat!: retire weekly leaderboard prize tier (#886)

biwasxyz:
Retires the automated top-3 weekly leaderboard prize tier. The Editor now pays quality signal filers manually, at editorial discretion, so the on-chain prize schedule no longer describes how correspondents actually get paid.

## Why now

The machinery was already effectively dead. Sweeping every ISO week of 2025–2026 against the live API, weekly prizes were recorded **exactly once — 2026-W13** (April 3). Three rows, all with real on-chain sBTC txids. Nothing in the ~15 weeks since.

The docs had also drifted out of sync with the code: `publisher.md` advertised 20000/10000/5000 sats while `constants.ts` paid 200000/100000/50000 — a 10× discrepancy that had been live for a while. Both are resolved here.

## Retired endpoints return 410, they are not deleted

`POST /api/leaderboard/payout` stays routed and returns **410 Gone** with a machine-readable tombstone:

```json
{
  "retired": true,
  "error": "The weekly top-3 leaderboard prize tier has been retired.",
  "reason": "Automated weekly prizes are no longer issued. The Editor now pays quality signal filers manually, at editorial discretion, on a weekly cadence.",
  "replaced_by": "Manual editor payouts — no API call is required or available to trigger them.",
  "action": "Stop calling this endpoint. Do not retry; this is permanent, not a transient failure.",
  "historical_records": "GET /api/leaderboard/payouts/:week still returns prizes paid before retirement."
}
```

This is deliberate. Publisher agents poll this endpoint weekly and unattended. A bare 404 reads as a transient routing fault worth retrying, so they would retry it indefinitely; 410 is terminal by HTTP definition and the body says in plain language what replaced it.

It answers **unauthenticated** on purpose — there is nothing left to protect, and making an agent sign a BIP-322 request just to learn the endpoint is gone wastes its time.

## The audit trail is preserved

`GET /api/leaderboard/payouts/:week` stays read-only. The 2026-W13 rows are real money already sent, and the Correspondent Guild reconciler (#454) matches them against on-chain txids — deleting the reader would hide a payment audit trail.

Its response now carries the same tombstone under a `retired` key, so an empty `payouts` array reads as *"the tier is gone"* rather than *"prizes have not been issued yet"*. Without that, a caller would sit and poll for a payout that will never arrive.

**No migration.** Existing `weekly_prize_*` earnings rows are untouched and still surface in `GET /api/earnings/:address`.

## Removed (write path)

- `WEEKLY_PRIZE_{1ST,2ND,3RD}_SATS` constants
- `POST /payouts/weekly` DO handler, along with its `weekly_payout` leaderboard snapshot write (the `launch_reset` snapshot path is unaffected)
- `recordWeeklyPayouts()` DO client
- `PayoutRecord` / `WeeklyPayoutResult` types

## Docs

- **`public/skills/publisher.md`** — the procedure a publisher agent actually follows, and it was still instructing agents to call the retired endpoint and wire sBTC prizes. The weekly checklist now pulls the leaderboard for editorial awareness only, keeps the treasury balance check (editor payouts and inscription fees still draw on it), and reports standings via Nostr without implying prizes.
- **`public/about/index.html`** — no longer advertises a 200k/100k/50k prize ladder that is not paid. Streaks are reframed as reputation and visibility rather than a share of prizes.
- **`src/routes/manifest.ts`** — the manifest is how agents discover endpoints, so it carries the retirement too; also adds the previously undocumented archival reader.

## Commit layout

One commit per file, ordered bottom-up (constants → types → client → DO → routes → tests → docs). Note that this means the intermediate commits do not typecheck in isolation — only the branch tip is green. Worth a squash merge.

## Verification

`typecheck`, `lint`, `deploy:dry-run`, and **470 tests across 47 files** all pass at the branch tip. Two new tests cover the 410 tombstone (including the unauthenticated path) and the archival `retired` flag.

## Deploy note

This removes a Durable Object route. Per the news-singleton's 50s keep-alive, merging alone will not cycle the DO onto the new code — this needs a manual `wrangler deploy` to actually take effect.

### DIFF
--- a/public/about/index.html
+++ b/public/about/index.html
@@ -408,7 +408,7 @@
       </div>
       <div class="about-step">
         <span class="about-step-num">5</span>
-        <div class="about-step-text"><strong>Earn Sats</strong> &mdash; Multiple earning paths: brief revenue share, weekly leaderboard prizes, and competitions. See <em>How Agents Earn</em> below.</div>
+        <div class="about-step-text"><strong>Earn Sats</strong> &mdash; Multiple earning paths: brief revenue share, editorial rewards for quality signals, and x402 intel sales. See <em>How Agents Earn</em> below.</div>
       </div>
     </div>
 
@@ -441,11 +441,11 @@
       </div>
       <div class="about-step">
         <span class="about-step-num">&bull;</span>
-        <div class="about-step-text"><strong>Weekly Leaderboard Prizes</strong> &mdash; Top correspondents earn weekly prizes: <strong>200,000 sats</strong> (1st), <strong>100,000 sats</strong> (2nd), <strong>50,000 sats</strong> (3rd).</div>
+        <div class="about-step-text"><strong>Editorial Rewards</strong> &mdash; The Editor rewards high-quality signals directly, at editorial discretion. There is no fixed schedule or automated prize tier &mdash; consistently good reporting is what gets paid.</div>
       </div>
       <div class="about-step">
         <span class="about-step-num">&bull;</span>
-        <div class="about-step-text"><strong>Streak Bonuses</strong> &mdash; Filing signals on consecutive days builds your streak. Longer streaks earn higher leaderboard position, increasing your share of weekly prizes.</div>
+        <div class="about-step-text"><strong>Streaks &amp; Standing</strong> &mdash; Filing signals on consecutive days builds your streak and raises your leaderboard position. Standing is reputation and visibility &mdash; it is not itself a payout.</div>
       </div>
       <div class="about-step">
         <span class="about-step-num">&bull;</span>
--- a/public/skills/publisher.md
+++ b/public/skills/publisher.md
@@ -85,13 +85,19 @@ For each payout, record:
 - Transaction ID
 - Number of signals paid for
 
-Maintain a running log for weekly leaderboard calculation.
+Maintain a running log of what was sent.
 
 ---
 
 ## Weekly Operations Checklist
 
-Run at the end of each 7-day competition period.
+Run at the end of each 7-day period.
+
+> **Weekly prizes are retired.** There is no automated top-3 prize tier and no API call
+> that issues one. `POST /api/leaderboard/payout` now returns **410 Gone** — do not call it,
+> and do not retry it. Quality signal filers are rewarded by the **Editor, manually**, at
+> editorial discretion. The leaderboard is a visibility and ranking surface only; position
+> on it does not entitle a correspondent to a payout.
 
 ### 1. Pull Leaderboard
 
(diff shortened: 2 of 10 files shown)


arc0btc (approved):
Retires the weekly top-3 leaderboard prize tier by returning a terminal 410 tombstone instead of deleting the route outright, and cleans up the write-path code that used to back it. Good, tightly-scoped change.

**What works well:**
- The 410-with-tombstone pattern (`src/routes/leaderboard.ts`) is exactly right for an unattended weekly poller. A bare 404 reads as "maybe transient, retry later" to an agent; a 410 with `retired: true`, `reason`, `replaced_by`, and an explicit "do not retry" is unambiguous. This mirrors our own operational lesson about giving polling agents a terminal signal instead of letting them loop forever.
- Leaving `GET /api/leaderboard/payouts/:week` read-only rather than deleting it is the correct call — the 2026-W13 rows carry real on-chain sBTC txids that the Correspondent Guild reconciler (#454) still needs. Adding the `retired` flag to that response so an empty `payouts` array reads as "tier is gone" rather than "nothing paid yet" closes a real ambiguity.
- Dead code removal is thorough and consistent: `WEEKLY_PRIZE_*_SATS`, `PayoutRecord`/`WeeklyPayoutResult` types, `recordWeeklyPayouts()`, the `/payouts/weekly` DO handler, and the now-unused `getPreviousISOWeek()` helper in `leaderboard.ts` are all removed together — no orphaned imports left behind (verified `do-client.ts`, `news-do.ts`, `leaderboard.ts` import lists were all updated to match).
- `publisher.md` and `public/about/index.html` are updated in the same PR, not left to drift — good, since the docs were already 10x out of sync with the actual constants before this (200k/100k/50k vs. advertised 20k/10k/5k), and that's called out explicitly rather than quietly fixed.
- New tests cover the 410 tombstone (including the unauthenticated path, which is intentional since there's nothing left to protect) and the archival `retired` flag on the GET endpoint. Confirmed at the branch tip there's no stale test still asserting a 201 for the old POST handler — the whole test suite is internally consistent with the new behavior.
- All CI checks pass (typecheck, CodeQL, Snyk, Workers build/deploy, preview).

**[question] Deploy note in the PR body**
The description flags that this removes a Durable Object route and, per the news-singleton's 50s keep-alive, merging alone won't cycle the DO onto new code — a manual `wrangler deploy` is needed after merge. Worth confirming whoever merges actually runs that step; we've been bitten before by "merged" reading as "deployed" when they weren't the same event. Not a blocker on the code, just flagging so it doesn't get lost between merge and verification.

**Code quality notes:**
- No reuse or simplification issues — the diff is a clean removal plus a small, well-commented tombstone constant. Nothing to simplify further.

*Operational context:* we don't operate this endpoint ourselves, but the "terminal status code + machine-readable body" pattern here is one we'd want for our own polling sensors when we retire something — good reference implementation.

Approving.

